PR #5, opened 2020-12-01
added wanchain collateral calculation

Jeffqg123:
FinNexus projects tlv caculation is updated, added the collateral caculation from wanchain blockchain


Jeffqg123:
Why the TLV in DefiLlama page is bigger 100 times than FinNexus actual number.
What problems with that ? How to deal with it?

PR #17, opened 2021-01-05
Add LINKSWAP adapter

giovannidisiena:
This PR adds an adapter for the YF Link community-governed automated market maker LINKSWAP.

PR #18, opened 2021-01-06
Add governance vault to reflect true YF Link TVL

giovannidisiena:
This PR follows from #17 and adds the YF Link governance vault to accurately reflect the true ecosystem TVL, which should currently be approximately $44,000.


DefiLlama:
Added

giovannidisiena:
> Added

@DefiLlama amazing, thanks for doing this so quickly!

PR #33, opened 2021-02-09
Update of Mantra Dao's borrowing and lending assets

spaceforce-dev:
We have added a large number of new assets that we would like to track. Thanks!


spaceforce-dev:
Thanks for merging! Any idea when you will push the update live?

Thanks

PR #39, opened 2021-02-21
Remove comments

spaceforce-dev:
Left some ugly comments in merge. 

Deleting.


DefiLlama:
I had already removed them locally, will push at one point.

PR #72, opened 2021-03-22
Add swipeswap project

gryincent:

### DIFF
--- a/projects/swipeswap.js
+++ b/projects/swipeswap.js
@@ -1,0 +1,86 @@
+const retry = require('async-retry')
+const axios = require("axios");
+const { __Directive } = require('graphql');
+const { GraphQLClient, gql } = require('graphql-request')
+
+async function fetchExchangeTVL() {
+
+  var totalValue = 0;
+
+  // total liquidity value
+  var exchangeFetchData = [
+    {
+      address: '0x8a93b6865c4492ff17252219b87ea6920848edc0',
+      endpoint: 'https://api.thegraph.com/subgraphs/name/swipewallet/exchange',
+    },
+    {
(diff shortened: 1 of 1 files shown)


0xngmi:
Hey! First of all, thanks for submitting the adapter. I've just taken a look at the source code of the subgraph (which I presume to be the one at https://github.com/SwipeWallet/swipeswap-subgraph) and, looking at lines 353-362 of `src/exchange/mappings/pair.ts`, we see the following code:
```ts
  // use derived amounts within pair
  pair.trackedReserveETH = trackedLiquidityETH
  pair.reserveETH = pair.reserve0
    .times(token0.derivedETH as BigDecimal)
    .plus(pair.reserve1.times(token1.derivedETH as BigDecimal))
  pair.reserveUSD = pair.reserveETH.times(bundle.ethPrice)

  // use tracked amounts globally
  factory.liquidityETH = factory.liquidityETH.plus(trackedLiquidityETH)
  factory.liquidityUSD = factory.liquidityETH.times(bundle.ethPrice)
```

From this it seems that the factory entity is just storing aggregate metrics over all pairs, as whenever a pair's TVL is updated, the global factory TVL is also updated with a similar value. Could you confirm whether this is true?

If it is, it would mean that currently the adapter is double counting the TVL, as it's summing the factory-provided one and the TVL of all pairs, so, following a [user report](https://twitter.com/Sam42867090/status/1374033322490404864) I've decided for now to err on the side of caution and disable that part of the adapter.

However, I'm aware that I'm not knowledgeable about swipeswap, so there might be something I'm missing. If so, please let me know what that is and I'll quickly revert my updates on the adapter. I'd also really appreciate if you could point me which contracts are being counted towards the factory TVL, as that would greatly speed up my research.

PR #81, opened 2021-03-26
add allinx.js

allinxproject:

### DIFF
--- a/projects/allinx.js
+++ b/projects/allinx.js
@@ -1,0 +1,14 @@
+const utils = require('./helper/utils');
+
+/* * * * * * * *
+* ==> Correct adapter needs to be created.
+*
+*****************/
+async function fetch() {
+  let staked = await utils.fetchURL('https://api.allinx.io/api/stats')
+  return staked.data.data.totalLockedValue;
+}
+
+module.exports = {
+  fetch
+}


0xngmi:
Could you please:
- fill in all the questions asked on the PR template
- provide an adapter that calculates the TVL from blockchain calls instead of through an API endpoint

allinxproject:
> Could you please:
> 
> * fill in all the questions asked on the PR template
> * provide an adapter that calculates the TVL from blockchain calls instead of through an API endpoint

Description of Readme

“2.Create a function called fetch returning your TVL (and export it)”

I fork the template from venus:

https://github.com/DefiLlama/DefiLlama-Adapters/blob/main/projects/venus.js

"provide an adapter that calculates the TVL from blockchain calls instead of through an API endpoint"

Do you mean to read totalSupply from the contract:

https://docs.llama.fi/submit-a-project

But does the wbtc.js code mean only the get total supply of the token, and what is the relationship with TVL?

0xngmi:
I've re-written the docs and hopefully now they should be easier to understand.

I've also written a prototype of your adapter on #81, which right now doesn't capture all the TVL because it's missing some LP tokens. Given that the LP contracts were not verified on etherscan and they don't follow the uniswap interface I was unable to unwrap those tokens.

0xngmi:
Replaced by #82

PR #82, opened 2021-03-27
Add allinx prototype

0xngmi:
See #81


0xngmi:
Closed in favor of #81

PR #85, opened 2021-03-30
added bsc chain option pool for finnexus

Jeffqg123:
Finnexus team deploy option pool on BSC chain,  this update is to add the tvl from bsc finnexus option pool

Thanks

jqg


0xngmi:
Thanks!

PR #90, opened 2021-03-31
Use Rari API

TransmissionsDev:
This PR updates the Rari Capital TVL fetching to use our API instead of fetching manually.


0xngmi:
Thanks for the PR. Do you have an API/subgraph from which we could pull historical TVLs or vault addresses? I'm trying to improve the historical graphs of protocols by moving the adapters to a new standard that makes it possible to get a protocol's TVL at a point in time.

TransmissionsDev:
> Thanks for the PR. Do you have an API/subgraph from which we could pull historical TVLs or vault addresses? I'm trying to improve the historical graphs of protocols by moving the adapters to a new standard that makes it possible to get a protocol's TVL at a point in time.

We have a prometheus instance that tracks our TVL... would that be useable?

0xngmi:
Is prometheus [this](https://prometheus.io/)? If we can query TVL by timestamp or if we can get a list of timestamps and TVLs associated that would be enough.

TransmissionsDev:
> Is prometheus [this](https://prometheus.io/)? If we can query TVL by timestamp or if we can get a list of timestamps and TVLs associated that would be enough.

Yep yep. When you guys start doing retroactive fetching let us know and we can help get everything setup

0xngmi:
It seems like the API endpoint is down

TransmissionsDev:
> It seems like the API endpoint is down @TransmissionsDev

Indeed sorry! Working on resolving asap.

PR #103, opened 2021-04-06
Alchemix Transmuter Migration Fixes

0xdef1:
Fix error in ALCX-ETH tvl calc, add yvdai value from alchemist, add alusd-3crv value, add alusd staked value, remove 'alusdSupply * 2' calc


0xngmi:
Thanks!

PR #105, opened 2021-04-07
Add TVL adapter for Lido's bLuna

vshvsh:
##### Twitter Link:

www.twitter.com/lidofinance

https://github.com/lidofinance/audits

##### Website Link: 

www.lido.fi

##### Contact Email: 

info@lido.fi

##### Current TVL: 

$765.6m 

##### Chain: 

Ethereum, Terra

##### Coingecko ID (so your TVL can appear on Coingecko): 

lido-dao

##### Coinmarketcap ID (so your TVL can appear on Coinmarketcap): 

8000

Stake your ETH with Lido and receive liquid stETH with daily rewards.

* LDO;  https://etherscan.io/token/0x5a98fcbea516cf06857215779fd812ca3bef1b32
* stETH; https://etherscan.io/token/0xae7ab96520de3a18e5e111b5eaab095312d7fe84 
* bLUNA; https://finder.terra.money/columbus-4/address/terra1kc87mu460fwkqte29rquh4hc20m54fxwtsx7gp

Yield, Staking

### DIFF
--- a/projects/bluna.js
+++ b/projects/bluna.js
@@ -1,0 +1,24 @@
+const utils = require('./helper/utils');
+
+async function fetch() {
+  let tvl = 0;
+
+  let { total_bond_amount } = (
+      await utils.fetchURL("https://lcd.terra.dev/wasm/contracts/terra1mtwph2juhj0rvjz7dy92gvl6xvukaxu8rfv8ts/store?query_msg=%7B%22state%22%3A%20%7B%7D%7D")
+    ).data.result;
+
+  let { amount } = (
+    await utils.fetchURL("https://lcd.terra.dev/market/swap?offer_coin=1000000uluna&ask_denom=uusd")
+  ).data.result;
+
+  tvl = Math.floor(
+    (amount*1.0/1000000)*(total_bond_amount/1000000)
+  );
+
+  return tvl;
+}
+
+
+module.exports = {
+  fetch
+}


0xngmi:
Thanks for the PR! Do you know if there's a way to get bLUNA's TVL at a previous point in time?

vshvsh:
You can do archive requests for the same data (just add `&height=2317393` to a request)  - but we didn't set up a public archive Terra node yet

Can have it in a few days, is it a blocker?

0xngmi:
Not a blocker but it'd be nice. If it works for you I can merge the PR now and, when the archive node is ready, I'll modify the adapter to also retrieve historical values.

0xngmi:
If this is a matter of getting an archive node (thus not specific to LIDO), do you know if there's any providers from which we could buy one?

vshvsh:
Don't know any! I think there will be, but for now we'll maintain our own

0xngmi:
Got it. I've just merged the adapters and pushed the new versions to the server. Can you ping me when you have the archive node available?

0xngmi:
I've gone ahead and implemented some code to fetch historical values and report the currencies separately. We'll use this to fill the chart and show how the TVL composition evolves over time.

vshvsh:
Just synced an archive node yesterday - do you still need it?

vshvsh:
Oh terra actually made their endpoint an archive node? That code would have worked in April.

vshvsh:
(it took almost a month to sync the archive terra node properly)

0xngmi:
I'm not sure, I thought so since it seems to respond fine when adding the height parameter but I'm getting some ` 'not found: contract terra1mtwph2juhj0rvjz7dy92gvl6xvukaxu8rfv8ts'` errors that shouldn't be there

0xngmi:
So it may not be an archive node, running some more tests now

vshvsh:
That contract didn't exist before middle of march - is that consistent with your errors? Also can look at height, this might be a sparsely pruned node with only data available at every X block (typically every 100 by default in that mode)

0xngmi:
Yep, it seems to be the case, when you query it at height `2792839` (1st of may) it returns an error:
https://lcd.terra.dev/wasm/contracts/terra1mtwph2juhj0rvjz7dy92gvl6xvukaxu8rfv8ts/store?query_msg=%7B%22state%22%3A%20%7B%7D%7D&height=2792839

0xngmi:
> Also can look at height, this might be a sparsely pruned node with only data available at every X block (typically every 100 by default in that mode)

Will run some tests with other heights, that may be the case

vshvsh:
https://lcd.terra.dev/wasm/contracts/terra1mtwph2juhj0rvjz7dy92gvl6xvukaxu8rfv8ts/store?query_msg=%7B%22state%22%3A%20%7B%7D%7D&height=2792800 works
it's a semi-pruned node, historical queries only work at every 100th block.

vshvsh:
we can get you a proper archive node now, but it'll be less reliable than official endpoint

0xngmi:
Do you know what's the average blocktime? (to measure precision loss)

0xngmi:
> (it took almost a month to sync the archive terra node properly)

F

0xngmi:
> we can get you a proper archive node now, but it'll be less reliable than official endpoint

It seems like 100 blocks of difference only results in an average precision loss of 5 minutes so using the official node will be just fine, that way we also don't impose on you.

Also, thanks for chiming in. Letting me know about pruned nodes just saved me several hours of work :clinking_glasses:

0xngmi:
Historical data is now available on defillama along with the token breakdown:
image

PR #109, opened 2021-04-14
Wasabix adpter update

wasabi-swap-team:
update adapter to include waBTC related values


0xngmi:
I see this assumes BTC=waBTC, how well is the peg being maintained?

PR #110, opened 2021-04-15
Add all IndexCoop Products

ncitron:
Update the IndexCoop space to include our newest products. Right now this adapter is called dpi.js, should we rename it to indexcoop.js?


0xngmi:
Can you confirm that the total amount should be ~191M?

ncitron:
Yes. 191M is correct.

0xngmi:
I've moved the adapter to an SDK-based one and renamed it

PR #112, opened 2021-04-15
Add stacks

aulneau:
##### Twitter Link:
https://twitter.com/stacks

https://forum.stacks.org/t/stacks-2-0-security-audits/11464

##### Website Link:
https://stacks.co

##### Contact Email:
community@stacks.org

##### Current TVL:
$992,353,822.57 USD

##### Chain:
Stacks

`blockstack`

`stacks`

Stacks makes Bitcoin programmable, enabling decentralized apps and smart contracts that inherit all of Bitcoin’s powers.

STX

Proof-of-Transfer (PoX) / Yield

---

I have added a few dependencies in order to fetch data directly from the contract, I couldn't find much info on what the policy is for that. I am happy to create an api wrapper for this (I run https://stacking.club). Just let me know!

Also what kind of email do you need? If it's for the project as a whole, it should not be my email :~)

### DIFF
--- a/projects/stacks.js
+++ b/projects/stacks.js
@@ -1,0 +1,53 @@
+const {Configuration, SmartContractsApi, InfoApi} = require('@stacks/blockchain-api-client');
+const {cvToHex, cvToString, hexToCV, uintCV} = require('@stacks/transactions');
+const retry = require('async-retry')
+const fetcher = require('cross-fetch');
+
+const config = new Configuration({
+    fetchApi: fetcher,
+    basePath: 'https://stacks-node-api.mainnet.stacks.co',
+});
+
+const principal = 'SP000000000000000000002Q6VF78.pox';
+
+const contractsApi = new SmartContractsApi(config);
+const infoApi = new InfoApi(config);
+
+function uStxToStx(microStx) {
+    return Number(Number(microStx) / Math.pow(10, 6));
+}
+
+async function getStxPrice() {
+    const res = await fetcher('https://api.coingecko.com/api/v3/simple/price?ids=blockstack&vs_currencies=usd');
+    const data = await res.json();
+    return data.blockstack.usd;
+}
+
+async function getTotalStacked() {
+    const [contractAddress, contractName] = principal.split('.');
+    const info = await infoApi.getPoxInfo()
+    const data = await contractsApi.callReadOnlyFunction({
+        contractAddress,
+        contractName,
+        functionName: 'get-total-ustx-stacked',
+        readOnlyFunctionArgs: {
+            sender: contractAddress,
+            arguments: [cvToHex(uintCV(info.reward_cycle_id))],
+        },
+    });
+    if (data.okay) {
+        const total_stacked = parseInt(cvToString(hexToCV(data.result)).replace('u', ''), 10);
+        return uStxToStx(total_stacked)
+    }
+}
+
+
+async function fetch() {
+    const usd = await retry(async bail => getStxPrice());
+    const stx = await retry(async bail => getTotalStacked());
+    return (stx * usd).toFixed(2);
+}
+
+module.exports = {
+    fetch
+}


0xngmi:
> Also what kind of email do you need? If it's for the project as a whole, it should not be my email :~)

Yours is fine, I'll just remove the question given that we don't use this anymore.

From some quick research this seems to be the staked STX on the Stacks blockchain, which would be similar to staked COSMOS or staked tokens on other blockchains that offer this kind of system. Could you confirm if this is correct?

aulneau:
> staked STX on the Stacks blockchain

Similar yes -- "Stacking" is the process in which holders of the native token on Stacks (STX) commit/lock their tokens via the [pox smart contract](https://explorer.stacks.co/txid/SP000000000000000000002Q6VF78.pox) for a given duration and in return receive bitcoin sent to addresses they provide by stacks miners. It is a core element of the consensus mechanism for the stacks blockchain.

Happy to answer any other questions you might have.

0xngmi:
Gave this some thought. In general, native blockchain staking is not considered DeFi nor TVL by the vast majority of people (for example nobody associates COSMOS staking with COSMOS TVL) so I'm a bit hesitant to list this project. However we want to be as inclusive as possible so what I could do is list STX staking with the following conditions:
- TVL won't be added to the general TVL
- It will be under a especial category (something like chain staking)

What do you think?

Feel free to also let me know if the comparisons I've drawn are wrong.

aulneau:
@0xngmi I see, yeah that makes sense!

I think under a special category makes sense. I know that there will be updates and new layers built such that those could fall under the real "DeFi" category :~). Until then this sounds super reasonable. Appreciate the inclusivity!

If there is anything else, please let me know how I can be of any more help with regards to the adapter.

0xngmi:
Thanks for understanding, could you change the adapter to make it provide the TVL at a point in time instead of simply providing the current one? This will allow us to back-fill the chart.

On top of that, could you try to remove the dependency on cross-fetch? We already use a bunch of HTTP libraries and I'd like to avoid adding another one on top.

aulneau:
> Thanks for understanding, could you change the adapter to make it provide the TVL at a point in time instead of simply providing the current one? This will allow us to back-fill the chart.

Hmm this is not exactly straight forward. Could you give more details about the exact kind of arguments you would need to pass? I would have to create an api endpoint on `stacking.club` to be able to expose that data -- it would no longer be directly from the contract. 

"Stacking" on the Stacks blockchain happens at a predictable cycle, which is aprox 2 weeks in duration. We are on cycle 6. I could do an endpoint that fetches the value for each cycle, but there would only be 6 data points currently.

Otherwise I could expose one that is similar to what is on https://stacking.club/cycles/next (see screenshot) but have it be all time.

Screen Shot 2021-04-16 at 1 38 26 PM

aulneau:
@0xngmi After thinking about it more, I will expose an endpoint to give you historical data on the total amount locked, not by cycle but by transaction. Once you let me know an optimal format for it I can update this PR :)

0xngmi:
The format that would make it easiest for me to integrate would be some endpoint where I pass a UNIX timestamp and it returns both the amount of STX locked and it's usd equivalent at that point in time.

0xngmi:
Something else that would work as well is some adapter code that, given a timestamp, gets the cycle it's in and returns that amount. Both would work for me so pick whichever you prefer.

aulneau:
> The format that would make it easiest for me to integrate would be some endpoint where I pass a UNIX timestamp and it returns both the amount of STX locked and it's usd equivalent at that point in time.

Okay, sounds good, will have something for you in the next couple of day.

One more question -- is this something that will be fetched directly on the client of your application, or is it behind an api/microservice?

0xngmi:
> One more question -- is this something that will be fetched directly on the client of your application, or is it behind an api/microservice?

We fetch these from our server hourly and then store it on our database, from which we serve the requests to the clients. Traffic to your endpoint should be pretty low.

Here's the source code: https://github.com/DefiLlama/defillama-server

0xngmi:
image

aulneau:
@0xngmi haha greetings! I have a working endpoint: `https://api.stacking.club/api/tvl?timestamp=1613125556000` as an example returns `{"stx":202157971.54764,"usd":130176943.73936343}`

I'm curious about the fetch method to export from the adapter, I don't see another example of an adapter that can take a parameter. Can you give an example?

Excited to get this through!! :)

0xngmi:
Sure, instead of a fetch adapter it would be better to use an SDK-based adapter here. Happy to take over and do that part myself since it should be pretty easy with your endpoint

aulneau:
Great! sounds good, please let me know how I can help further. And it should be able to handle the kind of traffic that you will send to it, but I have very occasionally been rate limited by coingecko when fetching historical prices for a larger set of dates. If anything pops up, happy to help debug it :)

aulneau:
@0xngmi one more note, the earliest timestamp you can request data for is `1610654610000`

0xngmi:
>  And it should be able to handle the kind of traffic that you will send to it, but I have very occasionally been rate limited by coingecko when fetching historical prices for a larger set of dates

If that is a problem feel free to remove the usd value from the response and the dependency on coingecko, I've already built a system that gets the prices from coingecko so I can just use that here.

aulneau:
> > And it should be able to handle the kind of traffic that you will send to it, but I have very occasionally been rate limited by coingecko when fetching historical prices for a larger set of dates
> 
> If that is a problem feel free to remove the usd value from the response and the dependency on coingecko, I've already built a system that gets the prices from coingecko so I can just use that here.

Oh, yeah that would make it easier -- I'll remove that quick :)

0xngmi:
Thanks!

aulneau:
@0xngmi alright, should be ready for ya `https://api.stacking.club/api/tvl?timestamp=1613125556000` now will return a `number` (up to 6 decimals) that represents the TVL of stx at the given timestamp (in ms).

PR #114, opened 2021-04-19
Added tvl for liquidity mining

neptune-v:
Added TVL for Bella Liquidity Mining


0xngmi:
Thanks! If you could change the adapter to be SDK-based (similar to what defipulse does) I'll be able to back-fill historical values and, in the future, display the breakdown of tokens locked in the protocol.

PR #118, opened 2021-04-22
Add Aave V2 data to DeFi Saver adapter

sterlu:
DeFi Saver recently enabled automation for the Aave V2 default market on the ETH chain. This PR updates the adapter to include the Aave V2 data.


0xngmi:
Thanks, if you convert it into an SDK adapter I'll be happy to change historical data aswell.

sterlu:
I thought about that but we don't really have the resources to get into it atm. Plus the data is mostly accurate as it is, so it should be good enough. Thanks for the quick merge 🚀

0xngmi:
No problem, thanks for the PR!

PR #124, opened 2021-04-26
Create Nerve SDK Adapter

aureliusbtc:
Nerve Square Icon Logo


0xngmi:
This data fetched from coingecko is provided by defistation which then fetches it from the source. Would it be possible to change it to get the data from the source as well?

PR #130, opened 2021-04-29
Add DeversiFi ETH TVL

plutoegg:


0xngmi:
Doesn't it have to be multiplied by the ETH price?

plutoegg:
> Doesn't it have to be multiplied by the ETH price?

Thats a pretty good point :)

0xngmi:
Thanks!

PR #138, opened 2021-05-12
changed eth web3 for FinNexus

Jeffqg123:
FinNexus ETH TVL is not correct, try use infura to fix it

### DIFF
--- a/projects/finnexus.js
+++ b/projects/finnexus.js
@@ -1,8 +1,8 @@
 var Web3 = require('web3');
 const env = require('dotenv').config()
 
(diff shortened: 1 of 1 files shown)


0xngmi:
This current PR is based on a stale branch. In order to fix yesterday's data error I've reimplemented the adapter to work with the raw token amounts and tomorrow I'll be backrunning it to replace old data.

0xngmi:
As we discussed before, the error that appeared on the chart was caused by a problem on the RPC nodes that day, and this has already been fixed, thus this PR is not needed anymore.

In regards to the historical error, I was waiting for my node provider to fix their nodes in order to run the script again at older block heights, but still haven't done it so I just went ahead and fixed the error manually, sorry for the wait, I should have done that in the start.

PR #143, opened 2021-05-14
refactoring of DefiLlama adpeter

Chiptuttofuso:


0xngmi:
Could you make the adapter return the balance of each token and use the block provided as an argument? That way we can display that information

PR #145, opened 2021-05-14
secretswap

scrtisland:

### DIFF
--- a/projects/secretswap/index.js
+++ b/projects/secretswap/index.js
@@ -1,0 +1,12 @@
+const retry = require('async-retry')
+const axios = require("axios");
+
+async function fetch() {
+  var response = await retry(async bail => await axios.get('https://api.secretanalytics.xyz/secretswap/tvl'))
+
+  return response.data || 0;
+}
+
+module.exports = {
+  fetch
+}


0xngmi:
Listed: https://defillama.com/protocol/secretswap

scrtisland:
Thanks @0xngmi . Can we add the token please? (below)

"id":"sefi","symbol":"sefi","name":"SEFI"

Coingecko ID (so your TVL can appear on Coingecko): sefi


Coinmarketcap ID (so your TVL can appear on Coinmarketcap): 9123


Description/bio (to be shown on DefiLlama): SecretSwap is the first front-running resistant, cross-chain and privacy-first AMM - secured by Secret Network.

Token address and ticker if any: 0x773258b03c730f84af10dfcb1bfaa7487558b8ac

0xngmi:
Done, I didn't see the token on the PR so I assumed it had no token, my bad.

0xngmi:
A day ago the tvl dropped heavily. Is this a real drop or is the api wrong?

0xngmi:
the api has stopped working, could you check on it and change the code to use the new api? I've looked at the requests made from the site and it seems to get the data from http://secretanalytics.xyz/_dash-update-component

I haven't updated the adapter because that api doesn't seem to be stable

pmuecke:
> the api has stopped working, could you check on it and change the code to use the new api? I've looked at the requests made from the site and it seems to get the data from http://secretanalytics.xyz/_dash-update-component
> 
> I haven't updated the adapter because that api doesn't seem to be stable

Hi, thank you for your message. We are upgrading the chain tomorrow so there is some work ongoing on the api in anticipation of that. Should be back up again soon though. It might be stale temporarily tomorrow while upgrading.

PR #157, opened 2021-05-26
Update Nexus Mutual pools and assets

horodnicdragos:


0xngmi:
Why remove the old pools?

roxdanila:
We're upgrading the pool to a new version and all assets will move over. The upgrade happens in ~30 mins.

0xngmi:
Okay, that explains why right now the adapter is reporting 0 TVL

0xngmi:
Seems to be working now, are you seeing any incorrect data?

PR #158, opened 2021-05-27
feat update fortube-v3 tvl

lrmuyi:

### DIFF
--- a/projects/fortube-v3/index.js
+++ b/projects/fortube-v3/index.js
@@ -5,7 +5,7 @@ const utils = require('./helper/utils');
 *
 *****************/
 async function fetch() {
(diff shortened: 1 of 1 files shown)


0xngmi:
I took a look at this one month ago but I couldn't really find where the TVL reported on that endpoint is coming from:

0xngmi:
The numbers on the previous image are outdated now but if you could help me find out how that number is calculated I'd appreciate it.

zhuoyue95:
The TVL is calculated from sum of the product of the value locked (in token amount) and its current price for each asset, in which the value locked for each asset requires more complicated calculation due to accrued and unaccrued interest for both deposits and loans. We calculate the value locked for an asset as follows:

value_locked = deposits - withdrawals - loans + repayments + repayments_through_liquidation

0xngmi:
Thanks for the PR!

PR #161, opened 2021-05-27
add okexchain

wepiggy-dev:

### DIFF
--- a/projects/wepiggy.js
+++ b/projects/wepiggy.js
@@ -2,55 +2,111 @@
   Modules
   ==================================================*/
 
-  const web3 = require('./config/web3.js');
-  const BigNumber = require('bignumber.js');
-  const abi = require('./config/wepiggy/abi.json');
-  const utils = require('./helper/utils');
+const sdk = require('@defillama/sdk');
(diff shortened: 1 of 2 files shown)


projects/wepiggy.js
```diff
@@ -2,55 +2,111 @@
...
-      }))
-    );
+async function fetch() {
+  let tvl = new BigNumber('0');
+  await Promise.all(
+    Object.keys(contracts).map(async key => {
+      let { comptroller, oracle } = contracts[key];
+      let chain = key;
+      let block = null;
+      // const { block } = await sdk.api.util.lookupBlock(timestamp, {
```
0xngmi, line 89: Did you comment this out because okex rpc is not an archive node?

wepiggy-dev: Because there are no parameters for fetch function. Tvl function has timestamp parameters but it is required to return the Balances object, which temporarily cannot solve the price problem. So we need to commented this first and use it later.

0xngmi: Makes sense. Thanks for submitting this, I'll got ahead and merge!

PR #166, opened 2021-05-29
TVL in bsc & fantom chain

petrovska-petro:
I added the code to have the breakdown of tvl in the markets of fantom and bsc chain. Also, fix the CRETH2 accumulated() method call.

Notes: 

- There is a minor issue for the calcs for bsc chain, which it is that token 0xB8c77482e45F1F44dE1745F52C74426C631bDD52 (BNB) it is assumed to be $0. So, it is missing a major portion of the bsc chain market place (at the moment $15M)
- Also some other tokens are defaulted to 0, which can be inspect in the console output while running this file thru the @defillama/sdk


projects/cream/index.js
```diff
@@ -198,57 +197,81 @@ const crYFI = "0xCbaE0A83f4f9926997c8339545fb8eE32eDc6b76";
...
-  const lpPositions = []
+  const lpPositions = [];
   cashValues.map((cashVal, idx) => {
-    if(underlyings[idx].output === null){
+    if (underlyings[idx].output == CRETH2) {
+      /* 
+        Despite of the underlying being the CRETH2, it is like sort of "wrapped" ETH version, so we can account as ETH pricing
+        Otherwise the tvl will miss M of $ into the calc
+      */
+      sdk.util.sumSingleBalance(
```
corollari, line 244: Won't this double-count TVL?

corollari: You are counting it here and further down a second time

PR #167, opened 2021-05-29
Update TrueFi adapter to include tfUSDC

yuchenlintt:

### DIFF
--- a/projects/truefi/index.js
+++ b/projects/truefi/index.js
@@ -2,19 +2,26 @@ const sdk = require('@defillama/sdk');
 const abi = require('./abi.json');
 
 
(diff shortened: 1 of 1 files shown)


0xngmi:
Thanks! Is there a subgraph or API we could use to pull these addresses from and keep them up-to-date?

yuchenlintt:
No problem! These contracts are behind proxies, so they shouldn't change address when we upgrade.

For now as we launch new lending pools we plan to manually send update PRs. I believe we don't have an API yet.

0xngmi:
Got it, if it works for you then no problem.

PR #170, opened 2021-05-30
Updating Index Coop Adapter

0xModene:
The Index Coop adapter needed updating to account for the deprecation of the CGI token. While in the adapter, I also renamed the FLI product line addresses to be more future proof/readable


0xngmi:
Why not leave it there? Will the total supply be left there even after the tokens becomes worthless upon deprecation? The benefit of leaving it there is that historical TVLs will continue being accurate

PR #178, opened 2021-06-02
update belt tvl

rlekdms:


0xngmi:
I'd rather avoid using the TVL provided by belt.

This is because there's some double counting between their yield aggregator and the curve pool. In other words, a user can take 1$, deposit it on the yield aggregator and then take the debt-token and deposit it on the curve pool, at which point belt counts it as 2$ towards their TVL, which I believe is double counting and I'd rather avoid.

Happy to discuss this btw, the original discussion on this is public on our discord

0xngmi:
Going to do a pretty big rebase, so I'll just leave the code here and close this PR. Let's continue this discussion, and, if we end up agreeing on how to move this forward I'll push this code into master directly.

Code:
```
    var data = await utils.fetchURL('https://s.belt.fi/info/all.json')
    let bscTvl = data.data.info.beltStat.tvl.bsc
    let hecoTvl = data.data.info.beltStat.tvl.heco
    let tvl = Number(bscTvl) + Number(hecoTvl)
    return tvl;
```

Please don't consider me closing this PR as a sign of trying to halt discussion, it's purely for tehcnical reasons (I'll rewrite git history so this PR will be listed as changing all files).

PR #179, opened 2021-06-02
Fixed: SSL error for MochiSwap Harmony Graph

frankydev1:
SSL is now fixed! The Harmony graph URL should no longer return ssl error. This push re-adds harmony TVL. 

Thanks!

### DIFF
--- a/projects/mochiswap/index.js
+++ b/projects/mochiswap/index.js
@@ -38,11 +38,8 @@ module.exports = {
   bsc: {
     tvl: bsc
   },
-  /*
   harmony: {
     tvl: harmony
   },
   tvl: sdk.util.sumChainTvls([harmony, bsc])
-  */
- tvl: bsc
 }


0xngmi:
Thanks!

0xngmi:
Seems like the subgraph is out of date:
```
 Error: Failed to decode `block.number` value: `subgraph QmQyQwXMLbZNBtvU1huhTq2aKQhAgfghuD9iyJkfxNZpiU has only indexed up to block number 7980957 and data for block number 7981036 is therefore not yet available
```

frankydev1:
@

> Seems like the subgraph is out of date:
> 
> ```
>  Error: Failed to decode `block.number` value: `subgraph QmQyQwXMLbZNBtvU1huhTq2aKQhAgfghuD9iyJkfxNZpiU has only indexed up to block number 7980957 and data for block number 7981036 is therefore not yet available
> ```

Thank you @0xngmi for the alert! Unfortunately this is an issue with the Harmony ONE Shard 0 Node, it is effecting the entire Harmony ONE network and including all Dapps using this for thegraph data.  Hopefully this will be resolved soon, and once it's resolved from Harmony ONE team it should automatically not show that error. 

Thanks

PR #192, opened 2021-06-04
Update index.js

jobsonlook:
Twitter Link:https://twitter.com/DepthFi
List of audit links if any:
https://depth.fi/static/media/Depth-Report-022721.6c156aca.pdf
https://depth.fi/static/media/Depth-Report2-EN.69622836.pdf
https://depth.fi/static/media/Depth-Report3-EN.d567c6f0.pdf
https://depth.fi/static/media/Depth-Report4-EN.a3aa7038.pdf

Website Link:https://depth.fi/
Logo (High resolution, preferably in .svg and .png, for application on both white and black backgrounds. Will be shown with rounded borders): 
https://drive.google.com/drive/folders/1rhsaMEbeByC55MalaFFR2QFnagRcfTTw?usp=sharing

Current TVL:241,639,149.50
Chain: HECO

Coingecko ID (so your TVL can appear on Coingecko): (depth-token)
Coinmarketcap ID (so your TVL can appear on Coinmarketcap): (9164)

Description/bio (to be shown on DefiLlama):
Depth is a safe and high efficient stablecoin asset management protocol, based on the HECO Chain. As a trusted stablecoin swap manager, Depth allows users earn interest on deposits and swap of stablecoins in a safe and secure environment, with good depth, low slippage, and low transaction fees.

Token address and ticker if any:  https://hecoinfo.com/token/0x48C859531254F25e57D1C1A8E030Ef0B1c895c27
$DEP

Category (Yield/Dexes/Lending/Minting): Dexes
[Optional] ETH addresss (to receive a LlamaPunk):


0xngmi:
Thanks, for the PR, left some comments on discord:
> Just had a look, would it be possible to keep the previous code? I'd prefer to have code that reads directly from on-chain data rather than an API
If vaults are something that will see a lot of changes we could use an API for that part, but I'd prefer to keep as much as possible off of APIs
The reason for this is that APIs break, they require trust and are non transparent in regards to how the TVL is calculated

jobsonlook:
> Thanks, for the PR, left some comments on discord:
> 
> > Just had a look, would it be possible to keep the previous code? I'd prefer to have code that reads directly from on-chain data rather than an API
> > If vaults are something that will see a lot of changes we could use an API for that part, but I'd prefer to keep as much as possible off of APIs
> > The reason for this is that APIs break, they require trust and are non transparent in regards to how the TVL is calculated

Ok,I will update this.

PR #195, opened 2021-06-05
Bakeryswap tvl adapter (bsc chain)

petrovska-petro:


0xngmi:
Noticed that there's very similar code used elsewhere, could you abstract that into a function that gets imported into these adapters? This will help for when we want to make changes

petrovska-petro:
> Noticed that there's very similar code used elsewhere, could you abstract that into a function that gets imported into these adapters? This will help for when we want to make changes

Alright!, will push in next commit the abstraction

PR #199, opened 2021-06-06
Polycat & Iron Finance tvl adapters

petrovska-petro:


petrovska-petro:
For some reason, in the following line https://github.com/petrovska-petro/DefiLlama-Adapters/blob/main/projects/polycat/index.js#L177 it seems not to catch up properly the balances of the LP tokens while breaking down the tvl per token0 & token1, any idea why?.

I tried to verify that it was passing indeed a correct lp address from the polygon network and some balance, and it was sending them, but still not summing them up on the balances variable at the end.

PR #203, opened 2021-06-08
Update alchemix adapter to include new staking pools

0xdef1:
Add new staking pools for Alchemix. 

- Dual rewards for ALCX-ETH from MasterChef v2
- Dual rewards for alusd-3crv from Curve gauge


0xngmi:
Thanks ser!

PR #211, opened 2021-06-10
Frax finance tvl adapter

petrovska-petro:


projects/fraxfinance/index.js
```diff
@@ -0,0 +1,129 @@
...
+    });
+  }
+
+  const transformAdress = await ethAddr();
+
+  await unwrapUniswapLPs(
+    balances,
+    lpPositions,
+    block,
+    "ethereum",
```
zashton, line 117: the 2 final parameters might not be necessary if its just only just ethereum tokens, its mainly for porting token address from side chain to ethereum address

petrovska-petro: You are completely right, just removed in the last commit. Thank you

PR #218, opened 2021-06-11
Add missing vault for alpaca finance & fix iron finance tvl

petrovska-petro:


projects/helper/unwrapLPs.js
```diff
@@ -82,12 +82,18 @@ async function unwrapUniswapLPs(balances, lpPositions, block, chain='ethereum',
         const token0 = (await tokens0).output.find(call=>call.input.target === lpToken).output
         const token1 = (await tokens1).output.find(call=>call.input.target === lpToken).output
         const supply = (await lpSupplies).output.find(call=>call.input.target === lpToken).output
-        const {_reserve0, _reserve1} = (await lpReserves).output.find(call=>call.input.target === lpToken).output
+        // -- Add try/catch block in this section, as sometimes some lp do not have _reserve0, _reserve1 ---
```
0xngmi, line 85: Which LPs don't have these methods?

petrovska-petro: I think in one of the scripts I run, it had lp's from balance and 1inch, which were giving problems. I think those have not this method. Not in this particular script, but another. But adding that protection there, avoids probably the script to stop in the other scenarios, until another helper is built probably to handle balancer/1inch pools data

0xngmi: It might be better to leave as it is. The problem is that sometimes RPC calls error out not because there's a problem with the call but because the service is down or it fails that call. If we allow failures on things that should be working then the adapter will ignore the failed calls and compute a TVL that is wrong (this has happened multiple times already)

0xngmi:
Could you fix merged conflicts?

projects/charmfinance/index.js
```diff
@@ -0,0 +1,89 @@
...
+
+const USDT = "0xdac17f958d2ee523a2206206994597c13d831ec7";
+const USDC = "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48";
+const WETH = "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2";
+const WBTC = "0x2260fac5e5542a773aa44fbcfedf7c193bc2c599";
+const ETH = "0x0000000000000000000000000000000000000000";
+
+const VAULT = "0x55535C4C56F6Bf373E06C43E44C0356aaFD0d21A";
+const CUBE_POOL = "0x23F6A2D8d691294c3A1144EeD14F5632e8bc1B67";
+const OPTIONS_CONTRACTS = [
```
0xngmi, line 14: Any way to get new addresses from github/api/smth?

petrovska-petro: Now it is using directly the yaml file from their github, done by the last commit

0xngmi:
Went ahead and merge the charm finance adapter manually, thanks!

PR #220, opened 2021-06-14
update Iron Finance TVL data

oramis1008:


0xngmi:
Thanks! We were already working on fixing your TVL but I hadn't merged that PR yet since there were some merged conflicts.

0xngmi:
Btw, I saw that you've included some pool2 LP tokens, we generally don't include these so later I'll go through this and propose a few changes. However since this is quite high priority I'll go ahead and push this live asap.

oramis1008:
Thank you. Please let us know what should be changed and we will update it accordingly.

0xngmi:
For other projects we avoid including pool2 LPs (LPs with the governance token) on the TVL, so in the case of Iron I imagine this would mean removing all the LPs of TITAN and STEEL

PR #221, opened 2021-06-14
Yfdai

yfdaifinance:


0xngmi:
Doesn't this calculate the marketcap of yfdai instead of the TVL?

yfdaifinance:
Modified mechanism to calculateTVL.
YFDAI has 4 farms and 3 staking contracts.

projects/yfdai/index.js
```diff
@@ -0,0 +1,42 @@
...
+const abi = require('./abi');
+
+async function fetch() {
+
+    const yfdaiTokenAddress = "0xf4CD3d3Fda8d7Fd6C5a500203e38640A70Bf9577";
+
+    const YfDaiETHLP72HRSAddress = "0x75E9F410e8d1D7240b67ec6FE35FA37580b814d9";
+    const YfDaiETHLP30DayAddress = "0x8D704D4107CBE5ebE8c0236C5506b30Bf8Bad305";
+    const YfDaiETHLP60DayAddress = "0x26572bf2620108cb5006987e6348c07dc4e14a0f";
+    const YfDaiETHLP90DayAddress = "0x175d6cbaeff93734ada4c5430815f2208a6b040c";
```
0xngmi, line 18: The yfdai on these contracts seem to be rewards for the LPs staked within, is that correct?

projects/yfdai/index.js
```diff
@@ -0,0 +1,42 @@
...
+    const YfDaiStakingAdddress = "0x44d771D0C998f524ff39aB6Df64B72bce1d09566";
+    const YfDaiSafetradeStakingAddress = "0x4599cDa238Fb71573fd5A0076C199320e09BCfF0";
```
0xngmi, line 20: Could you explain what is this contract?

projects/yfdai/index.js
Yara1990: Yes, that's correct. 
All these are YFDAI Farms, Where users can stake LPs and get YFDAI in rewards.

projects/yfdai/index.js
Yara1990: YfDaiETHLP60DayAddress:  Farm for Uniswap LP tokens(with 60 days locking)
YfDaiETHLP90DayAddress: Farm for Uniswap LP tokens (with 90 days locking)
YfDaiStakingAdddress: Staking contract(Users can stake YFDAI and get YFDAI as rewards)
YfDaiSafetradeStakingAddress:  SafeTrade is a bot where users can deposit YFDAI and the Bot can use this to trade in markets.

projects/yfdai/index.js
0xngmi: For other protocols we don't count the money used as rewards since those tokens are not even part of the circulating supply and have not been locked on a contract by anyone. To keep comparisons fair I believe it would be best not to include it here either.

projects/yfdai/index.js
0xngmi: Given that the amount of tokens in staking heavily dwarfs the amount of tokens on safeTrading I believe it would be best to exclude staking for now, since no other platform we have on defillama has this much staking

Yara1990: Right now,
Staking has ~11816 YFDAI tokens 
SafeTrade has ~ 157 YFDAI tokens
So you mean You won't be counting Staking tokens for TVL?
It should be counted along with all the farms right?

0xngmi: Yes, I believe it's best not to count staking to keep in line with what we count in other projects

Yara1990: But, 11816 YFDAs are a major chunk of our supply, and it's locked, so it should be counted in TVL.
If there is some technical challenge right now, then let's go ahead and list YFDAI for now, and then you can figure out about staking tokens later.

Yara1990: When can we expect this pull request to be merged?

0xngmi: Yes, I went ahead and wrote an alternative version of the adapter on https://github.com/DefiLlama/DefiLlama-Adapters/blob/main/projects/yfdai/index.js, let me know what you think

Yara1990: I think it's good to go. Please merge it to master/main.

0xngmi:
Already merged, will close this now then

projects/yfdai/index.js
yfdaifinance: @0xngmi Even after the YFDAI Project is merged in main, we are not able to see the token in the application.

PR #224, opened 2021-06-15
[Saddle] Update and add alETH, sETH tokens and WETH/alETH/sETH pool

penandlim:
At the time of writing, WETH-alETH-sETH pool is at $4.6MM. Our total TVL should be around $60-70MM

However since coingecko has not added ~~alETH~~ or vETH2 to their token list yet, we expect the adapter to report TVL slightly lower.

### DIFF
--- a/projects/saddle/index.js
+++ b/projects/saddle/index.js
@@ -13,6 +13,7 @@
   const btcPoolAddress = "0x4f6A43Ad7cba042606dECaCA730d4CE0A57ac62e"
   const usdPoolAddress = "0x3911f80530595fbd01ab1516ab61255d75aeb066"
   const veth2PoolAddress = "0xdec2157831D6ABC3Ec328291119cc91B337272b5"
+  const alethPoolAddress = "0xa6018520eaacc06c30ff2e1b3ee2c7c22e64196a"
 
   const tokens = {
     // TBTC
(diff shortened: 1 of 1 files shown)


penandlim:
Oh looks like coingecko has price feed for alETH! 
https://www.coingecko.com/en/coins/alchemix-eth

Still missing vETH2 though :( 
https://www.coingecko.com/en/coins/veth2

0xngmi:
Seeing how vETH2 is taking a while to get added I'm thinking of just replacing it with ETH on this adapter. I'm more concerned about getting the total TVL to be correct than making sure that the details the breakdown are accurate. Thoughts?

0xngmi:
Pushed an update that calculates vETH2 value as `WETH on the pool * 2`. I think this is currently undervaluing vETH2 value so once you have some time (I imagine you are busy with the recent events, no hurry) it would be awesome if you could take a look at the new code.

penandlim:
Yeah I think this makes sense to me. Since the pool is still heavily imbalanced, iI think it makes sense to assume 
`value of the pool = (amount of WETH ) * 2`
Thanks for the update!

0xngmi:
Got it, thanks for the quick turnaround and constant updates :smile:

PR #239, opened 2021-06-21
update wasabix cross-chain tvl

wasabi-swap-team:


0xngmi:
As mentioned on discord, I'd rather keep the adapter based on onchain data for transparency reasons

wasabi-swap-team:
@0xngmi Hi there, we have updated the adapter using on-chain data, please check it again, thanks!

PR #243, opened 2021-06-23
fix pangolin data query

dhrubabasu:
Currently, it's fetching this USDT token: https://info.pangolin.exchange/#/token/0x13d0d85fe3534e02c30ad87bc55a3f361e358c09

To not hardcode the correct USDT token, I changed the query to fetch the highest liquidity token with the symbol `USDT`.


0xngmi:
Could you hardcode the correct USDT token? I think that may be better

PR #250, opened 2021-06-25
add alusd and reth to convex.  Update tricrypto swap and lp token addresses

C2tP-C2tP:
This request adds reth and alusd pools for Convex Finance, as well as correct tricrypto's addresses.

### DIFF
--- a/projects/convex/pools-crv.js
+++ b/projects/convex/pools-crv.js
@@ -1159,19 +1159,81 @@ module.exports = [{
               name: '3CrvCrypto',
               symbol: '3CrvCrypto',
           },
-          assets: 'usdc+weth+wbtc',
+          assets: 'usdt+weth+wbtc',
           coins: [
-              coins.usdc,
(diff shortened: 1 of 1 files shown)


0xngmi:
Thanks, what do you think is the best way to keep it up to date in the future?

C2tP-C2tP:
> Thanks, what do you think is the best way to keep it up to date in the future?

i think leveraging Curve's registry system would be the easiest.  Havent had time to whip out a quick test script though.  It should be something like this

foreach pool in convex.poolinfo
- supply of pool.token (the deposit token) is how much lp tokens convex has
- get total supply of curve lp tokens -> calc share of pool
- curveRegistry.get_pool_from_lp_token(pool.lptoken) -> get swap pool
- curveRegistry.get_underlying_coins(swapPool) -> get underlying coin list
- curveRegistry.get_underlying_balances() -> get underlying coin balances
- use lp share to determine convex balances

PR #251, opened 2021-06-25
Create index.js for cardstarter

mafaiq:

### DIFF
--- a/projects/cardstarter/index.js
+++ b/projects/cardstarter/index.js
@@ -1,0 +1,44 @@
+const sdk = require("@defillama/sdk");
+const { unwrapUniswapLPs } = require('../helper/unwrapLPs');
+
+const cardsStakeAddr = "0xad2fd18932c39fa5085429853e1f0d39a65a438e";
+const lpStakeAddr = "0x7Dca3372A0a236A305FdEC3D48d52B09dff82E14";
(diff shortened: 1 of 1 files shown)


0xngmi:
Since we don't count pool2 towards TVL for other protocols should I remove `0x94ae6d2390680ac6e6ee6069be42067d6ad72e2a` from the array?

mafaiq:
What do u mean by pool2 ser?

You can stake cards/eth LP on cardstarter platform for cswap rewards.

So that should be normal farming, right?

Sorry im not on discord atm

0xngmi:
pool2 = the LP pair where one of the tokens is the governance token

> You can stake cards/eth LP on cardstarter platform for cswap rewards.

For other protocols we are not counting this towards the TVL, you could argue that this is TVL since it's money locked but the general consensus seems to be that this shouldn't be counted as TVL since it's nothing specific to the protocol

mafaiq:
I see, i'll look into it later tonight to find out whether thats for
governance or not.

But afai understood, thats a farm to mint more cswap tokens,

And the the cards staking pool is to take part in the IDOs.

0xngmi:
Isn't cards their own token?

mafaiq:
Yeah, but i meant they have multiple reasons to stake their token on the
platform.

Mainly as a way to be eligible for partaking in IDOs

PR #254, opened 2021-06-26
Update Liquity TVL: include Stability Pool deposits

edmulraney:
The Stability Pool is used to pay off bad debt in the system (undercollateralized loans) this deposited LUSD is held in the protocol and provides economic function.


0xngmi:
Thanks!

PR #260, opened 2021-07-01
add Balancer V2 support

TomAFrench:
This PR pulls TVL info from Balancer V2 as well as the new Polygon deployment

### DIFF
--- a/projects/balancer/index.js
+++ b/projects/balancer/index.js
@@ -1,8 +1,10 @@
+const sdk = require('@defillama/sdk');
+const BigNumber = require("bignumber.js");
 const { toUSDTBalances } = require('../helper/balances');
 const { GraphQLClient, gql } = require('graphql-request')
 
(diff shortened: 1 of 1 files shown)


projects/balancer/index.js
```diff
@@ -17,15 +19,27 @@ async function tvl(timestamp, ethBlock) {
...
   const results = await graphQLClient.request(query, {
-    block: ethBlock
+    block
   })
-  return toUSDTBalances(results.balancers[0].totalLiquidity);
+  return results.balancers[0].totalLiquidity;
+}
+
+async function ethereum(timestamp, ethBlock) {
+  const [v1,v2] = await Promise.all([getTVL("balancer", ethBlock), getTVL("balancer-v2", ethBlock)])
```
0xngmi, line 28: I see, so that's the right way to compute the TVL, was a bit confused earlier when trying to figure it out

https://twitter.com/0xngmi/status/1409167935088828417

PR #262, opened 2021-07-02
add DMM(polygon) to kyber

ssreelekhya:


projects/kyber/index.js
```diff
@@ -100,26 +101,50 @@ query get_tvl($block: Int) {
...
+      sdk.util.sumSingleBalance(balances, usdtAddress, ethTVL)
     }
 
     return balances;
   }
+  async function polygonTvl(){
+    var polygonEndpoint ='https://api.thegraph.com/subgraphs/name/piavgh/dmm-exchange-matic';
+      var query = gql`
+      {
+        dmmFactories(first: 1) {
```
0xngmi, line 122: I would prefer if you could pin this to the block number being passed as a parameter. There's two reasons for this:
- It allows us to get historical data correctly
- On polygon and other evm chains with high throughput it's common that applications that see a lot of traffic end up with outdated subgraphs (speed of updates is higher than the speed at which the subgraph can process them, so it stays constantly outdated). By pinning responses to a blockheight we make sure that, if the subgraph hasn't catched up, the response fails, thus preventing any misreporting.

Regarding the 2nd point I also wanted to ask if your polygon subgraph is just a copy of the ethereum one or it has been made lighter, as otherwise it's likely that this will happen and this code will need to be replaced with something else that reads data from onchain calls.

projects/kyber/index.js
```diff
@@ -100,26 +101,50 @@ query get_tvl($block: Int) {
...
 ==================================================*/
 
   module.exports = {
-    ethereum: tvl,
-    start: 1546515458,  // Jan-03-2019 11:37:38 AM +UTC
-    tvl,
+
+    ethereum: {
+      tvl: ethTvl,
+    start: 1617606651, // @5-Apr-2021, 7:10:51 AM+UTC
```
0xngmi, line 144: `start` should be outside the ethereum object, although you could also remove it

package.json
```diff
@@ -27,7 +27,7 @@
     "node-html-parser": "^3.3.5",
     "node-schedule": "^1.3.3",
     "papaparse": "^5.3.1",
-    "web3": "^1.3.6"
+    "web3": "^1.4.0"
```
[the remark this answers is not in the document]
ssreelekhya: just ran an update , no intention behind this. Will revert the update.

projects/kyber/index.js
ssreelekhya: Sure, to include block number makes sense.
And yes, it's copy of ethereum.

projects/kyber/index.js
ssreelekhya: ok

projects/kyber/index.js
ssreelekhya: changes done. Please review and help merge

PR #274, opened 2021-07-09
Mislabeled contracts

waynebruce0x:
Tiny PR because the yield token holder addresses were mixed up!


0xngmi:
Yeah I noticed, it was wrong on their github and given that it didn't really matter I just left it as is, still good to fix :)

PR #282, opened 2021-07-12
use totalLiquidityUSD

dhrubabasu:
We've recently updated the subgraph to accurately display `totalLiquidityUSD`: https://info.pangolin.exchange/


0xngmi:
If I merge this PR won't the USD be interpreted as AVAX (since the object returned is in avax)?

PR #284, opened 2021-07-13
New IronFinance TVL update

oramis1008:
We'd like to update our TVL data after our relaunch on July 12th.


0xngmi:
Thanks! Just did a light review since yours is a big protocol and want to list it quickly to keep total polygon TVL up to date. Will do a more in-depth review later and share anything I find here

oramis1008:
Thank you! When will you deploy the updates since I can't find Iron on DefiLlama yet

PR #294, opened 2021-07-15
add solyard

jianfengxiaojianfeng:
##### Twitter Link: 
https://twitter.com/SolyardFinance

##### Website Link: 
https://solyard.finance/

https://solyard.finance/logo.png

##### Current TVL:
$13,253,888.78

##### Chain:
Solana

No token yet

No token yet

Solyard is New Yield Farming Aggregator Optimizer protocol on #Solana

No token yet

Yield

#### [Optional] ETH addresss (to receive a LlamaPunk):

### DIFF
--- a/projects/solyard.js
+++ b/projects/solyard.js
@@ -1,0 +1,12 @@
+const retry = require('./helper/retry')
+const axios = require("axios");
+
+async function fetch() {
+  var response = await retry(async bail => await axios.get('https://solyard.finance/tvl'))
+
+  return response.tvl || 0;
+}
+
+module.exports = {
+  fetch
+}
--- a/projects/solyard.js
+++ b/projects/solyard.js
@@ -4,7 +4,7 @@ const axios = require("axios");
 async function fetch() {
   var response = await retry(async bail => await axios.get('https://solyard.finance/tvl'))
 
-  return response.tvl || 0;
+  return response.total || {};
 }
 
 module.exports = {


0xngmi:
A few minor nitpicks:
- Unless in a few cases where the protocol is built around LPs that contain the governance token (for example FEI) we tend to exclude pool2 TVL
- We are currently working on separating staking TVL from other sources of TVL

Could you modify the API so that TVL is separated by source? One way to do that would be to make an API that returns these 3 tvls separated or another would be to make an API that returns the TVL by pool, both would work for us

jianfengxiaojianfeng:
> A few minor nitpicks:
> 
> * Unless in a few cases where the protocol is built around LPs that contain the governance token (for example FEI) we tend to exclude pool2 TVL
> * We are currently working on separating staking TVL from other sources of TVL
> 
> Could you modify the API so that TVL is separated by source? One way to do that would be to make an API that returns these 3 tvls separated or another would be to make an API that returns the TVL by pool, both would work for us

solved

0xngmi:
Thanks, that was quick

0xngmi:
Will merge and then apply some changes to separate TVLs

0xngmi:
When executing the code I'm getting the following error:
```
 UnhandledPromiseRejectionWarning: Error: unable to verify the first certificate
 ```
 
 Could there be a problem with the certificate chain?

0xngmi:
@jianfengxiaojianfeng we can't use this adapter due to the problem explained above

jianfengxiaojianfeng:
> @jianfengxiaojianfeng we can't use this adapter due to the problem explained above

solved

0xngmi:
There seems to be a difference between the tvl reported by your API and the one you report on your website, a pretty big difference actually (about 10x), could you please update the adapter?

PR #297, opened 2021-07-16
Add SafeDollar TVL tracker

safedollar-finance:


0xngmi:
Would it be possible for you to develop an adapter that gets the TVL from on-chain calls instead of calling your API? This helps with both transparency and lets us get historical data

PR #300, opened 2021-07-17
Charm finance update tvl adapter

petrovska-petro:
New vaults v1 added, also handled token0 & token1 for each vault more dynamically via ABI's method


0xngmi:
Are these vaults their new product offering (the managed uni v3 liquidity)?

PR #302, opened 2021-07-17
Add hBTC and renBTC pool for Saber

michaelhly:
Hey @realdealshaman, thank you for adding us. 

I was thinking about exposing our own API for the TVL metric since we tend to add a lot of assets. @0xngmi, do you care if the data comes from a closed source API?

### DIFF
--- a/projects/saber.js
+++ b/projects/saber.js
@@ -1,18 +1,21 @@
 const {getTokenBalance} = require('./helper/solana')
 
 async function tvl() {
-    const [usdcAmount, usdtAmount, paiAmount, usdcAmount_2, btcAmount, renBtcAmount, pbtcAmount, renBtcAmount_2, ustAmount, daiAmount, busdAmount, fraxAmount, usdcAmount_3, wlunaAmount, renLunaAmount] = await Promise.all([
-        //usdc-usdt pool
+    const [usdcAmount, usdtAmount, paiAmount, usdcAmount_2, btcAmount, renBtcAmount, pbtcAmount, renBtcAmount_2,
+    hbtcAmount, renBtcAmount_3, ustAmount, daiAmount, busdAmount, fraxAmount, usdcAmount_3, wlunaAmount, renLunaAmount] = await Promise.all([
+        //usdc-usdt pool/Users/michaelhuang/Projects/DefiLlama-Adapters/projects 
         getTokenBalance("EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v", "5C1k9yV7y4CjMnKv8eGYDgWND8P89Pdfj79Trk2qmfGo"),
         getTokenBalance("Es9vMFrzaCERmJfrF4H2FYD4KCoNkY11McCe8BenwNYB", "5C1k9yV7y4CjMnKv8eGYDgWND8P89Pdfj79Trk2qmfGo"),
         //pai-usdc pool
         getTokenBalance("Ea5SjE2Y6yvCeW5dYTn7PYMuW5ikXkvbGdcmSnXeaLjS", "7W9KMACQT6UmjRPEUQKXyVf4NjZ9Ux4PHs1e1P5PxDtA"),
         getTokenBalance("EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v", "7W9KMACQT6UmjRPEUQKXyVf4NjZ9Ux4PHs1e1P5PxDtA"),
-        //BTC-renBTC & pBTC-renBTC pool
+        //BTC-renBTC & pBTC-renBTC & hBTC-renBTC pool
         getTokenBalance("SBTCB6pWqeDo6zGi9WVRMLCsKsN6JiR1RMUqvLtgSRv", "Fekck54VF2MdesR74trJteZbiKj1TD5AVQisXr8E7fjG"),
         getTokenBalance("CDJWUqTcYTVAKXAVXoQZFes5JUFc7owSeq7eMQcDSbo5", "Fekck54VF2MdesR74trJteZbiKj1TD5AVQisXr8E7fjG"),
         getTokenBalance("DYDWu4hE4MN3aH897xQ3sRTs5EAjJDmQsKLNhbpUiKun", "2wszCpUdVDFrJcP79wpV3FdBmU38UC1YKuoSUBA5mhWu"),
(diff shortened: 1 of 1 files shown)


0xngmi:
I started writing a short answer to this but ended up spending a few hours to fully articulate my position.

-----

> @0xngmi, do you care if the data comes from a closed source API?

I would rather avoid that, mostly because if data comes from a closed API it's not as transparent as if you can check the code that computes the TVL.

This might not seem like it matters much but this transparency is one of our core tennets at defillama and one of the things I spend a lot of time working on. Here are the reasons for it

#### Verifiability and reproducibility
It's important for our metrics to be verifiable and reproducible, meaning that anyone should be able to check how our TVLs are calculated, what's included as part of them, and then be able to reprodcue that data that we display on our website. For this reason we keep our whole stack open source, to make it possible to reproduce everything.

With API-based adapters that's just not possible since they are completely opaque and any changes to them are not tracked, whereas changes to our scripts are tracked through git. One of consequences of that is that verifying a historical TVL value obtained through an API requires a lot of work.

This is because a lot of the tools you would use to verify the current TVL (eg: by checking the tokens locked on a contract on etherscan, or querying some contracts through etherscan) don't allow you to check at older blockheights, requiring you to write a script for it instead. This already makes this verification impossible for a large part of the population, and for the other part it makes it heavily work-intensive.

#### Fairness
One of the main things that defillama has to provide is fair comparisons between protocols. It would be a failure on our side if for one protocol we include assets on a staking contract and for another protocol we don't include them, since we would be favoring a protocol over another. However, there's a lot of things to consider when choosing what to include as TVL so we are bound to make mistakes, so transparency here acts as a safeguard that allows other people to keep ourselves in check.

By making it possible to easily check how other TVLs are calculated, people can verify whether we count TVL in a similar way to the way we are counting it for them, and if we don't or if we include something for another project that we are excluding for them, they can raise the issue publicly.

If adapters rely on a closed API it's much harder and more time-intensive to check what is being included (especially since that may change through time, eg: in the past an API might include something as TVL that is not included anymore, and while checking current TVL is somewhat doable, crosschecking historical TVL against a number is very time intensive, a sisyphean task when you take into account that we host almost 300 projects), so it would be much harder or straight up impossible to keep ourselves in check (and not just ourselves, when data comes from an API).

#### Breaking
APIs break a lot. The very few adapters we have that use APIs do regularly break and then I have to scramble to fix since otherwise data gets stale and we miss updates, which is quite important, since usually APIs don't provide historical values, so if it takes me 2 days to fix that adapter then there will be a 2-day gap in our data.

On top of that, it's impossible for me to fix APIs, so all I can do is just ping the team to let them know that their API is broken or try to rewrite their adapter to avoid using their API.

#### No control over changes
The problem with APIs is that, even if we verify that the numbers are correct the moment we integrate them, we have no control over them and it's impossible to modify these numbers afterwards. An example of this would be the case of sovryn, where initially the API only included basic TVL stuff like liquidity on their pairs, resulting in a TVL of ~50M, but then they changed their API to also include staked tokens, which pumped the TVL to almost 2bn.

This generated quite a lot of drama on twitter (especially since sovryn claimed that their TVL had surpassed uniswap) and casted doubt on the fairness of our metrics, as it turned out that "staked tokens" were more like unvested tokens that hadn't been issued yet, so they were counting part of their FDV as TVL.

This is a bit of an extreme example, but it highlights a real problem with APIs. Even if changes are not like the one above, if suddenly something is added to the TVL through the API, it can lead to a jump on the TVL that is not real since TVL hasn't actually increased, it just that a new data point was added.

#### Historical data
This only applies to chains where it's possible to query values at previous blockheights (EVM chains and Terra fit here but Solana doesn't), but usually APIs only provide the most current value, which is a problem since:
- If we miss some API updates we'll have gaps on our data
- We can only start tracking TVL once the adapter has been integrated, it's not possible to get historical data, so our charts will look chopped, and initially there will be very few data points

Instead with on-chain adapters we can run them at old blockheights and get the TVL at different points in time, allowing us to fill our charts.

### Snowball effect
Above I listed the fundamental reasons why we avoid this kind of adapters, but another secondary reason why we avoid them is that accepting API-based adapters creates a snowball effect where other people ask us to also accept their API.

This mainly manifests in other people asking us why we won't merge their API adapter when we have merged other API adapters, thus accusing us of favoring some projects over others.

As you can imagine, this just grows bigger, as the more projects we merge with an API-adapter the less reasons we have to reject other API adapters, creating a negative feedback loop, that forces us to either merge even more API adapters or reject them and look like dicks.


## Exceptions
After this long ramble about how important this is, it's time to destroy my own image by explaining how we've skipped this rule multiple times in the past.

The truth is that, while I really want to avoid any kind of reliance on closed APIs on our adapters, there's multiple occasions where we've had to make compromises. I've tried to compile all the cases where we've merged adapters that use APIs:
- Historical cases: Before we developed much of our infrastructure (eg: our sdk for on-chain calls), we were much more liberal in what we accepted as an adapter, and at that time we merged a number of adapters that were based on opaque APIs. Over time I've been replacing them but there's still a lot of these legacy adapters around, actually most of the API adapters fall into this category.
- Yearn (no bandwith): In the case of yearn we purposefully chose to use their API because we didn't have the bandwith to build an on-chain adapter ourselves. Yearn has used a bunch of different vaults through their history and tracking everything down was a lot of work, especially considering that, to build the current API banteg, who's already quite familiar with everything, spent 2 weeks working on it, so for us to rebuild it we could spend a month, and afterwards we would have to keep it up-to-date with all the vaults that get added. With our resources at the time (just me) it didn't make sense to spend 1 month just to list a single protocol when the API was already working so we just used the API. Since then we've applied the same reasoning to other protocols where either it's really hard to build an on-chain-based adapter or keeping everything up to date is really hard.
- Backlog: Recently we accumulated quite a backlog of Pull Requests to review since I spent some time inactive (was ill) and, in order to clear up the backlog, I merged some PRs that used APIs. In the past it has been quite hard to get teams to understand that we want an on-chain adapters instead of one that uses their API, even after exposing our arguments around transparency and historical data fetching (I partially get them, defi teams are usually really busy and then suddenly we ask them to build something extra that isn't really needed since we could just use their API). A solution I've used before is to just implement an on-chain adapter for them, since that just solves all the problems. However if bandwith is short that becomes harder.
- Special: There was this one particular case where a team submitted an API-based adapter and then messaged me every day asking when it would be merged. I tried to verify their TVL by writing an adapter for them that used on-chain calls but, even after rewriting it three times, I couldn't manage to match their numbers. After some time receiving their messages I caved in and merged their API adapter, since their TVL was quite small and the difference between my numbers and theirs is also small. Still, I'm not happy with it and probably wouldn't do it again.

0xngmi:
If you think that keeping the pool list up to date through PRs will require a lot of work we could also use an API where you return:
- pool contracts
- token in the pool
- coingecko id of the token

Issue #304, opened 2021-07-18
Why we don't accept APIs


0xngmi:
A lot of projects ask us to use their closed API to get their TVL. I would rather avoid that, mostly because that's much less transparent than an open-source script that computes the TVL from on-chain calls.

This might not seem like it matters that much, but transparency is one of the core tenets of defillama and one of the things I spend most of my time working on. Here are the reasons for it:

#### Verifiability and reproducibility
It's important for our metrics to be verifiable and reproducible, meaning that anyone should be able to check how our TVLs are calculated, what's included as part of them, and then be able to reproduce the data that we display on our website. For this reason, we keep our whole stack open source, making it possible to reproduce everything.

With API-based adapters that's just not possible since they are completely opaque, making it completely impossible to check how the TVL is calculated.

Furthermore, one of the non-obvious consequences of that is that any changes to the API are not tracked, whereas changes to our scripts are tracked through git. The result of that is that verifying historical TVL values obtained through an API requires a lot of work.

This is because a lot of the tools you would use to verify the current TVL (checking the tokens locked on a contract or calling a contract function through etherscan) don't allow you to make queries at older blockheights, requiring you to write a script for it instead. This makes verification impossible for a large part of the population, and for the other part it makes it heavily work-intensive.

#### Fairness
One of the main things that defillama must provide are fair comparisons between protocols. It would be a failure on our side if for one protocol we count assets on a staking contract and for another protocol we don't include them, since then we would be favoring a protocol over another. However, there's a lot of things to consider when choosing what to include as TVL so we are bound to make mistakes, transparency acts as a safeguard that allows other people to keep us in check.

By making it possible to easily check how other TVLs are calculated, people can verify whether we count TVL for other projects in a similar way to the one we are using for them, and if we don't or if we include something for another project that we are excluding for them, they can raise the issue publicly.

If adapters rely on a closed API it's much harder and more time-intensive to check what is being included (especially since that may change through time, eg: in the past an API might include something as TVL that is not included anymore, and while checking current TVL is somewhat doable, crosschecking historical TVL against a number is very time intensive, a sisyphean task when you take into account that we host almost 300 projects), so it would be much harder or straight up impossible to run these comparisons against other protocols.

#### Breaking
[...]

PR #321, opened 2021-07-26
Add oraichain to project

hieunv25:
- Ticker ORAI, address on Ethereum : 0x4c11249814f11b9346808179cf06e71ac328c1b5
- Ticker:ORAI, address on bcc : 0xa325ad6d9c92b55a3fc5ad7e412b1518f96441c0

### DIFF
--- a/projects/oraichain/index.js
+++ b/projects/oraichain/index.js
@@ -1,0 +1,32 @@
+const utils = require("../helper/utils");
+
+async function fetchEth() {
+  const { data: totalTvl } = await utils.fetchURL(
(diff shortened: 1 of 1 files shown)


0xngmi:
Hey, just had a look at your project and it seems like almost all your TVL comes from either staking your native token or staking the pool2 LP token. We are trying to separate this so our users can choose what to include on the TVL, so that would need to be applied to the PR.

Furthermore, given that the number of your vaults is small and they are not complicated I think it'll be better to base the adapter off on-chain calls instead of your API. The rational for this is explained on #304

waynebruce0x:
Closing this because I made an on chain version

PR #335, opened 2021-07-30
Telcoin/TELx Network

rileyhoffman:
tel logo blue big

##### Description/bio (to be shown on DefiLlama): TELx is a decentralized liquidity network powered by active Telcoin users on decentralized financial (DeFi) protocols. It functions as the self-custodial “liquidity engine” of the Telcoin Platform’s user-owned products.

The goal of TELx is to offer intuitive and affordable financial services, where user ownership of the network enables any mobile phone user in the world to capture incentives that would typically be awarded to legacy financial institutions.

Traditional financial markets involve intermediaries and institutions that custody, execute, and settle transactions on behalf of their customers, often charging exorbitant fees and suffering from slow settlement times. TELx Markets are self-custodial, powered by a decentralized network of Telcoin users (TELx Miners) on DeFi protocols, and the transaction lifecycle is automated via smart contracts instead of intermediaries - with instant settlement.


0xngmi:
I think there's multiple problems with this PR:
- I'd like to avoid any modifications to shared files such as package-lock.json
- Not sure why you added an `id` parameter to the tvl functions but that breaks them since the chainBlocks parameter is not caught properly and you end up with the error `TypeError: Cannot read property 'polygon' of undefined`
- Even after fixing the previous error I get the following error `TypeError: Cannot destructure property 'uniswapFactory' of 'undefined' as it is undefined.`. The problem here is that you are sending the address of an LP token to the subgraph as a factory, so it breaks because it's not a factory.
- Assuming that your objective was to count all the tokens stored in those uniswap addreses as TVL, what we would be doing would be to simply count the pool2 TVL of TEL, which is not something that we consider as TVL for other projects.
- Furthermore, those tokens are not even locked on tel's protocol, instead they are within uniswap

Does TEL have any other sources of TVL that we could look into?

rileyhoffman:
@0xngmi I believe the added `id` came from some code refactoring where I meant to remove it. I am aiming to show the value of the Quickswap LP Token addresses shown here, https://www.telx.network/about/addresses-and-contracts, with the TEL/USDC pool shown individually, and the rest shown with combined total value. 

I was under the impression that using uniswapFactory was the correct entity to use for this, what should I actually be using to access this? I'm admittedly a little confused on whether I should be calling the Quickswap or Uniswap API as Quickswap's github lists the quickswapFactory function, whereas their subgraph lists uniswapFactory.

I will clear up the shared file modifications as well, and we have other TVL sources, but these are the ones we are focused on for now. 

Please let me know if you have any other questions/suggestions, thanks.

0xngmi:
If it's just getting the liquidity from a pair there's no need to query the subgraph, you could make direct blockchain queries to get the data you want. You could either make direct balanceOf() queries against the address that hosts the swap contract if you want to count all the liquidity there or you could also just use the function unwrapLPs() to get the tokens locked inside a fixed amount of LP tokens.

However, the main issue here are my lasts two points, namely that liquidity of your coin on another exchange is something that we've never counted as TVL and I don't see a good argument for doing so

rileyhoffman:
Okay, we are just looking to do what is needed to get listed on Zapper, as they require an adapter for us to get listed, do you know if this mandates a TVL? Or could we achieve this just by querying direct blockchain liquidities? 

We are simply trying to get staking on our platform (TELx and SMS networks) to show up on Zapper.

rileyhoffman:
@0xngmi any ideas as far as the above? If not, could you maybe point me toward someone who could help us out?

0xngmi:
Wrote the code on 928f3492cc7fef8226cda7562bee63f4cbca3ba5

0xngmi:
Following our conversations on discord it seem like this has already been resolved, is there something else you would like to discuss or can I close this PR?

rileyhoffman:
Yes, thanks so much for all the help!

_____________________________
Riley Hoffman
Software Engineer (Blockchain and Integrations) at Telcoin
<http://www.telco.in/>
Phone: +1 760 814 7692
Linkedin <https://www.linkedin.com/in/riley-hoffman-92362a13a/>

PR #336, opened 2021-07-31
[The Grand Banks] add thegrandbanks.finance

narwhalosaurus:


0xngmi:
Please use on-chain calls instead, check #304 for an explanation

narwhalosaurus:
> Please use on-chain calls instead, check #304 for an explanation

Our project is vault like autofarm or beefy so that complicated to built by on-chain calls data to the adaptor.

andres290487:
@0xngmi #9302 0x7F5c764cBc14f9669B88837ca1490cCa17c31607

PR #338, opened 2021-08-01
[The Grand Banks] added the grand banks (revise to use call on-chain data instead)

narwhalosaurus:


0xngmi:
Saw your comment earlier about being an aggregator and the API, if the amount of new vaults keeps increasing heavily and doing PRs involves too much work we could fetch a list of them from your API. If the vaults are also different enough that this also involves doing a lot work to accommodate each one then we can just directly switch to a TVL provided by your API directly

realdealshaman:
Hello @narwhalosaurus , thanks for contributing to Defillama. Your LlamaPunk has been sent: https://opensea.io/0x6abfae833ca1a4e9a9d482dfeebcc5ecec635e90

PR #345, opened 2021-08-04
b.protocol v2

yaronvel:


projects/bprotocol/index.js
```diff
@@ -149,13 +153,58 @@
...
+            block,
+            target: stabilityPoolAddress,
+            params: [bamm.output],
+            abi: abi["getCompoundedLUSDDeposit"]
+          });
+
+          totalBalance = totalBalance.plus(new BigNumber(balance.output));
+      }
+      catch {
+        break;
```
0xngmi, line 181: Is there any way we can get the length of `bamms`? The current reliance on reverts makes the implicit assumption that RPC calls will not fail for other reasons other than revert, an assumption that hasn't held true in the past

PR #347, opened 2021-08-05
Add adapter for Balanced.

bwhli:

Balanced is a decentralized finance platform designed for ease of use. Swap assets on the ICON blockchain, and get paid Balance Tokens (BALN) to borrow assets or supply liquidity. BALN holders have the power to vote on platform changes and earn network fees.

### DIFF
--- a/projects/balanced/index.js
+++ b/projects/balanced/index.js
@@ -2,12 +2,99 @@ const retry = require("async-retry");
 const axios = require("axios");
 const BigNumber = require("bignumber.js");
 
+const icxApiEndpoint = 'https://ctz.solidwallet.io/api/v3'
+const balancedDexContract = 'cxa0af3165c08318e988cb30993b3048335b94af6c'
+const balancedLoansContract = 'cx66d4d90f5f113eba575bf793570135f9b10cece1'
+const bandOracleContract = 'cx087b4164a87fdfb7b714f3bafe9dfb050fd6b132'
+const bnusdContract = 'cx88fd7df7ddff82f7cc735c871dc519838cb235bb'
+const sicxContract = 'cx2609b924e33ef00b648a409245c7ea394c467824'
+
+async function icxCall(address, method, params) {
+    try {
+        let response = await axios.post(icxApiEndpoint, {
+            jsonrpc: '2.0',
+            method: 'icx_call',
+            id: 1234,
+            params: {
+                to: address,
+                dataType: 'call',
+                data: {
+                    method: method,
+                    params: params
+                }
+            }
+        })
+        return response.data.result
+    }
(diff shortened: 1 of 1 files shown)


bwhli:
Add adapter for Balanced.

0xngmi:
I'd prefer if you could provide a script that uses on-chain calls rather than an opaque API. On #304 you can find our reasoning for this

bwhli:
@0xngmi Thanks, we will revise this with on-chain calls. Appreciate the feedback.

0xngmi:
No problem, I'm aware that this is bothersome since the api could just work, but I really want to avoid the snowball effect mentioned on the other issue

bwhli:
@0xngmi The latest commit calls on-chain data directly and result is inline with the original stats API.

projects/balanced/index.js
```diff
@@ -0,0 +1,100 @@
...
+
+const icxApiEndpoint = 'https://ctz.solidwallet.io/api/v3'
+const balancedDexContract = 'cxa0af3165c08318e988cb30993b3048335b94af6c'
+const balancedLoansContract = 'cx66d4d90f5f113eba575bf793570135f9b10cece1'
+const bandOracleContract = 'cx087b4164a87fdfb7b714f3bafe9dfb050fd6b132'
+const bnusdContract = 'cx88fd7df7ddff82f7cc735c871dc519838cb235bb'
+const sicxContract = 'cx2609b924e33ef00b648a409245c7ea394c467824'
+
+async function icxCall(address, method, params) {
+    try {
```
0xngmi, line 13: Is it possible to remove the try/catch? I'd rather have it fail hard and be alerted of it rather than have it fail partially and get incorrect data?

0xngmi:
Is there some way we could also run this script on previous block heights to get historical data? If not it's fine but would be pretty nice

projects/balanced/index.js
bwhli: I made a new commit and removed the try/catch block. I tested it with an incorrect endpoint and it hard failed. Let me know if it looks okay.

0xngmi: Thanks, it was just a very minor thing

bwhli:
Unfortunately I'm not aware of any way to run this script against previous blocks. There are a variety of off-chain endpoints here that give all sorts of data, but I don't see anything that would easily allow us to pull historical TVL data.

https://balanced.geometry.io/api/v1/docs

0xngmi:
Understood, thanks anyway!

Kuko0419:
hey....what means all this ,,,sorry guys is very hard for me ,i use traslate at the same time ...should i do something now ?.gracias

PR #364, opened 2021-08-13
Use locked method for StakeDAO veCRV

Reinis-FRP:
Signed-off-by: Reinis Martinsons <reinis@umaproject.org>

This PR fixes the reported TVL for StakeDAO. It was incorrectly using the `balanceOf` method on Curve voting escrow contract that represents the voting weight. Instead the `locked` method should be used that represents actual CRV locked in the voting escrow contract.


0xngmi:
Good catch!

PR #365, opened 2021-08-13
update dHEDGE data to add Polygon TVL

hedgedon:


0xngmi:
Would be nice if at some point we can get an on-chain-based script that reports token amounts

hedgedon:
> Would be nice if at some point we can get an on-chain-based script that reports token amounts

Yes i'll open another PR for this in the near future

PR #371, opened 2021-08-17
[update] flabble - Update groprotocol to use PWRD and Vault tokens directly

flabble-gro:
I'm part of the gro protcol team and have updated the adpter as follows:

Update groprotocol to use PWRD and Vault tokens to calc TVL, using DAI as a representation of 1USD (until coins on coingecko). This method works back to protocol start date which is included.

##### Twitter Link:
https://twitter.com/groprotocol

https://docs.gro.xyz/gro-docs/how-it-works/security-and-audits

##### Website Link:
https://www.gro.xyz

android-chrome-512x512

##### Current TVL:
Current TVL as of "Tue 17 Aug 20:08:24 UTC 2021" which matches our other internal counts:
DAI                       6.98 M
Total: 6.98 M

##### Chain:
Ethereum

Not yet

Not yet

Gro protocol is a stablecoin yield aggregator that tranches risk and yield. The first two products built on it are the PWRD stablecoin with deposit protection and yield, and Vault with leveraged stablecoin yields.

Not yet

Yield

### DIFF
--- a/projects/groprotocol/abi.json
+++ b/projects/groprotocol/abi.json
@@ -1,86 +1,15 @@
 {
-    "deadCoin": {
-        "inputs": [],
-        "name": "deadCoin",
-        "outputs": [
-            {
-                "internalType": "uint256",
-                "name": "",
-                "type": "uint256"
-            }
-        ],
-        "stateMutability": "view",
-        "type": "function"
-    },
-    "underlyingVaults": {
-        "inputs": [
-            {
-                "internalType": "uint256",
-                "name": "",
-                "type": "uint256"
-            }
-        ],
-        "name": "underlyingVaults",
-        "outputs": [
-            {
-                "internalType": "address",
-                "name": "",
-                "type": "address"
-            }
-        ],
-        "stateMutability": "view",
-        "type": "function"
-    },
-    "curveVault": {
-        "inputs": [],
-        "name": "curveVault",
-        "outputs": [
-            {
-                "internalType": "address",
-                "name": "",
-                "type": "address"
-            }
-        ],
-        "stateMutability": "view",
-        "type": "function"
-    },
-    "token": {
-        "inputs": [],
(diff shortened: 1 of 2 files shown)


flabble-gro:
> I'm viewing the way that you meant, but I don't know why did you choose the stablecoin DAI and not the USDC one instead for example? or it doesn't matter?

At a high level it doesn't matter - you can use any stablecoin which is pegged to the USD. We used DAI rather than USDC as it makes the code a little simpler - it happens that our Vault and PWRD tokens and DAI all use 18 decimals so you don't have to do any decimal conversions when using DAI. If we'd used USDC for example we would have had to convert the USD values from 18 to 6 decimals. But that isn't that hard, if there is a reason outputting USDC is preferred we could do that. (And this bounce through another coin will hopefully go away relatively soon when our coins are listed on coingecko, after that we can just output PWRD and Vault balances)

0xngmi:
Thanks!

Mohammed18062020:
> > I'm viewing the way that you meant, but I don't know why did you choose the stablecoin DAI and not the USDC one instead for example? or it doesn't matter?
> 
> At a high level it doesn't matter - you can use any stablecoin which is pegged to the USD. We used DAI rather than USDC as it makes the code a little simpler - it happens that our Vault and PWRD tokens and DAI all use 18 decimals so you don't have to do any decimal conversions when using DAI. If we'd used USDC for example we would have had to convert the USD values from 18 to 6 decimals. But that isn't that hard, if there is a reason outputting USDC is preferred we could do that. (And this bounce through another coin will hopefully go away relatively soon when our coins are listed on coingecko, after that we can just output PWRD and Vault balances)

Ahá, I understand. However, when we do TVLs in DefiLlama, we used to do breackdown for all tokens with their balances, we don't domain it in a unic stablecoin! I don't know if this is a particular case.

PR #380, opened 2021-08-19
Added Pe,P pool2 and some missing contracts

waynebruce0x:
pool2 currently contains 0 sushi but I believe there may be plans for pool2 to earn sushi rewards


0xngmi:
thanks

PR #381, opened 2021-08-20
Added port finance.

leonardgee:
##### Twitter Link:
https://twitter.com/port_finance

Auditted by Bramah System

##### Website Link:
https://port.finance/

https://raw.githubusercontent.com/solana-labs/token-list/main/assets/mainnet/PoRTjZMPXb9T7dyU7tpLEZRQj7e6ssfAE62j2oQuc6y/PORT.png

##### Current TVL:
10M

##### Chain:
Solana

https://www.coingecko.com/en/coins/port-finance

https://coinmarketcap.com/currencies/port-finance/

Port Finance is a decentralized Money Market protocol on Solana, aiming to provide a full suite of lending products including but not limited to: variable rate lending, fixed rate lending and interest rate swaps.

https://explorer.solana.com/address/PoRTjZMPXb9T7dyU7tpLEZRQj7e6ssfAE62j2oQuc6y

Lending

### DIFF
--- a/projects/portfinance/index.js
+++ b/projects/portfinance/index.js
@@ -2,13 +2,13 @@ const {getTokenBalance} = require('../helper/solana')
 
 async function tvl() {
     const [usdcAmount, usdtAmount, solAmount, paiAmount, merAmount, btcAmount, srmAmount] = await Promise.all([
-        getTokenBalance("EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v", "2xPnqU4bWhUSjZ74CibY63NrtkHHw5eKntsxf8dzwiid"),
-        getTokenBalance("Es9vMFrzaCERmJfrF4H2FYD4KCoNkY11McCe8BenwNYB", "QyvfrbqH7Mo8W5tHN31nzbfNiwFwqPqahjm9fnzo5EJ"),
-        getTokenBalance("So11111111111111111111111111111111111111112", "BLAFX12cDmsumyB6k3L6whJZqNqySaWeCmS5rVuzy3SS"),
-        getTokenBalance("Ea5SjE2Y6yvCeW5dYTn7PYMuW5ikXkvbGdcmSnXeaLjS", "42kNZrAuwZHLtuc7jvVX7zMfkfgwbPynqzFB3zdkAEGM"),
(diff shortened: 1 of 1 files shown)


0xngmi:
When running your code with `node test.js projects/portfinance/index.js` I get the following result:
```
--- tvl ---
usd-coin                  0
tether                    0
solana                    0
usdp                      0
mercurial                 0
bitcoin                   0
serum                     0
Total: 0
```

which doesn't seem to be correct

leonardgee:
I was mistaken. The fixed by a new commit:
```
--- tvl ---
solana                    6.60 M
mercurial                 1.06 M
usd-coin                  318.72 k
serum                     17.80 k
tether                    5.40832300125
usdp                      2.2088864627800002
bitcoin                   0.655368
Total: 8.00 M
```

0xngmi:
It seems that after the last commit was applied this PR doesn't make any changes to the file

leonardgee:
That is correct. However, it seems like Port is still now showing in Defi Llama, is there any action that we need to take?

leonardgee:
Closed due to no effect.

leonardgee:
Additionally, the current TVL calculation is just approximation, what will be a good way for us to provide an accurate calculation?

0xngmi:
>  However, it seems like Port is still now showing in Defi Llama, is there any action that we need to take?

We implemented it some time ago but back then we were getting different values when making the same query consecutively, we asked on the solana discord about it and were told that it was likely a bad node and to just wait a bit. After that I just forgot, but earlier today I came back to it and fixed a bug to get it listed.

> Additionally, the current TVL calculation is just approximation, what will be a good way for us to provide an accurate calculation?

Why is the current TVL calculation an approximation? Shouldn't it be exact since it's directly querying token balances?

0xngmi:
Btw, the protocol was listed earlier today https://defillama.com/protocol/port-finance

leonardgee:
> Btw, the protocol was listed earlier today https://defillama.com/protocol/port-finance

Thank you!

> Why is the current TVL calculation an approximation? Shouldn't it be exact since it's directly querying token balances?

The current query is only for the token that hasn't been lent out. However, I think it makes more sense if it also calculated the part that **has** been lent out. Otherwise, when there is more lending happens, the TVL decreases.

0xngmi:
> The current query is only for the token that hasn't been lent out. However, I think it makes more sense if it also calculated the part that has been lent out. Otherwise, when there is more lending happens, the TVL decreases.

The main issue with this is that people will recursively borrow from your protocol and inflate the tvl if we count borrowed coins as well. That's the reason why we don't count these for other lending markets, and, to keep comparisons straight, we shouldn't do it here either

PR #386, opened 2021-08-23
Adds ILV and ILV/ETH core pools to tvl

pedrobergamini:
This PR adds Illuvium core pools and new XYZ flash pool to the TVL.

### DIFF
--- a/projects/illuvium/index.js
+++ b/projects/illuvium/index.js
@@ -8,6 +8,8 @@ const snxstk = "0x9898d72c2901D09E72A426d1c24b6ab90eB100e7"
 const snx = "0xc011a73ee8576fb46f5e1c5751ca3b9fe0af2a6f"
 const axsstk = "0x099A3B242dceC87e729cEfc6157632d7D5F1c4ef"
 const axs = "0xbb0e17ef65f82ab018d8edd776e8dd940327b28b"
+const xyz = "0x618679df9efcd19694bb1daa8d00718eacfa2883"
+const xyzstk = "0x4C6997D462b0146fA54b747065411C1Ba0248595"
 
 async function getTvl(tokens, block){
   const balances = {}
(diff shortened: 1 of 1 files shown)


0xngmi:
Thanks for the PR.

I have no problem merging the new flashpools, however we are already counting the ILV and ILV/ETH pools under staking and pool2 respectively, so I don't think these should be included under the core tvl

pedrobergamini:
We were thinking on changing the staking and pool2 labels in order to count our core pools (ILV and ILV/ETH SLP) in the TVL that is displayed in other platforms such as Coingecko, which currently only counts TVL in the flash pools. What are your thoughts on that? Maybe we remove from staking and pool2 and just add to core TVL?

Thanks for your time!

0xngmi:
The reason why we separate staking and pool2 tvl apart from normal tvl is to keep comparisons fair between protocols. For sushi or curve we don't count xSUSHI or veCRV into their TVL (same applies with a lot of other protocols with staking), so it would be unfair to count ILV staked for illuvium but not for others (curve and sushi are not the best examples for this since their staking is different, but we also have plenty of other projects with similar staking for which we separate staking).

Another solution would be to count staking for all of them, however there are 2 main issues with this:
- Most users I've spoken with don't agree with counting this as TVL, since they use TVL as a proxy for adoption and trust and staking of the protocol's own token doesn't reflect that (except in some cases where the protocol is staking, such as synthetix). However, we recognize that some users also want to count this and the protocols themselves tend to want this as well, so our solution here is to separate these different tvl sources and let the user choose how they want to define tvl by themselves (they can pick what to include into tvl on our webpage)
- Historically we haven't counted staking for all the major protocols, so to keep our methodology backward compatible it's best to keep only counting the same stuff we were counting before and keep new tvl sources separate

0xngmi:
Regarding your comment on coingecko, the reason we've built the api this way is to keep it backwards compatible. We only started counting staking and pool2 recently, so earlier when coding adapters we just didn't include these at all. This means that now we have some adapters that include staking but not for others.

If we reported tvl to coingecko as the total sum of all sources of tvl, we'd be drawing unfair comparisons between protocols, since the new additions would have extra sources of tvl that the old protocols don't have (we are slowly fixing the old ones but it takes time). Furthermore, there's the issue of what to consider tvl, as there's a wide range of opinions on what should be counted and what shouldn't.

Not just that, but if we suddenly added this to the api, we'd be changing our methodology. So to keep our api backwards compatible we've maintained the tvl number under the old methodology the same and then added new additions that can be summed together by the parties that integrate us.

So, wrapping it up, our solution to this is to export the multiple tvl sources separately and let parties that integrate us such as coingecko choose how they want to define tvl. We've also communicated this to coingecko

0xngmi:
Closing because of inactivity

0xngmi:
Moved this into a discussion: https://github.com/DefiLlama/DefiLlama-Adapters/discussions/438

PR #388, opened 2021-08-23
Update TVL request for Sifchain to match new pools response format.

rolandnsharp:
Apologies for this. We updated some of our formatting for our API and our TVL calculation broke.


0xngmi:
thanks

PR #410, opened 2021-08-30
Adds Celo Reserve's Mento AMM for locked Celo to DefiLlama

ericnakagawa:

Mento is an automated market maker (AMM) type of decentralized exchange that’s native to the Celo platform. It’s responsible for helping the Celo Dollar (cUSD) maintain its peg to the USD. When demand for the stablecoin increases the Celo protocol mints cUSD and ‘sells’ them for CELO (Celo’s native asset) via Mento. The CELO assets are used as cUSD collateral and are stored in the Celo reserve. This all happens on-chain. If demand for cUSD decreases then the reverse process happens. Mento arbitrage opportunities arise in cases where there’s cUSD/USD depeg. This incentive for traders should help restore the cUSD/USD peg.

More info here: https://medium.com/celoorg/zooming-in-on-the-celo-expansion-contraction-mechanism-446ca7abe4f

### DIFF
--- a/projects/mento/index.js
+++ b/projects/mento/index.js
@@ -1,0 +1,29 @@
+const retry = require('async-retry')
+const axios = require("axios");
+const BigNumber = require("bignumber.js");
+
+async function fetch() {
+
+    /*
+        Mento is an automated market maker (AMM) type of decentralized exchange that’s native to the Celo platform. It’s responsible 
+        for helping the Celo Dollar (cUSD) maintain its peg to the USD. When demand for the stablecoin increases the Celo protocol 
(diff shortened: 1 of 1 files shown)


0xngmi:
I've written a different version of this script on https://github.com/DefiLlama/DefiLlama-Adapters/commit/a6d0cafaf3adc3f4f15c41855d01681e6e6610c8. The main difference is that we query at a specific block height, so once we have an archive node for celo we'll be able to get historical data as well

0xngmi:
Closing since this has already been listed

PR #418, opened 2021-08-31
Fix LP Token statistics

0xdefi2021:
Forgot to count the number of token1 when splitting LP Token.


0xngmi:
Yeah, I was looking through the code and thought that was weird

PR #425, opened 2021-09-01
Updated Siren Markets tvl to include polygon and mainnet

MysticDakra:
Update the tvl so we are calculating the correct value using mainnet and polygon

##### Twitter Link: 
- https://twitter.com/sirenprotocol?s=20

##### Website Link:
- https://sirenmarkets.com/
- https://app.sirenmarkets.com/trade

white_siren_circle

##### Current TVL: 
- Need to include this

##### Coingecko ID (so your TVL can appear on Coingecko):  
- https://www.coingecko.com/en/coins/siren

- Already listed 

- Decentralized option markets on Ethereum.

- Ticker: SI
- Address: 0xD23Ac27148aF6A2f339BD82D0e3CFF380b5093de

- Options

##### [Optional] ETH addresss (to receive a LlamaPunk):
- My address: 0x04BD82Bf46de00a048d66E9Eb00C5Bc28A07Cfb5

### DIFF
--- a/projects/siren/index.js
+++ b/projects/siren/index.js
@@ -3,61 +3,116 @@
   ==================================================*/
   const { request, gql } = require("graphql-request");
   const BigNumber = require('bignumber.js');
-  const sdk = require("@defillama/sdk")
+  const sdk = require("@defillama/sdk");
 
-  const GRAPH_URL = 'https://api.thegraph.com/subgraphs/name/sirenmarkets/protocol';
+  const MAINNET_GRAPH_URL = 'https://api.thegraph.com/subgraphs/name/sirenmarkets/protocol';
+
+  const POLYGON_GRAPH_URL = 'https://api.thegraph.com/subgraphs/name/sirenmarkets/protocol-v2-matic'
+
+  const POLYGON_SERIES_ADDRESS = '0x716c543b39a85aac0240ba7ed07e79f06e1fed48'
 
 /*==================================================
   TVL
   ==================================================*/
+async function calculateMainnetTVL(timestamp, block, chainBlocks) { 
 
-  async function tvl(timestamp, block) {
-    const {
-      amms,
-    } = await request(GRAPH_URL, GET_POOLS, {
-      block,
-    });
+  const ethereumBlock = chainBlocks['ethereum'];
 
-    const result = {
-      '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48': '0'
-    };
+  const mainnet_amms = await request(MAINNET_GRAPH_URL, GET_POOLS_MAINNET, {
+    ethereumBlock,
+  });
 
-    for (let i = 0; i < amms.length; i++) {
-      const amm = amms[i];
+  const mainnet_result = {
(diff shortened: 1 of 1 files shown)


projects/siren/index.js
```diff
@@ -3,61 +3,115 @@
...
   const BigNumber = require('bignumber.js');
-  const sdk = require("@defillama/sdk")
+  const sdk = require("@defillama/sdk");
 
-  const GRAPH_URL = 'https://api.thegraph.com/subgraphs/name/sirenmarkets/protocol';
+  const MAINNET_GRAPH_URL = 'https://api.thegraph.com/subgraphs/name/sirenmarkets/protocol';
+
+  const POLYGON_GRAPH_URL = 'https://api.thegraph.com/subgraphs/name/sirenmarkets/protocol-v2-matic'
+
+  const POLYGON_SERIES_ADDRESS = '0x716c543b39a85aac0240ba7ed07e79f06e1fed48'
```
zareth-san, line 12: First, see my comment down below about using SeriesVault, rather than SeriesController

You should fetch this from the POLYGON_GRAPH_URL, that way we can de-couple this adapter from our protocol. So, if we ever change the address used for the SeriesVault, we won't need to remember to update this address.

```
query {
  seriesVaults {
    id
  }

}

```

projects/siren/index.js
```diff
@@ -3,61 +3,115 @@
...
+
+    if (!polygon_result['polygon:'+collateralToken]) {
+      polygon_result['polygon:'+collateralToken] = '0';
+    }
+    polygon_result['polygon:'+collateralToken] = BigNumber(polygon_result['polygon:'+collateralToken]).plus(response.output).toFixed();
+
+    // Get collateral in Series
+    for (let im = 0; im < amm.series.length; im++) {
+
+      const response = await sdk.api.erc20.balanceOf({
```
zareth-san, line 98: this is going to get the collateral token balance on the SeriesController, but recall we don't store any tokens on the SeriesController. I think what you want here is the SeriesVault, which is what stores the collateral locked in the protocol used to cover the options.

projects/siren/index.js
```diff
@@ -69,6 +123,25 @@
...
   `
 
+const GET_POOLS_POLYGON = gql`
+query Pools($polygonBlock: Int) {
+  amms(block: { number: $polygonBlock }) {
+    id
+    collateralToken {
+      id
+    }
+    series{
```
zareth-san, line 133: See my comment above, we don't need the series anymore, just the series vault

zareth-san (changes requested):

projects/siren/index.js
MysticDakra: Right right that makes sense ty for review ill make these edits

0xngmi:
Saw that this PR just got reviewed, please ping me when you want me to review and merge

MysticDakra:
Updated this I removed the loop because we shouldnt have more than one series vault at a time right?  
If we expect to have more than one vault I can make this to loop over what is returned from the series vault

PR #450, opened 2021-09-04
Include TVL from Polygon for Honeyswap

ruvaag:
* Enhances Honeyswap's TVL computation to include TVL from their pools on the Polygon chain.
* Streamlines TVL computation using their updated subgraphs as listed in the [xDai](https://wiki.1hive.org/projects/honeyswap/honeyswap-on-xdai) and [Polygon](https://wiki.1hive.org/projects/honeyswap/honeyswap-on-polygon) docs respectively.


0xngmi:
While this calculation will retrieve the current TVL it won't be possible to retrieve historical tvls

PR #451, opened 2021-09-04
Fix shibance

lljxx1:


env.sample
```diff
@@ -0,0 +1,17 @@
+ETHEREUM_RPC=https://mainnet.infura.io/v3/9aa3d95b3bc440fa88ea12eaa4456161
```
0xngmi, line 1: You might want to revoke this api key

lljxx1: That was taken from MetaMask

lljxx1: > You might want to revoke this api key
That was taken from MetaMask

lljxx1: That was taken from MetaMask

lljxx1:
fix uniswapv3

0xngmi:
I'd prefer to avoid the use of hardcoded pool lists since these will go stale really quickly and maintaining them is a pain

0xngmi:
Also, not sure if for now we should stick to the old version of the adapter since the one on this PR only has bsc

lljxx1:
> I'd prefer to avoid the use of hardcoded pool lists since these will go stale really quickly and maintaining them is a pain

Yes, maintain is the big problem of this kind of product！

PR #460, opened 2021-09-06
Add parrot SOL pools

michaelhly:


0xngmi:
could you revert the formatting? It's easier to modify when it's compact

michaelhly:
@0xngmi

PR #461, opened 2021-09-06
Change to Sollet Markets endpoint

jhlx:
**Serum DEX TVL Calculation**

Fetch the market list from Sollet's endpoint rather than Bonfida's endpoint, as the former contains a more comprehensive list of Serum DEX markets.


0xngmi:
Thanks, btw do you think we should switch the serum code to instead use coingecko to price tokens? Been thinking that this simplify code and make manipulated priced much less likely

PR #462, opened 2021-09-07
add new parrot pools to sunny

sunnymicrosystems:
Up to date with all pools. Probably no updates for the next week. In the future, we may host this list on our own github somehow so a PR wouldn't be necessary.

### DIFF
--- a/projects/sunny.js
+++ b/projects/sunny.js
@@ -258,6 +258,32 @@ async function tvl() {
             decimals: 9,
             tvlReader: sunnySaberPoolReader
         },
+        {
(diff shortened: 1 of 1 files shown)


0xngmi:
One pool was merged independently on dd67974dda23ff5e98c706534b14db91024c6706 by @realdealshaman. Will merge the other pool manually since fixing the merge conflicts is a pain

0xngmi:
Merged in 34f316d. As you mentioned, if a lot of PRs need to be created you could just create a file on your github and we'll pull it into out adapter

PR #464, opened 2021-09-07
Added statistics type for TVL

0xdefi2021:
##### Please help me deal with it. Thank you.

> see: https://www.coinwind.pro/bonus


0xngmi:
- Are these DAO treasury tokens? If so I'd like to separate this TVL
- I'd rather avoid any logic based try/catch blocks since the RPC calls can fail for other reasons apart from the call failing

PR #467, opened 2021-09-08
[WIP] Count All of Kolibri's TVL

keefertaylor:
Defi Llama counts locked XTZ for the Kolibri Protocol's CDPs, but not other value locked in the protocol such as the liquidity pool and protocol farms.

I've adjusted our APIs to serve these stats, and fixed the adapter to pull them.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,8127 @@
 {
   "name": "defillama-adapters",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "defillama-adapters",
+      "version": "1.0.0",
+      "license": "ISC",
+      "dependencies": {
+        "@acala-network/api": "^2.0.1",
+        "@defillama/sdk": "^2.2.10",
+        "@makerdao/multicall": "^0.12.0",
+        "@polkadot/api": "^5.4.1",
+        "@project-serum/serum": "^0.13.38",
+        "@solana/web3.js": "^1.17.0",
+        "async-retry": "^1.3.1",
+        "atob": "^2.1.2",
+        "axios": "^0.21.1",
+        "bignumber.js": "^9.0.1",
+        "dotenv": "^8.6.0",
+        "graph-results-pager": "^1.0.3",
+        "graphql": "^15.5.0",
+        "graphql-request": "^3.4.0",
+        "js-yaml": "^4.1.0",
+        "mysql": "^2.18.1",
+        "node-html-parser": "^3.3.5",
+        "node-schedule": "^1.3.3",
+        "papaparse": "^5.3.1",
+        "tronweb": "^3.2.7",
+        "web3": "^1.3.6"
+      },
+      "devDependencies": {}
+    },
+    "node_modules/@acala-network/api": {
+      "version": "2.0.1",
+      "resolved": "https://registry.npmjs.org/@acala-network/api/-/api-2.0.1.tgz",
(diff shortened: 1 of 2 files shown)


0xngmi:
In our adapters we separate the different types of TVL. In your case it seems like value locked in the liquidity pool should be separated into pool2 tvl.

Regarding the tvl on farms, could you explain what tokens go in there and what's their purpose? We want to avoid double counting

keefertaylor:
I didn't realize you separated TVL types. Let me help clarify.

> In your case it seems like value locked in the liquidity pool should be separated into pool2 tvl.

The liquidity pool is a backstop for liquidating ovens (the CDPs). Users deposit kUSD in the pool, and anyone can use the pool to liquidate an underwater oven. Users who deposit in pools receive QLkUSD, an LP token. 

Importantly, this isn't a liquidity pool for token swaps and so it isn't a pool2 tvl (IIUC).

> Regarding the tvl on farms, could you explain what tokens go in there and what's their purpose?

Farms emit kDAO, the governance token of the Kolibri protocol. There are 3 farms:
- Deposit kUSD, receive kDAO: kUSD is the stable coin minted by the Kolibri protocol.
- Deposit QLkUSD, receive kDAO: QLkUSD is the LP token for the liquidity pool. We calculate the value of 1 QLkUSD by `liquidity pool TVL / QLkUSD supply`
- Deposit Quipuswap kUSD/XTZ LP Tokens, receive kDAO: [Quipuswap](https://quipuswap.com) is the most popular dex on tezos. Users who provide liquidity to the kUSD/XTZ pair receive LP tokens, which they can deposit in the farm. We calculate the value of one LP token by `quipuswap kUSD/XTZ pool TVL / LP token supply`

0xngmi:
I agree with you about including the liquidity pool, whoever I wouldn't count the kUSD and QLkUSD farms (we'd be double counting) and I'd count Quipuswap kUSD/XTZ separately

keefertaylor:
That sounds reasonable to me, and you all have a much better sense of how to count these across projects. 

The `totals.json` file we publish actually has this broken down by individual metric, so we just need to grab the right numbers. 

IIUC: 
- I'll change this PR to count the total XTZ locked in ovens (`data.totalBalance * xtzPrice` [1])  and the total kUSD locked in the liquidity pool (`data. liquidityPoolBalance`). 
- How do you propose to count the Quipuswap kUSD/XTZ separately? If you have a link to another adapter that does this I can use it as an example do the legwork for you

LMK if that sounds reasonable, and if so I'll make it so!

---
[1] I can probably just push a change to our stats watcher that tells you the USD balance of the XTZ in ovens, rather than making you all calculate this. I'll work on this async

keefertaylor:
Actually, on second thought, I don't understand why the kUSD farm would be double counting. 

We count the XTZ in locked in ovens as TVL, but we don't count the kUSD minted. By that logic, we count *some* kUSD when it ends up in the liquidity pool, and liquidity pool LP tokens that end up in a farm are double counted. 

The kUSD farm isn't double counted anywhere though.

0xngmi:
I think QLkUSD farm is double counting because if someone deposits 1$ in the protocol we'd be counting
- 1$ XTZ locked in oven
- 0.5$ kUSD generated and locked in liquidity pool
- 0.5$ QLkUSD on the farm

the kUSD is being counted both on the liquidity pool and on the farms, thus there's double counting taking place.

Deciding whether to include kUSD farms is more complex, I can understand choosing to count it there, however if we look at similar protocols:
- For compound-style lending markets we don't count the borrowed tokens as TVL
- For unit protocol we only count the collateral tokens

So here I think it would make sense to go with the current methodology to keep comparisons against competitors equal.

0xngmi:
> How do you propose to count the Quipuswap kUSD/XTZ separately? If you have a link to another adapter that does this I can use it as an example do the legwork for you

An export of the following type would work:
```
module.exports = {
pool2:{
fetch: pool2Fetch,
},
fetch
}
```

PR #468, opened 2021-09-08
ShadeCash Fantom

waynebruce0x:


projects/shadecash/index.js
```diff
@@ -0,0 +1,53 @@
...
+    let calls = [];
+
+    // GET FTM BALANCES
+    for (let [k, address] of Object.entries(addresses)) {
+        await sdk.api.eth.getBalance({
+            target: address, 
+            block: chainBlocks['fantom'],
+            chain: "fantom"
+        }).then(b => {
+            balances["0x21be370d5312f44cb42ce377bc9b8a0cef1a4c83"] =
```
0xngmi, line 25: could be simplified with `sdk.util.sumSingle..`

PR #478, opened 2021-09-09
Update TVL for IronFinance on Avalanche and Fantom

oramis1008:
As we are expanding to Avax and Fantom, we want to update our TVL on these chains.

Thank you!


0xngmi:
Thanks, really enjoying your PRs, they are good an can be directly merged

oramis1008:
Thank you!

We still can't see our TVL showed up for Fantom. Are we missing something in the code?

0xngmi:
Fixed, will be live in 10-30 mins (deploying)

PR #485, opened 2021-09-10
fix ramp tvl: use coingecko id instead of polygon address

michaelFranciscoOV:
Using coingecko's id instead of the token's address should fix the issue of RUSD's price not being fetched


0xngmi:
Thanks!

PR #488, opened 2021-09-10
Fix orca adapter

rawfalafel:
TVL is missing new tokens. 

I'm using the fetch adapter. Please let me know if this is okay.

Thank you!


0xngmi:
For now it's alright but if possible I'd love if you could provide token breakdowns since it allows us to provide charts like the following https://defillama.com/protocol/orca/all/Tokens(USD)

rawfalafel:
Got it. We'll get the breakdown soon. Thanks for the quick turnaround!

PR #500, opened 2021-09-13
update Arbitrum graph url for Uniswap

rajivpo:
See https://github.com/Uniswap/uniswap-interface/blob/a20b2ff7fae59ea951011bd59dacf8fb9d9edf4e/src/state/data/slice.ts#L13 for confirmation


0xngmi:
The subgraph used in https://info.uniswap.org/#/ is the dev one, do you know what is the difference between the two?

rajivpo:
Oh interesting! Looks like the analytics repo reverted from `arbitrum-one` to `arbitrum-dev` a [few days ago](https://github.com/Uniswap/uniswap-v3-info/commit/408752365b260161033f75cc3d5976ec1ca95703#diff-910aabe81ff10bb1d7d3fe8541e63df3c642d29a4935add9c6c70a221a353932) - would need to look into what the differences are but going to close this anyhow.

0xngmi:
For optimism what happened is that the `-one` subgraph got stuck and then they switched to the -dev one

PR #509, opened 2021-09-14
Moonriver: Solarbeam excl staking

waynebruce0x:
Historical pNetwork returns TVL successfully, but I'm not 100% sure about my date property on past TVL entries. Think it could be using too many unix timecode digits

### DIFF
--- a/projects/pnetwork.js
+++ b/projects/pnetwork.js
@@ -1,11 +1,20 @@
-const axios = require('axios')
+const axios = require('axios');
+const { getApiTvl } = require('./helper/historicalApi');
 
-async function fetch() {
-    let a = await axios.get('https://pnetwork.watch/api/datasources/proxy/1/query?db=pnetwork-volumes-1&q=SELECT%20%22tvl%22%20FROM%20%22tvl%22%20WHERE%20time%20%3E%3D%20now()%20-%201d&epoch=ms')
-    return a.data.results[0].series[0].values.slice(-1)[0][1]
+async function current() {
+    let response = await axios.get('https://pnetwork.watch/api/datasources/proxy/1/query?db=pnetwork-volumes-1&q=SELECT%20%22tvl%22%20FROM%20%22tvl%22%20WHERE%20time%20%3E%3D%20now()%20-%201d&epoch=ms')
+    return response.data.results[0].series[0].values.slice(-1)[0][1];
+}
+async function tvl(time) {
+    return getApiTvl(time, current, async () =>{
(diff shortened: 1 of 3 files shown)


projects/pnetwork.js
```diff
@@ -1,11 +1,20 @@
...
-    return a.data.results[0].series[0].values.slice(-1)[0][1]
+async function current() {
+    let response = await axios.get('https://pnetwork.watch/api/datasources/proxy/1/query?db=pnetwork-volumes-1&q=SELECT%20%22tvl%22%20FROM%20%22tvl%22%20WHERE%20time%20%3E%3D%20now()%20-%201d&epoch=ms')
+    return response.data.results[0].series[0].values.slice(-1)[0][1];
+}
+async function tvl(time) {
+    return getApiTvl(time, current, async () =>{
+        let response = await axios.get('https://pnetwork.watch/api/datasources/proxy/1/query?db=pnetwork-volumes-1&q=SELECT%20%22tvl%22%20FROM%20%22tvl%22%20WHERE%20time%20%3E%3D%20now()%20-%201d&epoch=ms')
+        return response.data.results[0].series[0].values.map(d=>({
+            date: d.find((e) => e[0] = Math.round(time / 1000)),
```
0xngmi, line 12: I'm not sure if I understand what you were trying to do here, d is going to be an array like `[ 1631674600000, 188946104.02 ]` and then you iterate over each element and set the `0` property of this element to a fixed number.

Given that Math.round(time / 1000) !== 0, then e[0] = Math.round(time / 1000) will always return a truthy value, so it seems like the code is equivalent to d[0], with the difference that a property is being set.

The reason why I built getApiTvl is that I kept seeing a lot of apis that returned lists of time and tvl, I kept repeating code that found the closest timestamp, so I getApiTvl() to abstract that way. The getHistorical function just needs to get the value from the api and normalize it so it can be processed properly by `getApiTvl`.

One way to do this here would be to just set date to `d[0]/1000` to convert the timestamp to seconds

projects/stormswap.js
```diff
@@ -0,0 +1,19 @@
...
+  await addFundsInMasterChef(
+      balances, STAKING_CONTRACT, chainBlocks.avax, 'avax', transformAddress);
+
+  return balances;
+};
+// node test.js projects/stormswap.js
+module.exports={
+    staking:{
+        tvl: avaxTvl
+    }
```
0xngmi, line 18: The backend code relies on there always being a `tvl` or `fetch` property on the exports, when there's no tvl I just create a function that returns an empty object

PR #533, opened 2021-09-17
Add invested sunny vaults of Parrot

ungrd:


0xngmi:
what are these vaults?

0xngmi:
so parrot uses these sunny vault tokens as collateral?

PR #540, opened 2021-09-18
Elk finance on Heco

waynebruce0x:
All Elk pools are paired against ELK token - so Heco TVL = 2 * locked ELK


0xngmi:
I'd suggest using `calculateUsdUniTvl` instead, mainly because if we use that then it'll continue to work even if more pools emerge which aren't paired against ELK

PR #486, opened 2021-09-10
Add adapter for sunrise-gaming-by-dao

sunrisedao:
##### Twitter Link:
https://twitter.com/SunriseGaming_d

##### List of audit links if any:
https://www.certik.org/projects/sunrisegaming

##### Website Link:
https://sunrisegaming-dao.com/

##### Logo (High resolution, preferably in .svg and .png, for application on both white and black backgrounds. Will be shown with rounded borders):
sunc

##### Current TVL:
$10 M

##### Chain:
ethereum

##### Coingecko ID (so your TVL can appear on Coingecko): (https://api.coingecko.com/api/v3/coins/list)
https://www.coingecko.com/en/coins/sunrise

##### Coinmarketcap ID (so your TVL can appear on Coinmarketcap): (https://api.coinmarketcap.com/data-api/v3/map/all?listing_status=active,inactive,untracked&start=1&limit=10000)
N/A

##### Short Description (to be shown on DefiLlama):
By using the blockchain technology, our objective is to create a fraud-free and fair online casino. 100% of the proceeds from the project will be returned to SUNC token holders (some will be returned to users who take part in the casino games). The SUNC token also functions as a governance token. We are aiming for a complete DAO without a centralized administrator in the future.

##### Token address and ticker if any:
https://etherscan.io/address/0x692accdd8b86692427e0aa4752ae917df01cc56f

##### Category (Yield/Dexes/Lending/Minting/Assets/Insurance/Options/Indexes/Staking) *Please choose only one:
Staking

##### Oracle used (Chainlink/Band/API3/TWAP or any other that you are using):
none

##### forkedFrom (Does your project originate from another project):
no

PR #545, opened 2021-09-19
[sunrise-gaming-by-dao] Show total tvl

sunrisedao:
Add total tvl to sunrise-gaming-by-dao

Continue to https://github.com/DefiLlama/DefiLlama-Adapters/pull/486

### DIFF
--- a/projects/sunrise-gaming-by-dao/index.js
+++ b/projects/sunrise-gaming-by-dao/index.js
@@ -3,6 +3,7 @@ const {staking} = require('../helper/staking')
 const {pool2} = require('../helper/pool2')
 
 const suncAddress = '0x692aCCdD8b86692427E0aa4752AE917Df01CC56F';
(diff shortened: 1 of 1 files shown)


0xngmi:
Hey, before I applied some changes that separate staking and pool2 from the total tvl, which we are applying to all projects, this PR is reverting those changes.

I understand that you'd like us to include all of them under core tvl but could you provide some reasoning as to why should we do so and make an exception for your case (we made one for synthethix since staking is an integral part of their core protocol)

sunrisedao:
When I check my project in https://defillama.com/protocols/staking the TVL is not shows as other projects.

Screen Shot 2021-09-21 at 09 50 55

Could you advice on how to show the tvl as well?

0xngmi:
you should include staking and pool2 through this
image

0xngmi:
Closing for now, feel free to re-open with some reasoning

PR #547, opened 2021-09-19
PaintSwap - Include Staking TVL 

paint-swap-finance:
Firstly, thanks for updating to whitelist RNDM & BRUSH!

I wanted to add our BRUSH single sided staking in our masterchef contract, as we have 1/3 of our liquidity there. Unfortunately we aren't on CoinGecko yet so it can't get the price when I tried using the `staking { }` field. We have an endpoint which returns, total tvl (incl staking), just LP and just staked tvl:

https://api.paintswap.finance/tvl
`{"updated_at":1632053590,"tvl_usd":5054841.330025606,"exchange_tvl":"3391837.895034088144048127148651982","staked_brush_usd":1663003.434991517}`

But I couldn't find a way to just add the staking part. `tvl_usd` includes everything, so seemed easier to just use that!

As an aside can we also be moved to DEX category instead of yield?

### DIFF
--- a/projects/paintswap/index.js
+++ b/projects/paintswap/index.js
@@ -1,7 +1,8 @@
-const web3 = require("../config/web3.js");
 const { calculateUsdUniTvl } = require('../helper/getUsdUniTvl.js')
 const sdk = require('@defillama/sdk');
 const BigNumber = require('bignumber.js');
+const Web3 = require('web3');
+const web3ftm = new Web3(new Web3.providers.HttpProvider(`https://rpc.ftm.tools/`));
 
 const factory = '0x733A9D1585f2d14c77b49d39BC7d7dd14CdA4aa5'
 const masterchef = '0xCb80F529724B9620145230A0C866AC2FACBE4e3D'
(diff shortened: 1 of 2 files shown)


0xngmi:
I'd separate this new portion into the staking{} section (it's what we do for all similar arrangements). If you can't use staking() to get the values because of the token not being on coingecko then I'd convert staked amount into another currency (eg: FTM) and report that

paint-swap-finance:
Thanks for the feedback @0xngmi, I'm now determining the staking balance using the contracts and converting BRUSH amount to FTM so we can use the `staking{}` property now. image

paint-swap-finance:
Also our description on the website is a little out of date, so to summarise.
1 - Change from Yield farm to DEX
2 - Change description:
> PaintSwap Finance has the premier open NFT Marketplace on Fantom. It has a unique and sustainable yield farming AMM platform where 50% of farming rewards are locked for 3 months in the Art Gallery. The first project to create a video game built on the Fantom network, as well as many other exciting features in the works.
3 - Accept this PR :D

projects/paintswap/index.js
```diff
@@ -1,14 +1,61 @@
-const {calculateUsdUniTvl} = require('../helper/getUsdUniTvl.js')
+const web3 = require("../config/web3.js");
```
0xngmi, line 1: Doesn't this use an ETH RPC endpoint?

paint-swap-finance: It uses what ever the environment variable is set to. I guess if it's always points to an Ethereum RPC server on your end, then that would be a problem

paint-swap-finance: I am using the acclaimed rpc.ftm.tools endpoint now. Was recently acquired by ANKR https://twitter.com/ankr/status/1439529700192292864

PR #551, opened 2021-09-20
fixed the minor issue in Gaj

Gajesh2007:

### DIFF
--- a/projects/GajFinance/index.js
+++ b/projects/GajFinance/index.js
@@ -11,12 +11,6 @@ const JUNGLEPOOL = '0xD45AB9b5655D1A3d58162ed1a311df178C04ddDe'
 
 const GAJDFYNVAULT = '0xbf26b582680e7525da0e27ea9527bb0bf4f22de9'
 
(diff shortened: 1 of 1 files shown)


0xngmi:
what this PR does is just move tokens from staking into core tvl, is that correct?

Gajesh2007:
correct

0xngmi:
I'd rather count staking and pool2 separately since that's the methodology we are using for all other projects

Gajesh2007:
Oh. Sure

Is it possible to separate NFT Farm and Jungle Pools and add it in Core TVL

Rewards for Jungle Pools are in DAI

Rewards for NFT farm are in Points

Gajesh2007:
Actually, just want to keep some TVL in core as Coingecko doesn't add Staking in the TVL

0xngmi:
we can include into core tvl any coins staked that are not
- GAJ
- an LP pool against GAJ

Gajesh2007:
👍🏻

0xngmi:
Closing, replaced by #564

PR #563, opened 2021-09-22
Calculate lido TVL via validators staked account balance

### LINKED ISSUE #562: Lido for Solana is using the amount at the start of the epoch, not the current amount
Hi, thanks for including Lido for Solana in Defi Llama!

I noticed that to get Lido's SOL balance, you are using the internal exchange rate:

https://github.com/DefiLlama/DefiLlama-Adapters/blob/e57e1507666c7d1c1bda267125fb9181199dabfc/projects/lido/index.js#L43

This value is updated once per epoch (an epoch is roughly 2.5 days), so depending on when you query it, it may be outdated by 2.5 days. If you want to know the current balance, you should instead take the sum of [`validator.stake_accounts_balance`](https://github.com/DefiLlama/DefiLlama-Adapters/blob/e57e1507666c7d1c1bda267125fb9181199dabfc/projects/lido/Lido.js#L159) over all validators, and add to that the balance of the reserve ([`3Kwv3pEAuoe4WevPB4rgMBTZndGDb53XT7qwQKnvHPfX`](https://docs.solana.lido.fi/deployments#mainnet-beta)). That would ensure that deposits and withdrawals are reflected immediately, without having to wait for the start of the next epoch.

For reference, [the Rust version of this computation is here](https://github.com/ChorusOne/solido/blob/a601d3a8e3aabd4056ac7b14d8534b197706712e/program/src/state.rs#L563-L574).

JEasterman:
Fixes #562

### DIFF
--- a/projects/lido/index.js
+++ b/projects/lido/index.js
@@ -1,9 +1,8 @@
 const axios = require('axios');
 const abis = require('./abis.json')
 const sdk = require('@defillama/sdk')
-const { Connection, PublicKey } = require('@solana/web3.js');
-const { deserializeUnchecked } = require('borsh');
-const {Lido, schema} = require('./Lido')
+const { Connection } = require('@solana/web3.js');
+const sol = require('./sol-helpers')
 
 const ethContract = '0xae7ab96520de3a18e5e111b5eaab095312d7fe84';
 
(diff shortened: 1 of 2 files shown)


0xngmi:
thanks!

ruuda:
Thanks, this looks good to me!

PR #576, opened 2021-09-24
Tree Defi fix

waynebruce0x:


0xngmi:
should we use the generalized solution on addFundsInMasterchef? This seems like it's likely to break in the future

PR #589, opened 2021-09-26
add wusd-4pool to Mercurial Finance TVL.

siong1987:
new pool on mercurial: https://mercurial.finance/pools/wusd-4pool


0xngmi:
LFG!

PR #592, opened 2021-09-27
add zonster adapter

monsterhalo:
##### Category (Yield/Dexes/Lending/Minting/Assets/Insurance/Options/Indexes/Staking):
Yield


0xngmi:
why close it?

PR #594, opened 2021-09-27
Bao + balancer fixes

waynebruce0x:


0xngmi:
is there any way we can get the bao list from code?

waynebruce0x:
Yeah that'd be easier - I'll look into it. I used a json since that was what was used previously.

PR #602, opened 2021-09-29
Add LatteSwap fetch TVL

cappuccino-kub:
##### Twitter Link:  
- https://twitter.com/latteswapOG

- https://app.inspex.co/library/latteswap#?scope=latteswap-farm-swap-token-and-vault
- https://github.com/latteswap-official/latteswap-contract/blob/main/audit/protocol/PeckShield-Audit-Report-LatteSwap-v1.0.pdf

##### Website Link:
- https://app.latteswap.com

- https://drive.google.com/file/d/1pggcIqcQjA5u04bBOBIKoIePMHgLV4fv/view

##### Current TVL:
- $874,245,943.14 USD

##### Chain:
- BSC

--

--

- Decentralized exchange with integrated NFT functionalities operating on the Binance Smart Chain (BSC). It is a one-stop-shop for traders, yield farmers, and NFT collectors across the Blockchain ecosystem.

- 0x8d78c2ff1fb4fba08c7691dfeac7bb425a91c81a

- Dexes

##### Oracle used (Chainlink/Band/API3/TWAP or any other that you are using):
--

- SushiSwapV2

- 'TVL accounts for the liquidity on all AMM pools and single asset staking pools, using the same source as TVL on https://app.latteswap.com/yield page.

### DIFF
--- a/projects/frostfinance/index.js
+++ b/projects/frostfinance/index.js
@@ -1,76 +1,12 @@
-const sdk = require("@defillama/sdk");
-const erc20 = require("../helper/abis/erc20.json");
-const abi = require("./abi.json");
-const { unwrapUniswapLPs } = require("../helper/unwrapLPs");
-const { transformAvaxAddress } = require("../helper/portedTokens");
+const { masterChefExports } = require('../helper/masterchef')
 
 const STAKING_CONTRACT = "0x87f1b38D0C158abe2F390E5E3482FDb97bC8D0C5";
-
-const avaxTvl = async (timestamp, ethBlock, chainBlocks) => {
-  const balances = {};
-
-  const lengthPool = (
-    await sdk.api.abi.call({
-      abi: abi.poolLength,
-      target: STAKING_CONTRACT,
-      chain: "avax",
-      block: chainBlocks["avax"],
-    })
-  ).output;
-
-  const lpPositions = [];
-
-  for (let index = 0; index < lengthPool; index++) {
-    const lpOrTokens = (
-      await sdk.api.abi.call({
-        abi: abi.poolInfo,
-        target: STAKING_CONTRACT,
-        params: index,
-        chain: "avax",
-        block: chainBlocks["avax"],
-      })
-    ).output[0];
-
-    const lpOrToken_bal = (
-      await sdk.api.abi.call({
-        abi: erc20.balanceOf,
-        target: lpOrTokens,
-        params: STAKING_CONTRACT,
-        chain: "avax",
-        block: chainBlocks["avax"],
-      })
-    ).output;
-
-    if (
-      (index >= 2 && index <= 7) ||
-      index == 9 ||
-      (index >= 13 && index <= 15) ||
-      index == 17
-    ) {
-      sdk.util.sumSingleBalance(balances, `avax:${lpOrTokens}`, lpOrToken_bal);
-    } else {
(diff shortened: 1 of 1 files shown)


waynebruce0x:
Can this be calculated using an on-chain adapter?

elgrayfrappuccino:
> Can this be calculated using an on-chain adapter?

Hi @waynebruce0x . To my understanding, you are asking if we could write a contract to calculate the TVL?. In that case, it might be possible, however I would not suggest doing so since our TVL came from 3 major sources

1. PancakeSwap LP (LATTE-BUSD) pool: This one is quite straightforward, you could write a contract to read the pool reserve and convert them to USD. 

2. LatteSwap LP pair pools: This is the same as above except the contract address must be LatteSwap's contract instead of PancakeSwap. However, since we are making more partnerships and let them listed on our site, the number of pools will keep growing so it would be hard to keep track of the newer pools using any on-chain solutions. For now, we are using The Graph to keep track of all pools. (which I believe that most of AMM are using)

3. Single asset staking pools: Its issue is the same as the number 2. Plus, we are having some pools that might need more logic to calculate the TVL (for example ibAlpaca token which needs to convert back to Alpaca before calculate the price) so it would be even harder to write an on-chain solution for that.

In conclusion, the TVL might be able to calculate using a contract/adaptor but I would not suggest to keep track of any DEXs' TVL using any on-chain solutions unless you could make sure it will be updated when the new pools are added and it can have a special calculation logic for some specific pools.

Does that answer you question?

0xngmi:
- LatteSwap LP pair pools: These are not a big issue for us, we can just get them from the factory and price them
- Single asset staking pools: Could you create a separate endpoint with the tvl of these?

mint-chocolate-frappuccino:
Can we instead give you an open-source version of our api and let you validate our TVL domain logic?

mint-chocolate-frappuccino:
Please note that we use this approach because of reusability,  we can provide a codebase for you for transparency.

0xngmi:
> Can we instead give you an open-source version of our api and let you validate our TVL domain logic?

i'd prefer other methods since then we can get historical values but this works too. Only hard requirement from our side is that you separate the tvl of each thing

mint-chocolate-frappuccino:
okay, we will open source our API for TVL calculation and share you guys in a PR description, sounds good?

cappuccino-kub:
Here is the source of our API for TVL calculation
https://github.com/latteswap-official/latteswap-open-api/blob/master/src/usecases/amm/get_total_value_locked.ts

0xngmi:
Could you make the API return the value locked on the LATTE pool separately?
image

For now it's very small, but if it grows large in the future it would represent a deviation from our standard methodology

cappuccino-kub:
We update API for fetch TVL and update code for this PR
This is source code about the API https://github.com/latteswap-official/latteswap-open-api/tree/master/src/usecases/amm/defi_llama

0xngmi:
Thanks for making the API changes, given that your PR changed 10k lines of package-locked, I opted for applying the changes directly instead of merging. I also separated LATTE staking from the other sources of tvl, in order to follow the same methodology we use for our other projects (eg: pancakeswap, aave..)

elgrayfrappuccino:
@0xngmi So the code was pushed to main directly and we could close this PR, correct?

PR #604, opened 2021-09-29
add Trava Finance

haiquan5396:


waynebruce0x:
With money markets we usually calculate TVL as total supply - total borrow. Can this be done for trava?

haiquan5396:
Thanks. I fixed it.

PR #619, opened 2021-10-01
uses abi calls to calculate tvl for paladin finance

Nemusonaneko:


projects/paladinfinance/index.js
```diff
@@ -1,40 +1,66 @@
...
-      target: target,
-      owner: owner,
-      block,
-    });
-
-    //If stkAAVE address then change target address to AAVE address
-    if (target === "0x4da27a545c0c5b758a6ba100e3a049001de870f5") {
-      target = "0x7fc66500c84a76ad7e9c93437bfc5ac33e2ddae9";
-    }
+  let calls = [];
```
0xngmi, line 17: js tip: use `calls = poolAddresses.map(address=>({target: address}))`

projects/paladinfinance/index.js
```diff
@@ -1,40 +1,66 @@
...
+  for(let i = 0; i < poolAddresses.length; i++) {
+    calls.push({target: poolAddresses[i]})
+  }
+
+  let underlyingTokens = await sdk.api.abi.multiCall({
+    calls,
+    abi: abi["underlying"],
+    block:block
```
0xngmi, line 25: js tip: you can just put `block`, if key and variable name are the same, you don't have to put them both

PR #620, opened 2021-10-02
Element + Keep3r + Barnbridge(avax) + Adamant(arbi)

waynebruce0x:
NB: Adamant Arbitrum does not work - will come back to this tomorrow. PR'd because I've already spent a while debugging and thought maybe I was going round in circles.

UnhandledPromiseRejectionWarning: TypeError: Cannot destructure property '_reserve0' of '(intermediate value).output.find(...).output' as it is null.
    at /Users/test/github/DefiLlama-Adapters/projects/helper/unwrapLPs.js:251:16
    
I can't figure out why this is happening. There was another issue woth unwrapUniswapLPs, where lpPosition.balance was NaN, throwing an error. That's why I've added the if statement.


0xngmi:
thanks, this is a life saver

projects/helper/unwrapLPs.js
```diff
@@ -249,11 +249,12 @@ async function unwrapUniswapLPs(balances, lpPositions, block, chain='ethereum',
         const token1 = (await tokens1).output.find(call=>call.input.target === lpToken).output
         const supply = (await lpSupplies).output.find(call=>call.input.target === lpToken).output
         const {_reserve0, _reserve1} = (await lpReserves).output.find(call=>call.input.target === lpToken).output
-
-        const token0Balance = BigNumber(lpPosition.balance).times(BigNumber(_reserve0)).div(BigNumber(supply))
-        sdk.util.sumSingleBalance(balances, await transformAddress(token0.toLowerCase()), token0Balance.toFixed(0))
-        const token1Balance = BigNumber(lpPosition.balance).times(BigNumber(_reserve1)).div(BigNumber(supply))
-        sdk.util.sumSingleBalance(balances, await transformAddress(token1.toLowerCase()), token1Balance.toFixed(0))
+        if (token0 && token1 && supply > 0 && supply && lpPosition.balance) {
```
0xngmi, line 252: this extra check could generate some issues, problem is that the RPC that we are using might cause some calls to fail incorrectly (RPC returns call as failed but it does't actually fail, it's just that the rpc failed for some other reason). With the current code this will lead to the adapter breaking, however, with this new addition the adapter will just ignore this LP position, so it will report incorrect numbers

PR #624, opened 2021-10-03
Update endpoint to take data from the staked API

joehawary:
Currently the endpoint used is for the value of rewards in reward programs

The API in this proposal contains the value of tokens locked/staked to those reward programs.


joehawary:
@0xngmi @realdealshaman
Hope this is helpful 🙏

PR #634, opened 2021-10-05
return zero for not indexed blocks

amirtotem:


0xngmi:
I don't like this idea, we will be reporting incorrect data for un-synced blocks

amirtotem:
Due to storage problem we are not storing whole blockchain but only blocks that includes our pool.
If you are adding up the results form this endpoint it seems rational to report 0 for the unindexed block because it is zero in terms of TOTM

0xngmi:
what if your subgraph ends up stale and unable to keep up with the blockchain tip? Then the queries will return 0 tvl erroneously

PR #655, opened 2021-10-09
add almond.so

theo-almond:


0xngmi:
How is the tvl being calculated on the API? What's included?

0xngmi:
okay, please don't include staking of your token or pool2 into this api in the future. We can include them into our calculations but would rather talk about this before (same applies to other changes in the methodology used in the api) so we can make sure that we apply the same rules we are using for other protocols

PR #656, opened 2021-10-09
add bagel

zhangyi999:

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,40 +1,40 @@
-##### Twitter Link:
+##### Twitter Link: https://twitter.com/BagelsFinance
 
 
-##### List of audit links if any:
+##### List of audit links if any: https://www.certik.org/projects/bagelsfinance
 
 
-##### Website Link:
+##### Website Link: https://app.bagels.finance/
 
 
-##### Logo (High resolution, preferably in .svg and .png, for application on both white and black backgrounds. Will be shown with rounded borders):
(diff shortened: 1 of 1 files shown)


0xngmi:
We need some code to calculate TVL

0xngmi:
To get listed on defillama we need a PR that contains some code that calculates the TVL of a project, your PR is lacking that. I've looked into writing your code myself but I've run into multiple issues:
- There's no easy way to find all the pools: 0x4889a67bd24901f1f41805bf156079a2bc4d6e66 seems to be the pool factory but contract is unverified
- There is no API we can use
- All contracts are unverified

It's not possible for us to integrate your protocol because of this, if you want us to integrate you we'd need a solution for these issues

zhangyi999:
> To get listed on defillama we need a PR that contains some code that calculates the TVL of a project, your PR is lacking that. I've looked into writing your code myself but I've run into multiple issues:
> 
> * There's no easy way to find all the pools: 0x4889a67bd24901f1f41805bf156079a2bc4d6e66 seems to be the pool factory but contract is unverified
> * There is no API we can use
> * All contracts are unverified
> 
> It's not possible for us to integrate your protocol because of this, if you want us to integrate you we'd need a solution for these issues

1. You can find all the pools here:https://github.com/bagels-dev/protocol-v1
This address if for CAKE leverage:0x4889A67bD24901f1f41805bf156079a2Bc4D6E66. It also already shows in the above link.
2.Our technical staff are working on the API and it needs more time to settle down. Is it necessary to provide API for list on Defi llama?
3.Some contracts are verified on BSCscan. Should we submit this requirement again until all the contracts verified?

PR #666, opened 2021-10-11
Multichain Miner + Kavaswap

waynebruce0x:

### DIFF
--- a/projects/kavaswap.js
+++ b/projects/kavaswap.js
@@ -2,7 +2,19 @@ const utils = require('./helper/utils');
 
 async function tvl(timestamp, ethBlock, chainBlocks) {
     let balances = {};
-    const response = await utils.fetchURL('https://api.data.kava.io/swap/pools');
+    let startBlock = 4698;
+    const dailyBlockInterval = 12495;
+    let startDate = new Date(Date.UTC(2021, 7, 31));
+    let dateNow = new Date(timestamp * 1000);
+
+    while (startDate < dateNow) {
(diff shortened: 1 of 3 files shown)


projects/multichainMiner.js
```diff
@@ -0,0 +1,80 @@
...
+    polygon: {
+        tvl: polygonTvl,
+    },
+    fantom: {
+        tvl: fantomTvl,
+    },
+    avax: {
+        tvl: avaxTvl,
+    },
+    tvl: sdk.util.sumChainTvls([bscTvl, polygonTvl, fantomTvl, avaxTvl]),
```
0xngmi, line 78: after my latest changes you don't need to include a tvl function anymore, the scripts automatically add the chain tvls into a total chain tvl

projects/kavaswap.js
```diff
@@ -2,7 +2,19 @@ const utils = require('./helper/utils');
 
 async function tvl(timestamp, ethBlock, chainBlocks) {
     let balances = {};
-    const response = await utils.fetchURL('https://api.data.kava.io/swap/pools');
+    let startBlock = 4698;
+    const dailyBlockInterval = 12495;
+    let startDate = new Date(Date.UTC(2021, 7, 31));
+    let dateNow = new Date(timestamp * 1000);
+
+    while (startDate < dateNow) {
```
0xngmi, line 10: maybe we could try to estimate this only when getting historical data, in other cases we could just get the most recent data

PR #668, opened 2021-10-11
Use serum-data endpoint for serum TVL

jhlx:
Methodology:

Serum collects data on serum TVL every 5 minutes and adds it to a database. The serum-data endpoint queries all entries in this database in the last hour, sums all of the TVLs, divides them by the number of entries that were found, and returns that as the TVL.

### DIFF
--- a/projects/serum.js
+++ b/projects/serum.js
@@ -1,60 +1,22 @@
-const axios = require("axios");
-const { Connection, PublicKey } = require('@solana/web3.js');
-const { Market } = require('@project-serum/serum');
-
-const serumProgramId = '9xQeWvG816bUx9EPjHmaT23yvVM2ZWbrrpZb9PusVFin'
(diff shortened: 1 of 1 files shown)


projects/serum.js
```diff
@@ -1,60 +1,22 @@
...
-  tokenPrices['USDC'] = 1
-
-  Object.entries(balances).forEach(([token, balance]) => {
-    const price = tokenPrices[token]
-    if (price === undefined) {
-      console.error(`There's no price for ${token}`)
-      return
+        return tvl || 0;
+    } catch (e) {
+        return 0;
```
0xngmi, line 16: imo would be better to fail hard to prevent incorrect data from getting into our database, i'll apply this change

0xngmi: applied

PR #692, opened 2021-10-18
add synthetify adapter

PawelKrawczyk96:
Synthetify is a decentralized synthetic assets protocol build on the Solana blockchain. The protocol allows for the creation and exchange of synthetic assets that closely track the price of specific assets. Synthetic tokens are based on the SPL-Token standard that gives them the ability to be easily integrated with other DeFi applications like AMM.

To obtain TVL of Synthetify we must add all colaterals and synthetics which was locked in staked/debt pool.

### DIFF
--- a/projects/synthetify/index.js
+++ b/projects/synthetify/index.js
@@ -9,7 +9,6 @@ async function fetch() {
     return response[index].collateralAll
   }
 
-  const tvl = fetch()
   module.exports = {
       fetch,
       methodology:


synthetify/index.js
```diff
@@ -0,0 +1,18 @@
+const retry = require('async-retry')
+const axios = require("axios");
+
+async function fetch() {
+    const response = (
+      await retry(async () => await axios.get("https://api.synthetify.io/stats/mainnet"))
+    ).data;
+    const index = response.length - 1
+    return response[index].collateralAll + response[index].synthetic
```
0xngmi, line 9: Isn't this adding the collateral and the debt together as TVL? This seems different that the methodology explanation, which outlines that only collateral is counted

PR #698, opened 2021-10-18
Kolibri + Jaguarswap patches

waynebruce0x:

### DIFF
--- a/projects/jaguarswap.js
+++ b/projects/jaguarswap.js
@@ -1,8 +1,8 @@
 const sdk = require("@defillama/sdk");
-const { transformArbitrumAddress } = require("../helper/portedTokens");
-const {getBlock} = require('../helper/getBlock');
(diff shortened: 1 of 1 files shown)


projects/kolibri/index.js
```diff
@@ -3,11 +3,16 @@ const BigNumber = require("bignumber.js");
...
+}
+async function pool2() {
+    return new BigNumber((await axios.get(tvlUrl)).data.quipuswapFarmBalanceUSD);
 }
-
 module.exports = {
     methodology: 'TVL counts the XTZ tokens that are deposited to mint kUSD, kUSD in the liquidity pool, and value locked in Kolibri Protocol\'s farms. Borrowed tokens are not counted.',
-    fetch
+    fetch: tvl,
+    pool2: {
```
0xngmi, line 15: imo would be better so separate by chain

PR #700, opened 2021-10-19
Update to Yieldlys TVL calcs using the official endpoint

AronTurner:

### DIFF
--- a/projects/yieldly/index.js
+++ b/projects/yieldly/index.js
@@ -1,65 +1,17 @@
-var atob = require("atob"); // needed for decrypting values from contract
-const utils = require('../helper/utils');
-const BN = require("bignumber.js");
-const web3 = require('../config/web3.js');
-var UNISWAP_POOL = require("./UNISWAP_POOL.json")
+//const utils = require("../helper/utils");
+const axios = require ("axios")
+async function tvl() {
+  let usdTVL = 0;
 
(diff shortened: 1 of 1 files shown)


0xngmi:
could you separate staking from other tvl then?

AronTurner:
Can I get an update on this please @0xngmi?

waynebruce0x:
Sorry for the delay - can native Yieldly token staking be separated into a 'staking' export, instead of being included as core TVL?

AronTurner:
From what I can understand,  I will seperate the No-loss prize game TVL from the rest of the staking TVL's. So this will have 2 TVL's one for the game (which is the amount of Algo locked up) and one for all our staking pools staked amounts summed in USD value. Much like what Pancakeswap has, correct?

waynebruce0x:
yes thats correct, one tvl export, and one staking export. Users then have the option whether or not they want to include native token staking in the tvl shown. This was a community feature request and we try to keep our method consistent across all protocols.

AronTurner:
I performed an update that split it into the NLL TVL and the staking TVL, please let me know if anything else is missing

AronTurner:
@waynebruce0x there seems to be an issue on the frontend, it does not appear to have been updated with the same schema as the newest iteration of the code we pushed as all values are now 0, can this be rectified soon please?

PR #720, opened 2021-10-21
Add Votium TVL

jo-chemla:
Votium TVL is the sum of CVX and cvxCRV locked into the Votium contract

##### Twitter Link:
https://twitter.com/VotiumProtocol

##### Website Link:
https://votium.app/

dark bg rounded: https://pbs.twimg.com/profile_images/1416420051117387787/EHLIUJT3_400x400.jpg

##### Current TVL:
13M

##### Chain:
Ethereum mainnet

Votium is an incentives platform where vlCVX holders can receive compensation from buyers interested in amassing voting power. Delegating/Voting on Snapshot incurs no risk to user funds, you always retain custody of your vlCVX.

Staking / Vote bribing

Votium TVL is the sum of CVX and cvxCRV locked into the Votium contract


benber86 (commented):
Votium does not have any CVX in its custody, they are just a delegator.
What you're looking at here is Convex's locker contract. 
The cvxCRV in it is for the rewards distributed to lockers so it does not really count as TVL.
There's no locked vlCVX in the contract, vlCVX is used for accounting purposes so it's not a real token.

If you want to see the locker contract's TVL added to Convex's TVL it might be better to add that to the Convex adaptor @ https://github.com/DefiLlama/DefiLlama-Adapters/blob/main/projects/convex/index.js
The TVL would just be the CVX balance of the contract.
You can reuse the code for the staking contract that's already there and change the address to that of the locker contract.

PR #729, opened 2021-10-22
Agave edit: use sdk.api.abi.multiCall instead of thegraph

jo-chemla:


projects/agave.js
```diff
@@ -72,31 +73,153 @@ function aaveChainTvl(chain, addressesProviderRegistry, transformAddress){
...
+  // Had to explode getV2Reserves because cannot find the correct parameter to getAddress(param) on addressProvider
+  // V2 TVLs
+  let balances = {}
+  let v2Atokens = []
+  let v2ReserveTokens = []
+  let addressSymbolMapping = {}
+
+  const chain = 'xdai'
+  const block = chainBlocks[chain]
+  validProtocolDataHelpers = '0xa874f66342a04c24b213BF0715dFf18818D24014'
```
0xngmi, line 100: minor nitpick: would be good to declare

PR #733, opened 2021-10-23
[StakeDAO] Avax network added

PumpkingWok:
This PR is for include the avax network into the StakeDAO resolver, it adds the USD strategy into the TVL.

- [X] Added the avax network TVL.
- [X] Added the `av3CRV Avax` swap contract infos into the `unwrapLPs.js` file.

To calculate the USD value of the underlying tokens (avDAI/avUSDC/avUSDT) for the 3crv pools, i had to use the mainnet address for the related DAI/USDC/USDT tokens, i think that coingecko can not return the value for these cause they are not listed.


0xngmi:
thanks!

PR #757, opened 2021-10-29
add tomomaster on tomochain

thanhnguyennguyen:
add tomomaster on tomochain
(This content is created via [Gittool](https://github.com/thanhnguyennguyen/lazy/tree/master/git)) 
 On behalf of [thanhnguyennguyen](https://github.com/thanhnguyennguyen)

### DIFF
--- a/projects/tomomaster/index.js
+++ b/projects/tomomaster/index.js
@@ -1,0 +1,14 @@
+const axios = require("axios");
+
+async function tvl() {
+    var lockedTomo = await axios.get('https://tomoscan.io/api/account/0x0000000000000000000000000000000000000088')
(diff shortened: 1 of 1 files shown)


0xngmi:
we don't track native chain staking. you will find that we don't track ETH2 staking, solana staking nor TRON staking

thanhnguyennguyen:
https://defillama.com/protocol/marinade-finance

this includes native staking SOL (Solana)

0xngmi:
yeah, we include that protocol, which is built on top of staking. however that protocol only has 1.77bn in tvl, while the total amount of solana staked is 397.2M SOL = 94.930 bn USD, so it's a tiny portion of SOL staked.

The short version is that we'll include protocols built on top of staking but not native staking from the chain

Source: https://solanabeach.io/

PR #808, opened 2021-11-03
Lootswap

Define101:
cr3gGywu_400x400


0xngmi:
could you look into pushing changes under different filenames? it would lower the amount of changes needed afterwards

PR #810, opened 2021-11-04
feat(gro-protocol): adapter updated with pools

anthanh:

### DIFF
--- a/projects/groprotocol/helpers.js
+++ b/projects/groprotocol/helpers.js
@@ -1,0 +1,236 @@
+const sdk = require("@defillama/sdk");
+const BigNumber = require("bignumber.js");
+
+const lpReservesAbi = {
+  constant: true,
+  inputs: [],
+  name: "getReserves",
+  outputs: [
+    { internalType: "uint112", name: "_reserve0", type: "uint112" },
+    { internalType: "uint112", name: "_reserve1", type: "uint112" },
+    { internalType: "uint32", name: "_blockTimestampLast", type: "uint32" },
+  ],
+  payable: false,
+  stateMutability: "view",
+  type: "function",
+};
+const lpSuppliesAbi = {
+  constant: true,
+  inputs: [],
+  name: "totalSupply",
+  outputs: [{ internalType: "uint256", name: "", type: "uint256" }],
+  payable: false,
+  stateMutability: "view",
+  type: "function",
+};
+const token0Abi = {
+  constant: true,
+  inputs: [],
+  name: "token0",
+  outputs: [{ internalType: "address", name: "", type: "address" }],
+  payable: false,
+  stateMutability: "view",
+  type: "function",
+};
+const token1Abi = {
+  constant: true,
+  inputs: [],
+  name: "token1",
+  outputs: [{ internalType: "address", name: "", type: "address" }],
+  payable: false,
+  stateMutability: "view",
+  type: "function",
+};
+
+/* lpPositions:{
+    balance,
+    token
+}[]
+*/
+async function unwrapUniswapLPs(
+  balances,
+  lpPositions,
+  block,
+  chain = "ethereum",
+  transformAddress = (addr) => addr,
+  excludeTokensRaw = [],
+  retry = false
+) {
+  const excludeTokens = excludeTokensRaw.map((addr) => addr.toLowerCase());
+  const lpTokenCalls = lpPositions.map((lpPosition) => ({
+    target: lpPosition.token,
+  }));
+  const lpReserves = sdk.api.abi.multiCall({
+    block,
+    abi: lpReservesAbi,
+    calls: lpTokenCalls,
+    chain,
+  });
+  const lpSupplies = sdk.api.abi.multiCall({
+    block,
+    abi: lpSuppliesAbi,
+    calls: lpTokenCalls,
+    chain,
+  });
+  const tokens0 = sdk.api.abi.multiCall({
+    block,
+    abi: token0Abi,
+    calls: lpTokenCalls,
+    chain,
+  });
+  const tokens1 = sdk.api.abi.multiCall({
+    block,
+    abi: token1Abi,
+    calls: lpTokenCalls,
+    chain,
+  });
+  if (retry) {
+    await Promise.all(
+      [
+        [lpReserves, lpReservesAbi],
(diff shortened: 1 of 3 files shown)


projects/groprotocol/helpers.js
```diff
@@ -0,0 +1,236 @@
...
+      .div(resolvedCrvTotalSupply);
+    sdk.util.sumSingleBalance(
+      balances,
+      transformAddress(call.input.target),
+      underlyingBalance.toFixed(0)
+    );
+  });
+}
+
+module.exports = {
```
0xngmi, line 232: wouldn't it be better to use these functions from our helper scripts? are these changed?

projects/groprotocol/index.js
```diff
@@ -1,32 +1,153 @@
...
+        abi: groTokenAbi["totalSupply"],
+        block: ethBlock,
+      });
+      sdk.util.sumSingleBalance(balances, token, current.output);
+    }
+
+    // Assets held in staking pools (not counting any PWRD or GVT assets as these are already counted)
+
+    // PO_SS_GRO
+    const p0 = await sdk.api.erc20.balanceOf({
```
0xngmi, line 68: are these all pool2? if so it would be better to separate them imo

flabble-gro: Is there a full definiton of pool2 we can luck at?  I think this pool 0 may fall under that. The others we think should all be part of base TVL

projects/groprotocol/helpers.js
flabble-gro: We had to make a number of changes. Happy to feed those back to see if they could be included in core helpers

anthanh: yep this helpers has some mayor changes from original ones so had to be inlined or ad-hoc for us.

projects/groprotocol/helpers.js
```diff
@@ -0,0 +1,236 @@
...
+  const crvTotalSupply = sdk.api.erc20.totalSupply({
+    target: crvToken,
+    block,
+    chain,
+  });
+  const underlyingSwapTokens = (
+    await sdk.api.abi.multiCall({
+      calls: underlyingTokens.map((token) => ({
+        target: token,
+        params: [crvToken],
```
anthanh, line 209: original `unwrapCrv` uses `crvPools[crvToken.toLowerCase()].swapContract` whereas we need `[crvToken]` to make it work

projects/groprotocol/helpers.js
```diff
@@ -0,0 +1,236 @@
...
+    lpPositions.map(async (lp) => {
+      try {
+        await unwrapCrv(
+          balances,
+          lp.token,
+          lp.balance,
+          block,
+          chain,
+          transformAddress,
+          excludeTokens
```
anthanh, line 174: `unwrapCrvLPs` with `excludeTokens` support

projects/groprotocol/helpers.js
anthanh: also has `excludeTokens` support

projects/groprotocol/helpers.js
```diff
@@ -0,0 +1,236 @@
...
+          const token0Balance = BigNumber(lpPosition.balance)
+            .times(BigNumber(_reserve0))
+            .div(BigNumber(supply));
+          sdk.util.sumSingleBalance(
+            balances,
+            await transformAddress(token0),
+            token0Balance.toFixed(0)
+          );
+        }
+        if (!excludeTokens.includes(token1)) {
```
anthanh, line 125: `excludeTokens` logic differs from original and not compatible between them

projects/groprotocol/helpers.js
anthanh: @0xngmi added some comments about why we need this ad-hoc helpers

projects/groprotocol/helpers.js
```diff
@@ -0,0 +1,236 @@
...
+        throw e;
+      }
+    })
+  );
+}
+
+const crvPools = {
+  // PWRD-3CRV
+  "0xbcb91e689114b9cc865ad7871845c95241df4105": {
+    swapContract: "0x001C249c09090D79Dc350A286247479F08c7aaD7",
```
anthanh, line 148: maybe we can resuse core `unwrapCrv` helper if we set `swapContract: '0xbcb91e689114b9cc865ad7871845c95241df4105'` and add exclusion logic in the core function. wdyt @0xngmi ?

projects/groprotocol/helpers.js
anthanh: actually, maybe we should check `unwrapUniswapLPs` exclusion logic, seems to be broken @0xngmi @flabble-gro

projects/groprotocol/index.js
0xngmi: pool2 are just LP tokens where one of the underlying tokens is the governance token. a lot of protocols have farming pools for them that are unrelated to the core TVL of the protocol, so we separate these and allow users to decide whether they want to include it into their tvl or not

0xngmi: for example the pool2 of UNI would be UNI-WETH

anthanh:
https://github.com/DefiLlama/DefiLlama-Adapters/pull/815

projects/groprotocol/helpers.js
0xngmi: got it!

projects/groprotocol/helpers.js
0xngmi: afaik the changes made in #815 make it so only the excluded tokens are excluded while the current code excludes the whole pair if any of it's tokens are on the excluded list. the reason why i made it that way was because i was using it to exclude pool2 pairs, so i wanted to exclude whole pairs if one of the tokens was the gov token

projects/groprotocol/helpers.js
0xngmi: could do it but if you prefer to keep it here i'm cool with that too

0xngmi:
everything looks good, only issue remaining are the pool2 bits and staking, we separate these and then allow users to choose whether to include them.

I made some changes for this on 1b00966bbdf128fe8e76560d58c1d41a9b912a7c. Please let me know if any change was incorrect or if i identified staking/pool2 incorrectly

PR #834, opened 2021-11-08
Change bagels finance

GriffinWhiting:

### DIFF
--- a/projects/bagel-finance/index.js
+++ b/projects/bagel-finance/index.js
@@ -10,7 +10,8 @@ const tokens = ["0x55d398326f99059ff775485246999027b3197955",
                 "0x2170ed0880ac9a755fd29b2688956bd959f933f8",
                 "0x9c65ab58d8d978db963e63f2bfb7121627e3a739",
                 "0x0e09fabb73bd3ade0a17ecc321fd13a19e81ce82",
-                "0x8ac76a51cc950d9822d68b83fe1ad97b32cd580d"]
+                "0x8ac76a51cc950d9822d68b83fe1ad97b32cd580d",
+				"0xbb238fce6e2ee90781cd160c9c6eaf3a4cfad801"]
 
 const lps = [{"0":"0x35e716c5D1038D41BdbA201413dBa9FDF16BBede","1":"0x0eD7e52944161450477ee417DE9Cd3a859b14fD0"},
              {"0":"0x9583a15ED9766634d5b947F9837fB9156e80Cc55","1":"0x58f876857a02d6762e0101bb5c46a8c1ed44dc16"},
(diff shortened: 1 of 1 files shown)


GriffinWhiting:
Add v1lp pool and bagel lending pool，Because our average lending rate is twice, so we need to add i6 twice

projects/bagel-finance/index.js
```diff
@@ -62,11 +74,11 @@ async function tvl(timestamp, ethBlock, chainBlocks) {
   }).map((n) => {
     const stakingTokenAddr = n.input.params[0];
     const i5 = n.output['5'];
-    //const i6 = n.output['6'];
+    const i6 = n.output['6'];  
     const i8 = n.output['8'];
     return {
       token: stakingTokenAddr,
-      balance: BigNumber(i5).plus(BigNumber(i8)).toFixed(0)
+      balance: BigNumber(i5).plus(BigNumber(i6)).plus(BigNumber(i6)).plus(BigNumber(i8)).toFixed(0)
```
0xngmi, line 81: doesn't i6 count borrowed money, which is used on the lps (which we are also counting)? therefore i'm worried that this may lead to double counting (count money borrowed from the pools once, and then again when counting the money in lps)

GriffinWhiting: Hi sir, the TVL calculated by : Leding + Borrwoing + LP.
i6 is for Borrowing, maybe it will be double counted by LP, cuz when the assets borrowed, user need to depostied the same value equivalent to borrowing into our platform. So it might be counted by LP too. But we can't change this contract , so could you pls update it in accordance with this calculation?
So that's the reason why TVL shows 15M+ in our website but shows 8M in Defi llama.

GriffinWhiting:
Hi sir, the TVL calculated by : Leding + Borrwoing + LP.
i6 is for Borrowing, maybe it will be double counted by LP, cuz when the assets borrowed, user need to depostied the same value equivalent to borrowing into our platform. So it might be counted by LP too. But we can't change this contract , so could you pls update it in accordance with this calculation?
So that's the reason why TVL shows 15M+ in our website but shows 8M in Defi llama.

projects/bagel-finance/index.js
0xngmi: If i6 is for borrowing, doesn't that money get borrowed into the LPs?
the issue that i'm worried about is that:
- I have 1 USD and deposit it into a pool
- Then someone else borrows and LPs with it
- Then I'll count the value of the LP staked (including 1$) and on top of that the value of the 1$ borrowed
- the 1$ i provided gets counted twice
is this correct?

waynebruce0x:
For money markets we don't count borrowed assets, which is probably where the 15M / 8M difference is

PR #836, opened 2021-11-08
Update DefiLlama TVL view to show Total Stake Value Locked rather than Olympus Treasury Balance

appleseed-iii:
DefiLlama TVL view for Olympus previously showed the Treasury Balance of the Olympus protocol rather than the Total Value Locked.

> The dollar amount of all OHM staked in the protocol.

This methodology now matches the methodology of other protocols.

### DIFF
--- a/projects/olympus/index.js
+++ b/projects/olympus/index.js
@@ -13,23 +13,11 @@ const OlympusStakings = [
 
 const OHM = "0x383518188c0c6d7730d91b2c03a03c837814a899";
 
-const treasuryAddresses = [
-  // V1
-  "0x886CE997aa9ee4F8c2282E182aB72A705762399D",
-  // V2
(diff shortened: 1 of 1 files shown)


0xngmi:
We purposefully separate olympus TVL into two sections:
- staking: Staked OHM
- tvl: treasury

we also do the same with all of our other projects (stkAAVE, xSUSHI)

is this PR just joining these two? if so we'd prefer to keep them separated

appleseed-iii:
> We purposefully separate olympus TVL into two sections:
> 
> * staking: Staked OHM
> * tvl: treasury
> 
> we also do the same with all of our other projects (stkAAVE, xSUSHI)
> 
> is this PR just joining these two? if so we'd prefer to keep them separated

aaah my fault. I somehow never saw that `staking` toggle before. Closing this PR; unnecessary now.
image

PR #848, opened 2021-11-09
Add MintSwap

secretemate:

### DIFF
--- a/projects/mintswap/index.js
+++ b/projects/mintswap/index.js
@@ -1,0 +1,82 @@
+const sdk = require("@defillama/sdk");
+const erc20 = require("../helper/abis/erc20.json");
+const abi = require("./abi.json");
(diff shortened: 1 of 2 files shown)


projects/mintswap/index.js
```diff
@@ -0,0 +1,82 @@
...
+        abi: erc20.balanceOf,
+        target: lpOrTokens,
+        params: FARMING_CONTRACT_ADDRESS,
+        chain: "avax",
+        block: chainBlocks["avax"],
+      })
+    ).output;
+
+    if (
+      index == 12 ||
```
0xngmi, line 45: this is bound to break as more tokens are added, what do you think about using addFundsInMasterchef?

0xngmi:
already merged on another PR, no activity here so I'll close

PR #851, opened 2021-11-10
Add more tokens to KogeFarm TVL - wMEMO balance need to be divided by 10**9

kogecoindev:
Hi! This commit is to more accurately count our TVL by including tokens that are previously valued at $0. We do so by updating token addresses not found on CoinGecko into the corresponding addresses on other chains that are found on CoinGecko before fetching the balance. 

There is one problem with the commit, which as I have discussed in Discord is that $wMEMO on Fantom is 18 decimals while the token it corresponds to 1:1, $Time on AVAX, is 9 decimal places. This causes our $wMEMO balance to be multiplied by 10**9. We would like to divide by it but are not sure how to cleanly do so, and have been asked to commit anyways.

### DIFF
--- a/projects/kogefarm/helper.js
+++ b/projects/kogefarm/helper.js
@@ -22,6 +22,22 @@ function transformAddressKF(chain = 'polygon') {
       return `ethereum:0xdac17f958d2ee523a2206206994597c13d831ec7`
     }
     if (
+      // Dai on Fantom
+      (chain === 'fantom' &&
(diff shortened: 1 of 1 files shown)


waynebruce0x (approved):

kogecoindev:
Hey as mentioned in this PR and on Discord this commit has a problem - our wMemo balance needed to be divided by 10^9. We were told to send a PR with what we have and that you guys will try to implement the required division. Just giving you guys a heads up on this because now our Fantom TVL is $65 trillion in the test suite.

PR #853, opened 2021-11-10
FODL: get eth mainnet TVL by summing all users positions suppliedAmmount

jo-chemla:
FODL: get eth mainnet TVL by summing all users positions suppliedAmmount. Use the erc721 contract and lens contract to get the user positions to query tokens for.


0xngmi:
ty

PR #890, opened 2021-11-15
TVL unirex finance

unirexfinance:
##### Twitter Link:
https://twitter.com/UnirexFinance


waynebruce0x:
The reason your TVL is exported as 'staking' is because all of the TVL we have recorded is in 'LIKE' tokens - which is your native token.

waynebruce0x (approved):

PR #893, opened 2021-11-16
add AxeDAO

axedao:


waynebruce0x:
Hey could you separate AXE_DAI_SLP tvl out into a separate pool2 export?

axedao:
Thanks for the feedback, we had updated.

waynebruce0x (approved):

PR #897, opened 2021-11-16
Newdex & DFS - single alt tokens are falsely increasing the TVL

DenisCarriere:
### ❗**ISSUE**

Single alt tokens are falsely increasing the TVL:
  - BEAN: +$74MM USD in NewDex
  - MASK +$31MM USD in DFS

### 🔨 **FIX**

- [x] remove alt tokens from being calculated in TVL
- [x] get currency balance directly from chain
- [x] remove EOS Flare API
- [x] only calculate tokens which are listed on CoinGecko

##### Current TVL

- NewDex: ~$921,979 USD
- DFS: ~$10,622,269 USD

### DIFF
--- a/projects/dfs/index.js
+++ b/projects/dfs/index.js
@@ -4,9 +4,8 @@ const EOSFLARE_ENDPOINT = "https://api.eosflare.io";
 
 async function get_account_tvl(account) {
   const response = await axios.default.post(EOSFLARE_ENDPOINT + "/v1/eosflare/get_account", {account});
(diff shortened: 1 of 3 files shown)


0xngmi:
given that we are rugging the tvl of newdex/dfs, what do you think is the best way to g about incorrect historical data? should we just remove it?

0xngmi:
I also pushed a new change (7df2cb9) that makes the adapters directly return the coingecko ids, that way we can display that on our page and provide more advanced analytics

PR #905, opened 2021-11-17
Update Allbridge contract address after migration

randyn:


randyn:
ABR's Celo TVL fell off a cliff after the 13th which is around the migration. This updated address should take care of calculating TVL again and also includes Optics USDC bridged from CELO through ABR

waynebruce0x (approved):

PR #923, opened 2021-11-19
Fodl mainnet tvl

jo-chemla:
After discussing with the FODL team, they finally accepted this revised version to account for their TVL - only counting for capital provided by user, and not that that results from leveraged positions on aave and compound using their borrow-lend loops using flashloans.

### DIFF
--- a/projects/fodl/index.js
+++ b/projects/fodl/index.js
@@ -22,27 +22,36 @@ async function tvl(timestamp, ethBlock, chainBlocks) {
     })
   ).output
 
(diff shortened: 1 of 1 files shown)


projects/fodl/index.js
```diff
@@ -90,6 +99,5 @@ module.exports = {
   ethereum: {
     tvl: tvl,
   },
-  tvl,
   pool2: ethPool2,
```
0xngmi, line 102: would be better to put pool2 inside ethereum, that way we know in which network pool2 is in

PR #958, opened 2021-11-23
Remove yarn.lock from gitignore

macalinao:
- Don't gitignore yarn.lock
- Remove yarn.lock from gitignore

This will make running the test scripts more consistent across machines.


0xngmi:
it's true that it would make things more consistent, but when we had it un-ignored people would commit >1000-line changes of yarn.lock into their PRs, and these changes were unreviewable, so I had to keep asking people to remove those changes from their PRs or merge and remove them myself. This became so common that I decided to stop tracking all lockfiles

0xngmi:
Closing because of inactivity

PR #967, opened 2021-11-24
Tangoswap

waynebruce0x:


0xngmi:
already merged I think? https://github.com/DefiLlama/DefiLlama-Adapters/blob/main/projects/tangoswap/index.js

PR #974, opened 2021-11-25
DeHive sdk adapter

NikMarin:


0xngmi:
Atm you are counting everything as tvl, will merge and split staking and pool2 into their own sections

PR #995, opened 2021-11-28
radao

radaofi:


0xngmi:
added some fixes since coin is missing from coingecko

PR #1017, opened 2021-11-30
Tonswap + Telos REX

waynebruce0x:
New chain: TON OS
https://tonswap.io/
TVL here: 
https://tonscan.io/accounts/0:0672930D0F4D76924C6C3D5BE976DBBB74685149C6717FB37C7A6202381CBD36


projects/telosRex.js
```diff
@@ -0,0 +1,16 @@
+const axios = require("axios");
+const retry = require("./helper/retry");
+
+async function tvl() {
+  const tvlTlos = (await retry(async (bail) => axios.get(
```
0xngmi, line 5: can we get historical through this as well?

0xngmi: sorry, was wrong

PR #1038, opened 2021-12-02
Fix SORA adapter

stefashkaa:


stefashkaa:
@0xngmi Hello! I've fixed SORA adapter, so, please review

0xngmi:
do you know why is the number returned by your code different from the number on https://polkaview.io/polkaswap/ ?

stefashkaa:
> do you know why is the number returned by your code different from the number on https://polkaview.io/polkaswap/ ?

Maybe we just have different approach. We use subquery to calculate all prices according to DAI price each hour.

PR #1051, opened 2021-12-03
Mixed

waynebruce0x:


0xngmi:
i checked rvirkitty yesterday and it seems like their contract deposits the pool2 tokens into hucleberry's masterchef, so counting the pool2 tokens on their masterchef would miss some tokens (would also be great to split these into a pool2 export)

waynebruce0x:
> i checked rvirkitty yesterday and it seems like their contract deposits the pool2 tokens into hucleberry's masterchef, so counting the pool2 tokens on their masterchef would miss some tokens (would also be great to split these into a pool2 export)

Added

projects/idleDao.js
```diff
@@ -0,0 +1,61 @@
+
+const sdk = require('@defillama/sdk');
```
0xngmi, line 2: this only counts the tokens on polygon, whereas most idle tvl is on ethereum. the current idle adapter has already been updated to have polygon

flabble-gro (commented):
flabble from Gro here. Many thanks for turning this round so quickly and looks good for the new Avalanche assets. One quick thing - the current PR has the previous Ethereum parts commented out, I assume this was just for testing the Avalanche pieces but please do uncomment this before merging :)

waynebruce0x:
> flabble from Gro here. Many thanks for turning this round so quickly and looks good for the new Avalanche assets. One quick thing - the current PR has the previous Ethereum parts commented out, I assume this was just for testing the Avalanche pieces but please do uncomment this before merging :)

yeah so sorry, I missed that, I've uncommented it now

flabble-gro:
Is there anything we can do to help get the gro protocol part of this PR merged - we'd love to see the Avax funds shown

PR #1068, opened 2021-12-06
Update UMA TVL

abg4:
Made the following updates to the UMA TVL calc:
- Included TVL of LSP Contracts on Ethereum/Polygon
- Included TVL of Jarvis LSP contracts on Polygon which aren't captured by the LongShortPairCreator
- Included TVL of Across contracts

### DIFF
--- a/projects/uma/index.js
+++ b/projects/uma/index.js
@@ -1,60 +1,259 @@
-const { request, gql } = require("graphql-request");
-const sdk = require('@defillama/sdk');
+const sdk = require("@defillama/sdk");
 const { sumTokens } = require("../helper/unwrapLPs");
 const { requery } = require("../helper/requery");
+const { transformPolygonAddress } = require("../helper/portedTokens");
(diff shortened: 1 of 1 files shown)


projects/uma/index.js
```diff
@@ -1,60 +1,259 @@
...
+      target: `0x${poolLog[1].slice(26)}`,
+    })),
+    block,
+    abi,
+  });
+  await requery(collaterals, "ethereum", block, abi);
+  await sumTokens(
+    balances,
+    collaterals.output
+      .filter((t) => t.output !== null)
```
0xngmi, line 54: can we remove this? in some cases our rpc will fail and then we should crash the entire script to avoid writing wrong data to storage

abg4: The commit uses the contract creator events instead of the registry so it is essentially filtering out the contracts we can't calculate the collateral for and gets us to the same number. Let me know if that works.

PR #1070, opened 2021-12-06
tvl update

ramverma-oro:


waynebruce0x:
does this just add try, catch blocks? We prefer not to use them so that we can easily flag broken adapters.

ramverma-oro:
okay let me remove that

PR #1071, opened 2021-12-06
tvl update

ramverma-oro:


ramverma-oro:
Hi @waynebruce0x,
We have multiple tokens and multiple cohorts. We are using single token multiple time in different cohorts.
like token - t1, t2, t3, t4, t5 ... etc
cohorts - c1, c2, c3, c4 ... etc
calls: [
   { 
      target: c1
      params: t1
   }
  { 
      target: c2
      params: t1
   }
   { 
      target: c3
      params: t1
   }
  { 
      target: c4
      params: t1
   }
 ... 
]

sdk will return object. There might be chances that we are getting value once in object. or may be reason we are getting very less  TVL.
actual TVl : 18M 
showing : 4-5M

Could you please help us with this?

ramverma-oro:
@themohitmadan

PR #1073, opened 2021-12-06
urgent: fix vikings finance

morazzela:
I made a mistake when adding vikings adapter, I used the staked token instead of the main token, resulting in very high staking TVL

Also, we got listed on coingecko and CMC, here are the IDs :

Coingecko : vikings-finance
CMC : 15611

Thanks !


Define101:
Working on it should be live within 1 hour 
Thanks a lot for informing us

PR #1088, opened 2021-12-08
Planet Finance GAMMA + AQUA

ash-greninja-poke:


waynebruce0x:
Can GAMMA & AQUA TVL be exported as a seperate 'staking' export?

waynebruce0x:
Since GAMMA & AQUA are your own incentive / governance tokens, can the value of these tokens be exported from the adapter seperately to the core TVL? Like:

module.exports = {
    tvl, 
    staking (<= this is where to record the value of locked GAMMA & AQUA)
}

PR #1104, opened 2021-12-11
Pickle staking pool2s

MisterMard:
- Added new chains ( Arbitrum, Harmony, OkexChain, Moonriver, Cronos, Aurora).
- Added staking and pool2s.
- Data is fetched from Pickle API.


projects/pickle/index.js
```diff
@@ -1,114 +1,92 @@
...
+  return retryModule(func, {
+    retries: 3,
+  });
+}
 
-// Contracts
-const dillAddress = "0xbBCf169eE191A1Ba7371F30A1C344bFC498b29Cf";
+const endpoints = [
+  "https://api.pickle.finance/prod/protocol/pfcore/",
+  "https://f8wgg18t1h.execute-api.us-west-1.amazonaws.com/prod/protocol/pfcore/",
```
janklimo, line 13: Nice job @MisterMard I think we can drop this fallback API endpoint to simplify things, it shouldn't really be exposed publicly. Similarly, retry logic shouldn't be necessary. DefiLlama will try again next cycle if anything fails.

MisterMard: Thanks for your comments. All resolved now.

PR #1112, opened 2021-12-13
Add borrowed and fixed eth tvl

0xodia:
There is a fix for wewETH TVL that I snuck in here as it wasn't being included. If this PR is going to take a while to merge, let me know and I'll put up a separate one for that.

### Description
Adds a borrowed method for Solend to pick up borrowed value.

TVL + borrowed will get you a view of what's deposited as requested @0xngmi .

Did this based on my understanding in tg. Feel free to takeover and fix if not fully to specs.

### Testing
Ran test script to see that it calculates correct borrowed amount

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "@project-serum/anchor": "^0.18.2",
     "@project-serum/serum": "^0.13.38",
     "@solana/web3.js": "^1.30.2",
+    "@solendprotocol/solend-sdk": "^0.1.64",
     "@vechain/connex-driver": "^2.0.7",
     "@vechain/connex-framework": "^2.0.7",
     "async-retry": "^1.3.1",
--- a/package.json
+++ b/package.json
@@ -21,7 +21,6 @@
     "@project-serum/anchor": "^0.18.2",
     "@project-serum/serum": "^0.13.38",
     "@solana/web3.js": "^1.30.2",
-    "@solendprotocol/solend-sdk": "^0.1.64",
     "@vechain/connex-driver": "^2.0.7",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -21,6 +21,7 @@
     "@project-serum/anchor": "^0.18.2",
     "@project-serum/serum": "^0.13.38",
     "@solana/web3.js": "^1.30.2",
+    "@solendprotocol/solend-sdk": "^0.1.64",
```
0xngmi, line 24: is parseReserve just a small snippet or code or is it significant? If former might be good to just port the code directly here, since then we can avoid an extra dependency

also, what is weweth?

0xodia:
parseReserve might be a bit awkward to rip out. It also might update soon
so maybe it’s a worthwhile dependency and let us read other solend accounts
better in the future?

Also weweth is wormhole eth, one of the other eth bridges on solana with
the other being sollet eth. This will likely be the higher liquidity
Eth going forward but they are both fwiw, considered Eth.

package.json
0xodia: actually ignore my other message about taking the snippet. I was able to get out the needed files out :)

0xodia: Also weweth is wormhole eth, one of the other eth bridges on solana with
the other being sollet eth. This will likely be the higher liquidity
Eth going forward but they are both fwiw, considered Eth.

waynebruce0x (approved):

waynebruce0x:
@0xngmi ready to merge this since comments seem resolved. Let me know if this should be kept open.

package.json
0xngmi: ty, all good

PR #1135, opened 2021-12-17
cohort added

ramverma-oro:


ramverma-oro:
Hi Team,
Our TVL is almost 15 million now. Previously it was showing very less. If it does not nearby 15 million. Please let  us know the possible reason cause I'm having trouble in getting value in terminal.

PR #1138, opened 2021-12-17
Add Aurora as a project under NEAR

kcole16:
##### Twitter Link:
https://twitter.com/auroraisnear

##### Website Link:
https://aurora.dev

aurora

##### Current TVL:
$209.55m

##### Chain:
NEAR

Aurora is the EVM environment on NEAR, implemented as a smart contract, bridges, and network of relayers

0xaaaaaa20d9e0e2461697782ef11675f668207961, AURORA

Includes the Aurora TVL as a project under NEAR

### DIFF
--- a/projects/aurora.js
+++ b/projects/aurora.js
@@ -1,0 +1,14 @@
+const utils = require('./helper/utils');
+
+async function tvl() {
+  const totalTvl = await utils.fetchURL('https://api.llama.fi/charts/aurora')
+
+  return {
+    near: totalTvl.data.pop().totalLiquidityUSD
+  }
+}
+
+module.exports = {
+    methodology: 'Includes the Aurora TVL as a project under NEAR',
+    tvl,
+  }


0xngmi:
I don't think we should do this, we'd be double counting aurora's tvl under 2 chains

kcole16:
Right now, NEAR's tvl is undercounted, as all Aurora protocols are running on NEAR and thus it's tvl is also NEAR's tvl.

I don't see this as double counting per se, as it should be clear to anyone viewing the page that NEAR's tvl is all applications running on NEAR (which includes Aurora) and Aurora's tvl is only including Aurora-specific protocols. 

Another way to do this may be to remove Aurora from the chain list, but still have the breakdown of protocols viewable in some way.

0xngmi:
Imo what's most important for us is to be consistent and a similar case to this one is what we have with arbitrum/optimism. I'm aware that aurora is not a rollup but from a end-users perspective these are both similar scenarios.

I created a poll to gauge general sentiment (https://twitter.com/0xngmi/status/1471974693859172356) and most people seem to think that it would be best to keep these separate. I also understand that this affects near heavily since tvl on aurora represents a big chunk of their TVL, so what I could do is just move all aurora projects to be directly under near.

In any case I don't believe that counting aurora as a dapp on near is the solution, since it goes against users' mental models (our current dapps are things like curve or aave, aurora is very different from this), but I am open to other solutions that would involve changes in our UI to represent this relationship better

kcole16:
> I also understand that this affects near heavily since tvl on aurora represents a big chunk of their TVL, so what I could do is just move all aurora projects to be directly under near.

This seems like a reasonable approach, and is probably the right one for now. 

Ideally, there would be some filter for users that wanted to easily see which projects use Aurora, but maybe that's a separate feature request.

0xngmi:
I'm okay with that but want to get the ok from aurora before proceeding with that

ilblackdragon:
Firstly, looking at your poll, it's a pretty high percentage of people who would expect Ethereum rollups as part of the Ethereum TVL: indeed the Ethereum consensus is what secures that TVL, which means that even things like understanding what is the TVL / Mkt.Cap fraction for security considerations should account for it. Although there is also a difference between captures funds in the rollup settlement contract and TVL of the rollup itself, and it might make sense to show the former.

Secondly, I would argue that Aurora as a solution is sufficiently different from the current rollup offerings, app chains or parachains. Albeit it has it's own ecosystem, it is also an "app" on NEAR - every single tx to Aurora goes via NEAR network - https://explorer.near.org/accounts/aurora, NEAR accounts or contracts can call into Aurora's contracts within the same tx and etc.

We can expect more of such applications appearing on NEAR, as it provides a rich programming environment where anyone can launch another container with sub-apps in them.

I agree with @kcole16 that it would be better to have some label on indicate to users that this is not just an app but it has it's own ecosystem and to link to it. Would that be possible to add?

alexauroradev:
Adding to what Illia said, in fact NEAR has some features of the L2s:
Due to the architecture of the Rainbow Bridge (trustless setup); a NEAR light client node is working as a smart contract on Ethereum (see [here](https://etherscan.io/address/0x3be7df8db39996a837041bb8ee0dadf60f767038)). This means that against retrospective attacks NEAR blockchain is ... as secure as Ethereum! And if the transaction execution challenges are implemented, the security assumptions of NEAR will be equal to rollups.

So, actually, with the evolution of the advanced tech; there's no fine line between L1 and L2. Even now there're a lot of confusion around it and people do not understand how to assess whether a solution is an L2 or L1 or side-chain or whatever. Mainly these labels are used for marketing.

My point of view:
* Aurora is working on top of NEAR and if NEAR is dead, Aurora is dead
* Aurora ecosystem is separated from NEAR, since Aurora users are not able to cascade transactions in NEAR through the composability feature of the EVM (syncronous execution)
* It's true that some Aurora users are also NEAR users, but not all of them

That's why the best way to depict the TVL would be:
1) Continue having the separate Aurora page to show the stats of the Aurora ecosystem
2) In the NEAR page insert generic stats of Aurora as a sub-category or a separate new block of data

Whether to calculate Aurora's TVL in the NEAR's TVL is up to you; but the connection of Aurora to NEAR should be there in the NEAR page.

0xngmi:
What do you think about this solution:
- on the chains page (https://defillama.com/chains) we list aurora as a "child" blockchain of near, and add a toggle to allow users to sum it's TVL to near (unsure if we should sum it directly, there are some issues about where to draw the line for parent-child chain relationships, for example for cosmos-terra, and in those cases summit it directly might cause problems)
- we keep https://defillama.com/chain/Near the same, highlighting the apps that are directly on top of near

I'm also open to other suggestions on how to change the UI to appease everyone

ilblackdragon:
> there are some issues about where to draw the line for parent-child chain relationships, for example for cosmos-terra,

I think the definition of the child should be around security. If the parent ensures security of another app/chain/whatever, it's fine to sum up.

Terra is not a child, just using the same Cosmos SDK as many other chains.
Similarly how ETC is not child of ETH or Kusama is not a child of Polkadot. 
But parachains for Polkadot are actually children by this definition.

> highlighting the apps that are directly on top of near

By this definition "Aurora App" (https://explorer.near.org/accounts/aurora) should be actually listed with assets deposited into the contract (not the TVL of apps on Aurora though)

alexauroradev:
The current version of the [chains page](https://defillama.com/chains) looks compelling. I believe the change recently implemented actually closes this issue.

0xngmi:
Yes, this conversation is the reason why we implemented that change

0xngmi:
If everyone is okay with this fix then we can just close this issue

ilblackdragon:
Epic change @0xngmi ! good to close.

PR #1154, opened 2021-12-20
Fix

waynebruce0x:


0xngmi:
would be good to convert it to sdk so we can refill it

0xngmi:
a bunch of the code is just doing basic stuff

PR #1183, opened 2021-12-24
Add methodology

ramverma-oro:


ramverma-oro:
Hi, Ethereum TVL is right. But BSC and Polygon token TVl is not getting properly it showas 0 for many tokens. So please help us to complete it.

Or should we create fetch adapter for BSC and Polygon? Please update on it.

PR #1186, opened 2021-12-25
add market.xyz tvl on fantom + polygon

0xDaksh:
[market-logos.zip](https://github.com/DefiLlama/DefiLlama-Adapters/files/7776017/market-logos.zip)

### DIFF
--- a/projects/marketxyz.js
+++ b/projects/marketxyz.js
@@ -1,0 +1,27 @@
+const axios = require("axios");
+const retry = require('async-retry')
+
+const getEthPrice = () => {
+  return retry(async (bail) => {
+    const {data} = await axios.get("https://api.coingecko.com/api/v3/simple/price?vs_currencies=usd&ids=ethereum");
+    return data.ethereum.usd;
+  })
+}
+
+ function fetchTVL(name) {
+   return async function() {
+    const pools = (await retry(async bail => await axios.get(`https://${name}.market.xyz/api/getAllPools`))).data;
+    const tvl = await getEthPrice() * (pools["2"].reduce((total, poolSupply) => total + parseInt(poolSupply), 0)/1e18);
+    return tvl;
+   }
+}
+
+module.exports = {
+  fantom: {
+    fetch: fetchTVL("fantom")
+  },
+  polygon: {
+    fetch: fetchTVL("polygon")
+  },
+  fetch: async () => (await fetchTVL("fantom")()) + (await fetchTVL("polygon")()),
+}


0xngmi:
aren't these comp forks? imo would be better to use the onchain comp helpers here

waynebruce0x:
I refactored into a (mostly) on chain adapter here: https://github.com/DefiLlama/DefiLlama-Adapters/pull/1197 reporting ~72M TVL. I took the MAI x Beefy locked TVL from your stats page though because it was easier than unwrapping the beefy curve LP tokens. Does it look right to you?

0xDaksh:
@waynebruce0x but why do you need to do it separately? You can just call the oracle on the compound pool and multiply with total supply of that token.

waynebruce0x:
> @waynebruce0x but why do you need to do it separately? You can just call the oracle on the compound pool and multiply with total supply of that token.

I'm not sure how to do that - would the contract be the comptroller at 0x296233b4f2FEf1Ce7977340cEb3cec4CbE3ada42? What function should I call, is the ABI the same as Compounds comptrollers? Do you mean it'll return USD per collateral token?

waynebruce0x:
The other vaults were okay because the tokens are listed on coingecko, which we use to get our token prices. The problem I was trying to get around was that mooCrvxxx tokens arent on coingecko. So we'd usually unwrap LP tokens into their unuderlying (eg 3crv => dai, usdt, usdc) to get the $ value of the LP tokens. This takes quite a few steps, especially with the moo wrapper on top, so I went the stats page route

0xDaksh:
Before I address the technical concerns, I was wondering why has market been listed on DeFiLlama with incorrect information and non working adapter?

@waynebruce0x

waynebruce0x:
Because we thought we'd write you an on-chain adapter, and get you listed. This happens often and usually teams are happy that we've helped. If you've got a preferred method then please let us know how to do it, or submit a PR for the changes.

0xDaksh:
No worries, here you go: https://github.com/DefiLlama/DefiLlama-Adapters/pull/1222

0xDaksh:
@waynebruce0x is this published on the website? Because I see a desprecency in the TVL showed on DeFillama and our internal dashboard.

0xngmi:
i just checked your pages at https://fantom.market.xyz/ and https://polygon.market.xyz/ and I'm seeing the same values that we report, you need to toggle "borrows" to have the values match since otherwise we only display collateral

0xDaksh:
yes @0xngmi, the collateral is the TVL we have. Which is higher than what is reported on DeFiLlama.

image

image

image

0xDaksh:
Hey @0xngmi and @waynebruce0x, any update on this? the difference now is very significant (over 16M) and we would prefer for it to be accounted as others are referring to this data!

PR #1197, opened 2021-12-27
Edits

waynebruce0x:


0xDaksh:
@waynebruce0x the way you have calculated is not correct. I’ll make a pr to change it.

0xDaksh:
FusePoolLens directly gives the total balance in each pool in ETH. That includes moo tokens as well.

You don’t have to individually calculate or depend on our metrics page which is less reliable and can cause wrong datapoints to show up in the defillama dashboard.

PR #1184, opened 2021-12-25
Implement adapter for Friktion.Fi ($40M TVL, Solana structured products)

smolbirdy:
We have a cron job on GitHub actions that should push the latest snapshot to that repo. Our script has careful santization to make sure only good data is on that file.

Since the protocol is really young and changing, we decided to just get data from the app since that would be more accurate. Due to how the friktion smart contracts are structured, the TVL info for each asset is actually in multiple accounts and changes every week.

##### Twitter Link:
https://twitter.com/friktion_labs

##### List of audit links if any:
None

##### Website Link:
https://friktion.fi/

##### Logo (High resolution, preferably in .svg and .png, for application on both white and black backgrounds. Will be shown with rounded borders):
logo

##### Current TVL:
```
--- tvl ---
solana                    11.41 M
msol                      10.32 M
bitcoin                   6.91 M
ftx-token                 137.77 k
serum                     133.13 k
ethereum                  121.23 k
Total: 29.03 M 

------ TVL ------

total                    29.03 M
```

##### Chain:
Solana

##### Coingecko ID (so your TVL can appear on Coingecko): (https://api.coingecko.com/api/v3/coins/list)
No token yet

##### Coinmarketcap ID (so your TVL can appear on Coinmarketcap): (https://api.coinmarketcap.com/data-api/v3/map/all?listing_status=active,inactive,untracked&start=1&limit=10000)
No token yet

##### Short Description (to be shown on DefiLlama):
Friktion is a crypto-asset portfolio management platform, built to perform across market cycles. Volts, Friktion’s native structured products, offer sustainable income generation and volatility yield strategies. Designed for individuals and DAOs.

##### Token address and ticker if any:
No token yet

##### Category (Yield/Dexes/Lending/Minting/Assets/Insurance/Options/Indexes/Staking) *Please choose only one:
Options

##### Oracle used (Chainlink/Band/API3/TWAP or any other that you are using):
N/A

##### forkedFrom (Does your project originate from another project):
original

##### methodology (what is being counted as tvl, how is tvl being calculated):
Methodology is in the `module.exports.methodology`.

Please feel free to change it. I drafted it up in 20 seconds and then decided to not look at it again.

PR #1202, opened 2021-12-27
Alternate adapter for Friktion.fi

smolbirdy:
See original pull request: https://github.com/DefiLlama/DefiLlama-Adapters/pull/1184

I made that first pull request because I didn't want to clog up DefiLlama-Adapters with puppeteer. It's up to you which one you want to take.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "node-html-parser": "^3.3.5",
     "node-schedule": "^1.3.3",
     "papaparse": "^5.3.1",
+    "puppeteer": "^13.0.1",
     "tronweb": "^3.2.7",
     "underscore": "^1.13.1",
     "web3": "^1.5.3"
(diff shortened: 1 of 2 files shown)


0xngmi:
could it be possible to use the script that is used to calculate window.friktionSnapshot ?

smolbirdy:
Unfortunately, that is not possible, because it is not a script. It is literally the entirety of our frontend. It's a spaghetti ball of code that depends on react hooks a lot.

It is possible to read from the blockchain, but that code is not written.

I pinky promise that the `window.friktionSnapshot` will be maintained through 2022. Because we are going to use the `mainnet-tvl-snapshots` (from the original pull request) to calculate historical analytics to be displayed in the app. So it's not just an afterthought.

I think it is better to use PR #1184 as opposed to this one #1202 because this one is MORE fragile.

Idealistically, it would be nice to have a DefiLlama specific adapter. Realistically, using #1184 will be more reliable and require less work from your end.

If you feel like writing a friktion adapter using only RPC requests and buffer decoding, let me know and i'll tell you the different accounts to read and how to find pointers to other accounts. But it'll quickly fall out of date.

0xngmi:
closing this pr since i merged the other as you recommended

PR #1221, opened 2021-12-29
Update index.js

meyers4549:
Updating existing adapter so it shows correct TVL

### DIFF
--- a/projects/tethys/index.js
+++ b/projects/tethys/index.js
@@ -1,23 +1,28 @@
-const {calculateUsdUniTvl} = require('../helper/getUsdUniTvl.js')
-const { staking } = require('../helper/staking.js');
-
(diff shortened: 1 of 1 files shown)


waynebruce0x:
Do the whitelisted tokens all have high liquidity? If tokens with lower liquidity are whitelisted then TVL becomes volatile - the function works by recursively calculating TVL of factory pools, starting with whitelisted tokens as a reference

0xngmi:
yeah, that's a very long list of whitelisted tokens, note that a token doesn't need to be whitelisted to be included into the tvl calculation, whitelisted is only used to find pairs. Script goes through all pairs that are against a whitelisted token or the core asset

waynebruce0x:
Closed since realshaman made an update instead. Please let us know in discord if you'd like this PR looking at again

PR #1234, opened 2021-12-30
Fixed wrong link for DODO Aurora

kashepavadan:
This project has already been added, see https://dodoex.io/ for more info.


0xngmi:
bsc subgraph url also seems to be broken, do you know what's the right url for it?

PR #1243, opened 2021-12-31
API3 staking

jo-chemla:
WIP: understand why API3 staked in Pool DAO is higher than API3 mcap

API3 staking

##### Twitter Link:
https://twitter.com/API3DAO

##### Website Link:
https://api3.eth.link/#/

dark bg svg: https://api3.eth.link/api3-logo-white.svg

##### Current TVL:
260M (staking)

##### Chain:
Ethereum mainnet

api3

7737

The Web3 API Economy: Create trustless applications that interact with Web APIs
Connect Smart Contracts to Real-World Data. API3 is leading the movement from legacy third-party oracle networks to first-party oracle solutions that deliver more security, efficiency, regulatory compliance, and simplicity

API3 0x0b38210ea11411557c13457d4da7dc6ea731b88a

Insurance / Other

API3 TVL is all API3 token staked in the API3 DAO Pool contract

### DIFF
--- a/projects/api3/abi.json
+++ b/projects/api3/abi.json
@@ -1,0 +1,7 @@
+{
+  "getLockedByGovernance": {"inputs":[],"name":"getLockedByGovernance","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"view","type":"function"},
+  "getLockedRewards": {"inputs":[],"name":"getLockedRewards","outputs":[{"internalType":"uint256","name":"totalLockedRewards","type":"uint256"}],"stateMutability":"view","type":"function"},
+  "getLockedVestings": {"inputs":[],"name":"getLockedVestings","outputs":[{"internalType":"uint256","name":"totalLockedVestings","type":"uint256"}],"stateMutability":"view","type":"function"},
+  "getTimelocked": {"inputs":[],"name":"getTimelocked","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"view","type":"function"},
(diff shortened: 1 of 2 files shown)


jo-chemla:
The reason is vested API3 tokens can also be part of the staking contract, and be slashed in case of a reimbursing event. Statement from their team: "Since they can be slashed, they are liquid, therefore should be part of the staking TVL"

0xngmi:
I think we shouldn't count vested tokens, my reasons are:
- it opens a whole can of worms (what if a new project pops up, counts a lot of the tokens that are not circulating and massively inflates their TVL that way).
- these vested tokens cannot be valued at market price, first because they are illiquid and second because there's enough of them for the market to crash if a part of them were to be sold
- it just makes no sense that staked TVL is higher than marketcap, it's not just a small amount of vested tokens but a huge one

Something similar already happened with sovryn and there we chose not to count staking for them.

I propose not counting their vested tokens here, if it's very hard to calculate this I'd just not list API3 for now.

jo-chemla:
@0xngmi yes totally understood. From the back of my mind, it was not easy to split vested and unvested, their team did not come up with a way to do it, need to investigate or ask the team again. A rule of thumb would be to apply the vested over total ratio to the staked token volume, which might as well be really inaccurate. 

Most metrics regarding vesting are available here, seems like 35M out of 48M API3 tokens are vested. If they display it on this page, then it is possible to query it https://enormous.cloud/dao/api3/tracker/wallets

jo-chemla:
Pushed a fix to remove vested tokens from staking TVL! This is retrieved from the contract `api3CirculatingSupply = "cD34bC5B03C954268d27c9Bc165a623c318bD0a8"`

PR #1261, opened 2022-01-03
refactor and update GMX

xyse:
Refactor Adapter to Pull Tokens from API Endpoint instead of hardcoding token addresses.
Add TVL for upcoming Avalanche launch.

After confirming with lead protocol dev I removed the bsc tvl of the previous GMX version called Gambit. Once GMX is deployed on BSC I will add it again.

### DIFF
--- a/projects/gmx/index.js
+++ b/projects/gmx/index.js
@@ -2,81 +2,75 @@ const sdk = require('@defillama/sdk');
 const axios = require('axios')
 const {staking} = require('../helper/staking')
 //Arbitrum
-const VAULT = '0x489ee077994B6658eAfA855C308275EAd8097C4A';
-const STAKING = '0x908C4D94D34924765f1eDc22A1DD098397c59dD4';
-const GMX = '0xfc5A1A6EB076a2C7aD06eD22C90d7E710E35ad0a';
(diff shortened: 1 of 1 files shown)


projects/gmx/index.js
```diff
@@ -2,83 +2,55 @@ const sdk = require('@defillama/sdk');
...
+          target: token.id,
+          params: [avalancheVault]
+      })),
       abi: 'erc20:balanceOf',
-      block: pChainBlocks.arbitrum,
-      chain: 'arbitrum'
-    });
-    return result.output*Math.pow(10, -decimals.output);
+      chain: 'avax',
+      block: chainBlocks.arbitrum
```
xyse, line 41: Just saw that avalanche tvl calculation seemed to work even though i specified block as chainBlocks.arbitrum? Can someone explain briefly, or should I change it to chainBlocks.avax?

waynebruce0x: you probably want chainBlocks.avax if you're reading data from the avax chain. chainBlocks.arbitrum will go through without errors because avax has a higher block number than arbitrum (at the moment at least). So you're basically reading old data on avax by accident

xyse: alright i will fix thanks for the explaination

PR #1298, opened 2022-01-07
killswitch fetch adapter

tanopwan:
{
id: "killswitch",
symbol: "ksw",
name: "KillSwitch"
}

KillSwitch
a smart yield aggregator
Users are free to unstake, withdraw Liquidity Position from their funds and sell their coins instantly in one click.

All funds that managed by KillSwitch contracts.

### DIFF
--- a/projects/killswitch/index.js
+++ b/projects/killswitch/index.js
@@ -1,0 +1,41 @@
+const utils = require("../helper/utils");
+
+const tvlUrl = "https://api.killswitch.finance/ksw/tvl";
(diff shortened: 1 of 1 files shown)


waynebruce0x:
Can this be calcuated using on-chain calls? We prefer this because its more transparent

tanopwan:
Actually the API is calculating using on-chain data but we cache it in the api for 15 seconds. Is this acceptable? @waynebruce0x

PR #575, opened 2021-09-24
add eCurve adapter

DenisCarriere:
##### Twitter Link:

- https://twitter.com/DAPPAccountDAO

##### List of audit links if any:

- N/A

##### Website Link:

- https://ecurve.finance/

##### Logo (High resolution, preferably in .svg and .png, for application on both white and black backgrounds. Will be shown with rounded borders):

https://ecurve.finance/static/media/ecurve-logo.e40d32e1.png

image

##### Current TVL:

- $5,169,846 USD

##### Chain:

- EOS

##### Coingecko ID (so your TVL can appear on Coingecko): (https://api.coingecko.com/api/v3/coins/list)

- N/A

##### Coinmarketcap ID (so your TVL can appear on Coinmarketcap): (https://api.coinmarketcap.com/data-api/v3/map/all?listing_status=active,inactive,untracked&start=1&limit=10000)

- N/A

##### Short Description (to be shown on DefiLlama):

eCurve is an exchange liquidity pool on EOS, designed for extremely efficient stablecoin trading and low risk, supplemental fee income for liquidity providers, without an opportunity cost.

##### Token address and ticker if any:

- 6,ECRV@ecurvetoken1

https://eosflare.io/token/ecurvetoken1/ECRV

##### Category (Yield/Dexes/Lending/Minting/Assets/Insurance/Options/Indexes/Staking) *Please choose only one:

- Dexes

##### Oracle used (Chainlink/Band/API3/TWAP or any other that you are using):

N/A

##### forkedFrom (Does your project originate from another project):

- CurveFi's StableSwap algorithm: https://curve.fi/files/stableswap-paper.pdf

PR #1306, opened 2022-01-07
disable eCurve

DenisCarriere:
Site & smart contract is currently disabled: https://ecurve.finance

Set TVL $0 until resolved or remove from TVL list

Ref: https://github.com/DefiLlama/DefiLlama-Adapters/pull/575

### DIFF
--- a/projects/ecurve/index.js
+++ b/projects/ecurve/index.js
@@ -1,10 +1,10 @@
-const utils = require("../helper/utils");
+// const utils = require("../helper/utils");
 
 async function fetch() {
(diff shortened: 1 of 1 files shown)


realdealshaman:
So, the adapter as-is, returns TVL that is not there?

DenisCarriere:
Correct, the API feed is no longer active, however still reporting stale data

Additionally, API is not reporting using on-chain balances, an updated adapter would need to be submitted to adjust to a fully on-chain TVL methodology (can submit when eCurve dapp resumes operations)

PR #1319, opened 2022-01-09
Argano, Prxy

waynebruce0x:


realdealshaman:
BTC prxy is already listed https://defillama.com/protocol/btc-proxy

PR #1323, opened 2022-01-10
update zoomswap tvl

zoomswapbot:


waynebruce0x:
can this change be made using on-chain calls instead?

waynebruce0x:
the problem is we cant see how you calculate the TVL - so it might not be in line with our methodology. What was wrong with the adapter before? Is there not a way we can update it?

PR #1324, opened 2022-01-10
update cyclone tvl

bf945:


waynebruce0x:
Can this change be made using on-chain calls instead? We prefer on-chain where possible for transparency

bf945:
Will try it update later. Cyclone is a privacy protocol; there is some difficulty to do it on chain.

bf945:
@waynebruce0x could you please review the PR? Thanks.

PR #1352, opened 2022-01-12
Delete projects/necc directory

rej156:
necc v1 is sunset for now, please remove

---

##### Chain: Aurora


rej156:
bootstrap process for liquidity for v2 is still TBD
and will differ to where v1 left off as remaining funds are reimbursed in the sunsetting process

would leave it for now I'd say
has to start over again from testnet etc too

PR #1363, opened 2022-01-13
add perifinance

Enitsed:
##### methodology (what is being counted as tvl, how is tvl being calculated): We have added all amount of minted pUSD as TVL from every mainnet.(for minting pUSD, users have to stake their PERI/USDC/DAI token on dApp)


waynebruce0x:
Is this measuring total supply of pUSD? can we count the locked collateral instead?

chaarlesjeong:
Yes, it's the same one. When the user has locked their collateral on the staking dApp, the contract issues synthetic asset(pUSD) to that user.

chaarlesjeong:
Okay, we've modified it for exporting the overall locked value. 
And we've also added TVL for each mainnet's value(ETH, BSC, Polygon). Check it please!

PR #1425, opened 2022-01-18
Agora

waynebruce0x:
Coingecko NETT to be valued. Not sure why its not picking up, maybe because Metis is a relatively new chain?

### DIFF
--- a/projects/helper/portedTokens.js
+++ b/projects/helper/portedTokens.js
@@ -482,6 +482,27 @@ async function transformKccAddress() {
   };
 }
 
+function transformMetisAddress() {
+  return (addr) => {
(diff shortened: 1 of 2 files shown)


projects/helper/portedTokens.js
```diff
@@ -482,6 +482,27 @@ async function transformKccAddress() {
   };
 }
 
+function transformMetisAddress() {
+  return (addr) => {
+    if (compareAddresses(addr, "0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000")) {
```
0xngmi, line 487: maybe would be better to use an object/dictionary to lower LoC

PR #1438, opened 2022-01-20
update Rubic project

kolebayevv:
- added moonriver USDC pool 
- tried to implement solana TVL, as discussed with @0xngmi in discord, there need some specific fix for solana chain


0xngmi:
fixed in 27af4334

PR #1445, opened 2022-01-20
Add support for yearn vaults

RadialFinance:
Add support for yearn vault tokens. Tested on Fantom against Radial.


0xngmi:
please add a flag parameter and only run this code if that flag is on, addFundsInMasterchef is used by a lot of adapters and this is non-backwards compatible

PR #1446, opened 2022-01-20
MoreMoney Project

BrunoEleodoro:
##### Methodology (what is being counted as tvl, how is tvl being calculated): TVL is collateral for loan and Stablecoin LP


projects/moremoney/index.js
```diff
@@ -0,0 +1,244 @@
+/*==================================================
+  Modules
+  ==================================================*/
+
+const sdk = require("@defillama/sdk");
+
+const IsolatedLending = "0xDc5CCAAA928De5D318605A76eEDE50f205Aa6D93";
+const CurvePoolRewards = "0x9727D535165e19590013bdDea8Fd85dd618b9aF7";
+const account = "0x0000000000000000000000000000000000000000";
+const tokenList = require("./tokenlist.json");
```
0xngmi, line 10: can we get the tokenlist in a way that auto-updates (eg: by querying smart contracts)? seems like this is likely to get stale with time

BrunoEleodoro: @0xngmi can you have a look again, just updated the code to query the contracts instead of searching for the info locally on the `tokenlist.json` file

BrunoEleodoro: Added back the `TokenList.json` so if the `decimals` is not available from the smartcontract, we have the `tokenlist.json` as a backup to lookup.

waynebruce0x:
How does totalValueLocked() on 0x9727D535165e19590013bdDea8Fd85dd618b9aF7 work? It looks like it counts total supply of tokens. We try to avoid this because it isnt actually counting TVL, just issuance. Can we not read locked tokens instead?

PR #1451, opened 2022-01-21
updated treasury balances

hazelnutcloud:

### DIFF
--- a/projects/exodia/index.js
+++ b/projects/exodia/index.js
@@ -77,7 +77,6 @@ async function tvl(timestamp, block, chainBlocks) {
 
   const treasuryBalances = await getPoolTokens(chainBlocks.fantom)
   const tokens = [mai, dai, wftm, gohm, exod, wsexod]
-  console.log(treasuryBalances)
 
   tokens.forEach(token => {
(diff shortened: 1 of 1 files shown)


waynebruce0x:
is there any particular reason you've changed to thegraph? We get difficulties with thegraph adapters when subgraphs fall behind current block height, so we try to avoid it.

hazelnutcloud:
yes we recently migrated our liquidity to beethoven and I figured trying to get all the balances in our pool through blockchain calls was a bit verbose. It would be a bit easier to maintain this way. Regarding the block height issues, I used timestamp instead of number to query the subgraph so it shouldn't be a problem.

0xngmi:
the problem still exists since if if it goes through it just means that we are fetching outdated values

hazelnutcloud:
alright I've changed from subgraph query to blockchain calls. appreciate if you could review again. thanks for the guidance!

0xngmi:
ty, really appreciated, sorry for earlier miscommunication about whether subgraphs were ok. Now there's increasing scrutiny on fantom adapters and we have a strong preference for on-chain ones

PR #1467, opened 2022-01-24
feat: avalanche network tvl

agAgent050:
Add Avalanche TVL to Augmented Finance


0xngmi:
@Nemusonaneko this contained an error that count avax addresses as if they were on bsc

Nemusonaneko:
> @Nemusonaneko this contained an error that count avax addresses as if they were on bsc

Must've missed that. My bad 🥲

PR #1476, opened 2022-01-25
updated ZENTEREST TVL calculations and included new staking pools

tomtsaomantra:


waynebruce0x:
Can you count borrows separately and export them as 'borrowed'. We count supply - borrows as core tvl

tomtsaomantra:
@waynebruce0x We have separated and exported borrows, thank you!

PR #1510, opened 2022-01-29
Add new project

end0cr1ne:
##### methodology (what is being counted as tvl, how is tvl being calculated): The value of the total supply of mintable items.

### DIFF
--- a/projects/kyber/index.js
+++ b/projects/kyber/index.js
@@ -108,6 +108,10 @@ const { transformAvaxAddress, fixAvaxBalances, transformFantomAddress } = requir
   async function ftmTVL(timestamp, ethBlock, chainBlocks) {
     return calcTvl(addr => `fantom:${addr}`, chainBlocks['fantom'], 'fantom', '0x78df70615ffc8066cc0887917f2Cd72092C86409', 0, true);
   }
+  // tracking TVL for KyberDMM Cronos
+  async function croTVL(timestamp, ethBlock, chainBlocks) {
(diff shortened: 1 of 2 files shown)


waynebruce0x:
Can we count the collateral / tokens locked in your contracts instead of measuring total supplies? At the moment, as far as I understand, it isn't really counting 'TVL'

end0cr1ne:
That's true though for a game I think this makes sense. All of these tokens are minted and burned within the game with no other use case. So any outstanding Gold tokens for instance, though in users wallets, can only be used to craft in-game items thus they are 'pseudo locked'. Besides, there isn't a staking mechanism so I'm not sure how else to calculate things.

waynebruce0x:
Our definition of TVL is total value of funds locked by users into a protocols smart contracts. It could be that your protocol has no TVL, in which case we cant list it. This is the case with quite a lot of gamefi protocols

end0cr1ne:
Ok, I've re-worked it to only include the value locked in the contract from the crafting mechanism.

waynebruce0x:
Hey I merged it but we're getting 0 TVL because MFF isn't on coingecko. Are there any UniV2 LP tokens with MFF in them?  That we can use to derive the price of MFF instead of using coingecko

end0cr1ne:
Yes there's this LP: https://explorer.mainnet.aurora.dev/token/0x3c508FC05C289BA989CF877bb2e3f6e54eF3fc95
Pool: https://www.trisolaris.io/#/swap?inputCurrency=0x5ce9F0B6AFb36135b5ddBF11705cEB65E634A9dC&outputCurrency=0x78B65477bBa78fc11735801D559C386611d07529

Thanks!

Pencab (commented):
end0cr1ne:main

PR #1527, opened 2022-02-01
using graph to get tvl

sameretto:
using graph for tvl

### DIFF
--- a/projects/hector/index.js
+++ b/projects/hector/index.js
@@ -1,92 +1,27 @@
-const { sumTokensAndLPsSharedOwners } = require("../helper/unwrapLPs");
-const sdk = require("@defillama/sdk");
-const erc20 = require("../helper/abis/erc20.json");
-const { transformFantomAddress } = require("../helper/portedTokens");
-const { getBlock } = require("../helper/getBlock");
+const { GraphQLClient, gql } = require("graphql-request");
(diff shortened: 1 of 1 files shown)


waynebruce0x:
we try to avoid thegraph because it gives us issues when it falls behind current block

sameretto:
so what can we do about this? We need this data

sameretto:
> we try to avoid thegraph because it gives us issues when it falls behind current block

sameretto:
> we try to avoid thegraph because it gives us issues when it falls behind current block

also its supported in your docs

waynebruce0x:
Yeah it is supported, but as I say, we try to avoid it because your adapter will be broken a lot. I only asked because it looks like it was previously calculated using on-chain calls. What was missing from the adapter before? Maybe I can help add it to the on-chain version

sameretto:
> Yeah it is supported, but as I say, we try to avoid it because your adapter will be broken a lot. I only asked because it looks like it was previously calculated using on-chain calls. What was missing from the adapter before? Maybe I can help add it to the on-chain version

yea we did previously support that but we have new data that cant be achieved with onchain calls due to the team that created it but our graph has it. This adds to our tvl which is why we would like this submitted even though sometimes it might be broken on visits.

PR #1544, opened 2022-02-03
update Bridge Mutual

solenemep:

TVL is now calculated based on virtual USDT in Capital Pool, which is more representative

### DIFF
--- a/projects/bridge-mutual/abi.json
+++ b/projects/bridge-mutual/abi.json
@@ -61,5 +61,18 @@
     ],
     "stateMutability": "view",
     "type": "function"
+  },
+  "virtualUsdtAccumulatedBalance": {
+    "inputs": [],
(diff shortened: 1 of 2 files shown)


waynebruce0x:
How is virtualUsdtAccumulatedBalance calculated?

solenemep:
It represents the usdt locked in protocol. So each time liquidity is added or withrawn, virtualUsdtAccumulatedBalance is updated accordingly. You can check CapitalPool.sol

waynebruce0x:
is this different to reading the USDT balance of the contract?

solenemep:
Yes because the protocol has a system of Leverage and Liquidity Cushion that disperse liquidity. Thanks to this function in CapitalPool, we got all USDT in protocol.

PR #1579, opened 2022-02-08
Update Celer cBridge adapter

Dominator008:
Include more bridge contracts and tokens

### DIFF
--- a/projects/celerbridge/index.js
+++ b/projects/celerbridge/index.js
@@ -2,21 +2,65 @@ const { getBlock } = require("../helper/getBlock")
 const { chainExports } = require("../helper/exports")
 const sdk = require('@defillama/sdk')
 const { sumTokens } = require("../helper/unwrapLPs")
+const ethers = require("ethers")
+const { config } = require('@defillama/sdk/build/api');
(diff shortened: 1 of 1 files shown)


projects/celerbridge/index.js
```diff
@@ -84,16 +428,28 @@ function chainTvl(chain) {
...
-            })
+            let balanceV1;
+            try {
+                balanceV1 = await sdk.api.erc20.balanceOf({
+                    chain,
+                    block: block,
+                    target: token[chain],
+                    owner: bridgeContractV1
+                })
+            } catch (error) {}
```
0xngmi, line 439: better to fail hard in case this fails

0xngmi:
why all the try/catches?

Dominator008:
> why all the try/catches?

The test seems flaky when I ran it. Getting call exceptions pretty frequently.

projects/celerbridge/index.js
Dominator008: Was getting `CALL_EXCEPTION`s pretty frequently but not deterministically. Any suggestion?

Dominator008:
> why all the try/catches?

Removed them. PTAL.

PR #1584, opened 2022-02-09
add ForTube Polygon and IoTeX data

jialianbao:


waynebruce0x:
whats the difference in balance_in_cheque and global_token_reserved?

jialianbao:
balance_in_cheque is the total deposit amount, and the global_token_reserved is Total value locked which is the data your site might need.

PR #1610, opened 2022-02-11
add impactMarket

obernardovieira:

### DIFF
--- a/projects/impact-market/index.js
+++ b/projects/impact-market/index.js
@@ -1,0 +1,36 @@
+const {getBlock} = require('../helper/getBlock')
+const sdk = require('@defillama/sdk')
+
+const CELOTokenAddress = "0x471EcE3750Da237f93B8E339c536989b8978a438"
+const PACTTokenAddress = "0x46c9757C5497c5B1f2eb73aE79b6B67D119B0B58"
+
+const pactDelegatorContract = "0x8f8BB984e652Cb8D0aa7C9D6712Ec2020EB1BAb4"
+
+async function tvl(timestamp, ethBlock, chainBlocks) {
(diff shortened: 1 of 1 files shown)


obernardovieira:
staking will be a different thing (to be released in some months from now). That PACT is a reserve.

waynebruce0x:
is the PACT in reserve all deposited by users? I can only find where to deposit cUSD on the UI

obernardovieira:
No, PACT on the reserve is there since deployment and there's no way to deposit more.
We will be adding staking along with other features in some months from now and by then it will be possible.

Yes, cUSD, but donate, not deposit (yes, deposited in a contract until someone else gets it, that's true).

waynebruce0x:
I dont think we should count PACT as TVL then because it doesnt really sound like theyre issued tokens

waynebruce0x:
Yeah actually if its your governance token it shouldn't be counted in treasury. TVL would return everything apart from staking and treasury, so in the example you gave, tvl would return an empty object, like:

module.exports={
    celo: {
        treasury,
        staking,
        tvl: () => ({})
    }
}

obernardovieira:
ahh I see. Thank you.
Given that I think makes sense that the locked PACT on the DAO is then considered only on TVL.
I've updated the code.

obernardovieira:
Hi @waynebruce0x 
I see that what was merged is different from my last submission. And the information on the website is different from what I shared here on the PR.
We do have an audit. And we don't have any staking, that locked PACT isn't staking as we've discussed here.
One of my partners submitted a form some time ago. Might have led to this confusion.

Could you please help me fix it? Thanks.

PR #1615, opened 2022-02-12
Add files via upload

BirdsBSC:
[https://github.com/DefiLlama/DefiLlama-Adapters/pull/1614#issue-1133283894](url)


waynebruce0x:
instead of counting the BIRDS in the pancakeswap pool I think it'd be better to count the LP tokens locked in the farms on https://farm.birdstoken.com/

PR #1648, opened 2022-02-16
Bugfix DeFi Saver

Cussone:
Add missing MATIC token


Cussone:
@waynebruce0x DeFi Saver's adapter is now working properly

PR #1655, opened 2022-02-16
Update UMA TVL

abg4:
1. Move Sherlock TVL to ethereum as it is inaccurately being included with polygon right now.
2. Add Polymarket TVL as they now use UMA to resolve their prediction markets.

### DIFF
--- a/projects/uma/index.js
+++ b/projects/uma/index.js
@@ -240,11 +240,25 @@ async function sherlock(timestamp, block) {
   return balances;
 }
 
+// Captures TVL for Polymarket liquidity that uses UMA's OO/DVM to resolve its markets
+async function polymarket(timestamp, block, chainBlocks) {
+  const conditionalTokensContract = '0x4D97DCd97eC945f40cF65F87097ACe5EA0476045'
+  const polygonUsdcContract = '0x2791bca1f2de4661ed88a30c99a7a9449aa84174'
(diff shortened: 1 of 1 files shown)


waynebruce0x:
We have a new Oracle dashboard, where we can include all the projects where UMA secures TVL - would that be more suitable?

abg4:
Just noticed the Oracle dashboard, looks great! We use the TVL endpoint for a few things. Would it impact how values are reflected in the API?

waynebruce0x:
Yes we can create an API for oracle data

waynebruce0x:
It might be better if we separate all the protocols in the UMA adapter, and put link them to UMA on the Oracle page. That way you can be compared directly against other oracles, increase your total value secured, and increase number of protocols secured. What do you think? I can give you a hand separating everything out if you want. TVL/TVS wont decrease, it just recatagorises UMA as an oracle instead of a protocol

abg4:
Yea, I think that makes a lot of sense and would be a better approach. Are there any docs or protocols that have made the change I could use to make the change?

waynebruce0x:
I can try separating out the adapter into each different product if you want? Then the UI will sort out everything on the Oracle page automatically

abg4:
That would be great! Is there anything I can do to help the process?

PR #1746, opened 2022-02-25
Add OnX multichain and TVL

vscode1111:


waynebruce0x:
Is it right that the lending TVL is (totalBorrow).plus(totalStake)  ? Is totalBorrow a negative value?

vscode1111:
> 

Currently I have these correct results
image

PR #1759, opened 2022-02-26
Remove Moonriver from Hundred temporarily

vfat0:
Moonriver does not work currently due to some RPC issue, causing Hundred's TVL not to update. 

Also xdai does not work currently but I don't know how to debug it:
`Error on token [object Promise], we'll just assume it's price is 0... TypeError: Cannot read properties of undefined (reading '[object promise]')`


waynebruce0x:
Yeah thats really weird, looks like an await is missing somewhere maybe. I'll have a look into this asap

PR #1763, opened 2022-02-27
updated treasury assets and tvl

sameretto:


waynebruce0x:
is torCurveLP counting the size of the TOR pool on curve or something else? For bank, can you export borrowed funds separately as 'borrowed'? Can 'staking' and 'borrowed' be excluded from the total TVL (fetch) function?

sameretto:
> is torCurveLP counting the size of the TOR pool on curve or something else? For bank, can you export borrowed funds separately as 'borrowed'? Can 'staking' and 'borrowed' be excluded from the total TVL (fetch) function?

Yes torCurveLP is the amount thats deposited in that LP.  For bank TVL we use the supplied and borrowed amount as the total tvl because those are the funds that currently used for it to operate. Why would we exclude it 'staking' and 'borrowed' from the total tvl (fetch)? The total fetch is the TVL where funds are either staked, borrowed, and any remaining sitting in our treasury.

sameretto:
Audit to be added 
https://www.certik.com/projects/hector-finance

realdealshaman:
Our goal is to give visibility to everything that can be considered TVL, with that in mind, we also want the values ​​to have a clear categorization. With this in play, each user can decide which categories they want to count or exclude from what they see in the totals as there is debate about what should be considered TVL

realdealshaman:
I have now added the audit link to the listing, this will be visible with the next hourly update

sameretto:
> Our goal is to give visibility to everything that can be considered TVL, with that in mind, we also want the values ​​to have a clear categorization. With this in play, each user can decide which categories they want to count or exclude from what they see in the totals as there is debate about what should be considered TVL

understood i have made changes regarding this. Is there an issue with the values or can we proceed?

PR #1795, opened 2022-03-02
Update Minswap TVL function

longngn:
Firstly, thanks @0xngmi for writing a TVL function for us!
I would like to update the function though. Since a constant-product liquidity pool consists of an equal amount of 2 tokens, the total value locked would be equal to the total value of 2 tokens
We already have the total ADA amount. The MIN/ADA pool on Minswap would be the most liquid and reflecting market price for MIN, so we can take double the total locked ADA to arrive at the TVL of MIN/ADA pool.

### DIFF
--- a/projects/minswap/index.js
+++ b/projects/minswap/index.js
@@ -1,19 +1,34 @@
-const axios = require('axios')
+const axios = require("axios");
 
-async function tvl(){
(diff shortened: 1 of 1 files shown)


0xngmi:
AFAIK durint LBE MIN is not tradeable, and, more importantly, it cannot be deposited by users (afaik it's just minted), that's why I didn't consider it as TVL.

0xngmi:
we can update this after the end of the LBE period, when protocol goes live

longngn:
> we can update this after the end of the LBE period, when protocol goes live

Hi, I've updated the function for the mainnet smart contracts

0xngmi:
ty

PR #1808, opened 2022-03-03
Add TVL for The Parallel

theparalleldotio:


waynebruce0x:
What is locked PRL? how does it get into the locked contract? 

Can the TVL from PRL in the mining contract be exported as a separate 'staking' export?

theparalleldotio:
> What is locked PRL? how does it get into the locked contract?
> 
> Can the TVL from PRL in the mining contract be exported as a separate 'staking' export?

People stake PRL to receive tickets used to receive special benefits of PRL such as early purchase of NFT Box or participation in special events.

PR #1833, opened 2022-03-07
Knit finance

sruthi23:

knitlogo

##### Chain:
ETH
BSC
MATIC
FANTOM
HECO
MOONBEAM
AVALANCHE

https://www.coingecko.com/en/coins/knit-finance

##### Short Description (to be shown on DefiLlama): Multi Chain Wrapped Assets ( Coins and NFTs) and Liquid stacking protocol for all the assets.

Developed Single infrastructure which will allow various assets , liquidity to move from one chain to another seamlessely as well enable liquidity staking for DPoS assets

Bridge

##### methodology (what is being counted as tvl, how is tvl being calculated): Value of total coins locked with us


realdealshaman:
@sruthi23 Hello, I was seeing that your adapter has not been updated after the most recent changes. Apparently the calculation is combining the TVL of all the blockchains in a single export and this is causing problems on the server since in its listing it states that it is live on more than one blockchain, it also returns a total of more than 2 billion TVL I think this might not be correct. Would it be possible to update it so that the TVL of each blockchain is reflected in its own export? or should it all fall under one blockchain?

PR #1847, opened 2022-03-08
Update adapter overnight.fi

Vorh:
Hello! 
We have updated the core of the project, so the calculation of TVL has changed. 
Also been audited by Hacken.
Is it possible to add it to the page?
Audit link: https://github.com/ovnstable/ovnstable-app/blob/master/audits/hacken_2022-02-03.pdf


waynebruce0x:
fao front end devs, I've added the audit link already

PR #1887, opened 2022-03-11
add babena.finance

rodsouto:
I picked babena.finance from the list of requested projects. 

The TVL returned by the adapter is the same that they show in the website (http://babena.finance/stats)

```
------ TVL ------

total                    13.27 M
```

I added `pact-lang-api` as dependency to be able to fetch data from `chainweb` API.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "node-fetch": "^3.1.0",
     "node-html-parser": "^3.3.5",
     "node-schedule": "^1.3.3",
+    "pact-lang-api": "^4.3.3",
     "papaparse": "^5.3.1",
     "tronweb": "^3.2.7",
     "underscore": "^1.13.1",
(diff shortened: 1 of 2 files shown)


waynebruce0x:
thanks for the PR. Its a bit unconventional, but we try to avoid try/catch blocks so that we can easily tell when an adapter stops working. We dont check the console logs on any adapter runs but the system flags any scripts that don't finish successfully. Do you mind removing some of the safeguard?

rodsouto:
Sure, I can change that. 

Is it ok to throw an error if the API call fails or it should fail silently setting price=0?

waynebruce0x:
Yeah I'd still throw an error. If price is set to 0, I guess TVL would return as 0, and we wouldn't think to investigate it, as quite a few working adapters return 0 TVL. So yeah basically we like adapter scripts to fail in every case apart from behaving as expected

PR #1889, opened 2022-03-11
update Serum adapter

sayantank:
Updated the Serum fetch adapter to point to a newer endpoint (https://serum-volume-tracker.vercel.app) for tracking TVL.


0xngmi:
@sayantank can we get some verification that is is actually correct?

PR #1891, opened 2022-03-11
Update Gamblefi

ravirasadiya:
Gamblefi was listed Yesterday and as we can see right now few Infos are wrong on page like token symbol and address is concat with cronos: 

Please let me know what exactly should do to make it perfect!


waynebruce0x:
we prefix all token addresses with the chain name so that users can more easily find the token. We'll change the symbol to BETIFY. Let me know if theres anything else

PR #1894, opened 2022-03-11
Update of Ankr project and refactoring of OnX one

vscode1111:


projects/config/ankr/ethereum/index.js
```diff
@@ -0,0 +1,39 @@
...
+  console.log(111, `aETHc ${totalSupply}`);
+  // return { [tokenAddresses.aETHc]: totalSupply };
+  return { [wethAddress]: totalSupply };
+}
+
+const getaMATICbTvl = async () => {
+  const totalSupply = await aMATICbTokenContract.methods.totalSupply().call();
+  console.log(111, `aMATICb ${totalSupply}`)
+  // return { [tokenAddresses.aMATICb]: totalSupply };
+  return { [wethAddress]: totalSupply };
```
0xngmi, line 17: this should be matic instead of weth

projects/config/ankr/ethereum/index.js
```diff
@@ -0,0 +1,39 @@
...
+const wethAddress = '0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2'
+
+const getaETHcTvl = async () => {
+  const totalSupply = await aETHcTokenContract.methods.totalSupply().call();
+  console.log(111, `aETHc ${totalSupply}`);
+  // return { [tokenAddresses.aETHc]: totalSupply };
+  return { [wethAddress]: totalSupply };
+}
+
+const getaMATICbTvl = async () => {
```
0xngmi, line 13: this:
- should use the sdk
- should use block height provided instead of none

0xngmi:
why move onx files into ankr's config directory? @vscode1111

0xngmi:
moving stuff into the helper repo is also polluting it

0xngmi:
I've merged another update that greatly simplifies the code, makes it all self-contained, and works with historical data. I've put all tokens under ethereum since that's the network where users stake their tokens

PR #1903, opened 2022-03-12
update : sRHEA V2 migration

RHEADAO:
sRHEA V2 migration 

Change address


0xngmi:
better to keep both addresses to have historical data

RHEADAO:
sREHA V1 Address added 👍

PR #1917, opened 2022-03-14
METF Finance

croc12344:
##### methodology (what is being counted as tvl, how is tvl being calculated): Total amount of staked assets in our staking pools

METF Token (2)

### DIFF
--- a/projects/metf-finance/TreasuryContract.json
+++ b/projects/metf-finance/TreasuryContract.json
@@ -1,0 +1,19 @@
+{
+  "inputs": [
+    {
+      "internalType": "address",
+      "name": "token",
+      "type": "address"
+    }
+  ],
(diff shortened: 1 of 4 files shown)


croc12344:
> Also count you get rid of the try/catch? We like adapters to fail instead of log errors. It makes it easier for us to find bugs

Fixed 

> Can the TVL from METF-MMF LP be exported as 'pool2'? 

Done 

> Do you know on which contracts ( can Iind the source code for getPairPrice() and valueOfAsset() functions? I'm struggling to find them through the proxys, sorry

We implemented this custom function on our contract. Its a way to calculate pair price. This allows us to easily derive the pair price in the frontend, i.e., totalLP / total number of MMF in LP * priceOfMMF * 2.

@waynebruce0x Thanks :)

PR #1960, opened 2022-03-17
Update ValleySwap project config

ValleySwapDeployer:
This is update adapter request.
We have started the migration process to follow Paladin auditors requirements.
And adapter should use all LP versions.


llamatester:
The adapter at projects/valleyswap/index.js exports TVL: 
                
 
 
oasis                     126.79 M

total                    126.79 M

PR #1962, opened 2022-03-17
Update Zunami main contract address

MioGreen:


g1nt0ki:
This code has a bug, instead of taking the block from function argument, it is using a constant `const startBlock = 14255781`

0xngmi:
current chart, i guess all caused by the constant block
image

waynebruce0x:
I PRd to fix it

MioGreen:
@MioGreen for now we are going to be deployed only on Ethereum.

PR #1976, opened 2022-03-18
excalibur: adapt UST address

percivalExcalibur:
Made so that $UST value is correctly fetched


llamatester:
The adapter at projects/excalibur/index.js exports TVL: 
                
 
 
fantom                    24.00 M

total                    24.00 M

PR #1990, opened 2022-03-21
remove staking contract

ckLee8:
I had to revert changes I made to staking calculation because the migration was to an address that also has escrowed tokens. There doesn't appear to be any simple way to filter these out with a subgraph.


llamatester:
The adapter at projects/tokemak/index.js exports TVL: 
                
 
 
ethereum-staking          14.42 M
ethereum                  235.25 M
ethereum-pool2            232.08 M
staking                   14.42 M
pool2                     232.08 M

total                    235.25 M

PR #2002, opened 2022-03-22
Add Llama Airforce cvxFXS pounder

0xAlunara:
I'm not sure if DefiLlama has a properly working cvxFXS oracle yet, but this PR adds the cvxFXS pounder TVL to the Llama Airforce project.

I've made this PR in hackerman style directly through Github, so naturally it's untested.

##### Twitter Link:
https://twitter.com/0xAlunara

##### Website Link:
https://llama.airforce/#/union/pounders

##### Current TVL:
$961.0k (the cvxFXS pounder)

##### Chain:
Ethereum

### DIFF
--- a/projects/llama-airforce/index.js
+++ b/projects/llama-airforce/index.js
@@ -6,6 +6,11 @@ async function tvl(time, block){
             target: "0x83507cc8c8b67ed48badd1f59f684d5d02884c81",
             abi: {"inputs":[],"name":"totalUnderlying","outputs":[{"internalType":"uint256","name":"total","type":"uint256"}],"stateMutability":"view","type":"function"},
             block
+        })).output,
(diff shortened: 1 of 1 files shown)


waynebruce0x:
We can merge it, but since cvxFXS isnt on CoinGecko we can't value it and it wont be added to TVL. Once CoinGecko has a price it'll automatically get included. Alternatively, if theres a way to calculate its relative price through contract calls (eg uniV2 pool weights etc) we can use that

0xAlunara:
The only source of liquidity atm with a TVL of $116m is the cvxFXS/FXS pool on Curve: https://curve.fi/factory-crypto/18

This is a v2 pool which has an internal price oracle, so to get the price of cvxFXS you can multiply the FXS price (you can get this one from CoinGecko I think) with `price_oracle / 10^18` from https://etherscan.io/address/0xd658a338613198204dca1143ac3f01a722b5d94a.

PR #2018, opened 2022-03-23
add staked near realm to the calculation

pysrbastion:
Bastion launch staked near realm yesterday. Add the new realm's comptroller to the calculation.


llamatester:
The adapter at projects/bastion/index.js exports TVL: 
                
 
 
aurora-borrowed           136.96 M
aurora                    108.47 M
borrowed                  136.96 M

total                    108.47 M

PR #2033, opened 2022-03-24
add aurora

wepiggy-dev:


waynebruce0x:
Error: missing revert data in call exception [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] (code=CALL_EXCEPTION, version=providers/5.6.1)

Are you sure this is the right address / ABI?

0xngmi:
tested locally and it seemed fine

PR #2069, opened 2022-03-28
Implemented Trustednode DefiLlama SDK adapter

DTverdov:

### DIFF
--- a/projects/trustednode/index.js
+++ b/projects/trustednode/index.js
@@ -1,0 +1,91 @@
+const sdk = require('@defillama/sdk')
+const { getChainTransform } = require('../helper/portedTokens')
+const { unwrapUniswapLPs } = require('../helper/unwrapLPs')
+const bn = require("bn.js")
+
+const bscVaults = [
+    {
+        name: "TNODE VAULT",
+        chain: "bsc",
+        vaultAddr: "0x98386F210af731ECbeE7cbbA12C47A8E65bC8856",
+        contractAddr: "0x7f12a37b6921ffac11fab16338b3ae67ee0c462b",
+        isLp: false,
+    },
+    {
+        name: "BSC LIQUIDITY VAULT TNODE/BUSD",
+        chain: "bsc",
+        vaultAddr: "0x44dC7FE8e51076De1B9f863138107148b441853C",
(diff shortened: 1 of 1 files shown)


llamatester:
The adapter at projects/trustednode/index.js exports TVL: 
        
 
 
    bsc                       503.66 k
    fantom                    278.34 k
    
    total                    782.00 k

projects/trustednode/index.js
```diff
@@ -0,0 +1,91 @@
...
+        isLp: true,
+    },
+]
+
+async function vaultTvl(timesatamp, block, chainBlocks, vaults) {
+    const balances = {}
+
+    for (const vault of vaults) {
+        const totalSupply = (await sdk.api.abi.call({
+            abi: 'erc20:totalSupply',
```
waynebruce0x, line 38: can we count deposits to the vault instead of total supply?

JGJP: Hi @waynebruce0x I'm TL at Trusted Node, this function `totalSupply` on our vaults contracts has the same name as the ERC20 method, but it actually returns the token amounts deposited by users. It's required for differentiating between assets in the vault owned by users, and assets in the vault owned by the vault (rewards that have yet to be distributed). The distinction wouldn't be necessary in the case that the staking token isn't the same as the rewards token, but for one of our vaults they are the same.

To clear up confusion I propose that we would specify our own ABI for this method, instead of reusing the ERC20 one, even if the end effect is the same), or add a comment here specifying that this totalSupply call is returning something different. What do you think?

waynebruce0x:
Hey @JGJP any update?

DTverdov:
> Hey @JGJP any update?

Hi @waynebruce0x at that moment we verified vault contract on Fantom Opera only (https://ftmscan.com/address/0xe056aba40572f64d98a8c8e717c34e96056c4aad#code), two others on Binance Smart Chain not verified yet. We will let you know once they will be verified.

DTverdov:
Hi @waynebruce0x vaults on BSC also verified (https://bscscan.com/address/0x98386f210af731ecbee7cbba12c47a8e65bc8856#code, https://bscscan.com/address/0x44dc7fe8e51076de1b9f863138107148b441853c#code)

PR #2084, opened 2022-03-29
add TokensFarm adapter

casualuser:
https://drive.google.com/file/d/1yrWcvbguLfTF2QsgTZpCaU4-qkhbWJq2/view

##### methodology (what is being counted as tvl, how is tvl being calculated): The worth in $ of all tokens on each farm contract

### DIFF
--- a/projects/tokensfarm/index.js
+++ b/projects/tokensfarm/index.js
@@ -1,0 +1,13 @@
+const retry = require('async-retry')
+const axios = require("axios");
+
+async function fetch() {
+    let response = await retry(async bail => await axios.get('https://api.tokensfarm.com/totals'))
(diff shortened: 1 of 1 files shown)


llamatester:
The adapter at projects/tokensfarm/index.js exports TVL: 
        
 
 
    
    total                    23.94 M

waynebruce0x:
Can TVL be broken down chain by chain?

casualuser:
@waynebruce0x at the moment there is no possibility to split TVL by chain because it was calculated altogether
it may be implemented but it will take some time

please let us know if we can update this later

waynebruce0x:
we have to assign TVL to a chain in our database so we're unable to list as is

PR #2089, opened 2022-03-29
scion-finance add fantom to tvl

0xSaga:
add fantom to tvl


llamatester:
The adapter at projects/scion-finance/index.js exports TVL: 
        
 
 
    fantom                    101.9266641965
    moonriver                 224.88 k
    
    total                    224.99 k

PR #2135, opened 2022-04-01
update Polaris Finance with LUNAR

Cocoahomology:
Added the new LUNAR LPs, and the new staking pool for SPOLAR.

TVL is now up-to-date and accurate except it's still missing the PBOND LP, which I'm not sure how to price.

PR #2146, opened 2022-04-02
update Polaris Finance with LUNAR

eYdr1en:
Added new LUNAR LPs with POLAR single staking, and single stake for SPOLAR that earns LUNAR

Also TVL is updated to work so total won't be 0 and we can use DefiLlama's api to fetch TVL

Fixes total 0 in #2135


llamatester:
The adapter at projects/polaris-finance/index.js exports TVL: 
        
 
 
    aurora-staking            1.05 M
    aurora-pool2              1.77 M
    aurora                    2.82 M
    staking                   1.05 M
    pool2                     1.77 M
    
    total                    2.82 M

waynebruce0x:
We dont include pool2 and staking tvl as part of core tvl, so that users can choose whether or not they include it in the dashboard data. You can fetch pool2 and staking data from our /protocols/ endpoint though

PR #2173, opened 2022-04-05
Minor updates to KogeFarm vaults

kogecoindev:
Hi! This is to provide some minor updates to KogeFarm vaults which doesn't have a huge impact in our TVL but is for completeness.

One concern we'd like to raise: recently our DefiLlama TVL has been collapsing from ~$6 million to ~$600k once in a while, when nothing drastic is happening on chain and our internal TVL metrics are constant. It happened on March 26, fixed itself, and again today on April 5. The test.js projects/kogefarm test script returns the right value. 

If this issue can be fixed I'd appreciate it very much as we display the DefiLlama TVL in our main site and it is a key KPI for us.


kogecoindev:
Sorry this request was sent in error

PR #2200, opened 2022-04-07
use correct endpoint for oxygen project

evgeniymelnikov:
Have started using correct endpoint, previous doesn't cover need about tvl calculation.

### DIFF
--- a/projects/oxygen/index.js
+++ b/projects/oxygen/index.js
@@ -1,14 +1,10 @@
 const axios = require("axios")
 const { toUSDTBalances } = require('../helper/balances');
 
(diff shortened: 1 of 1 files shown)


llamatester:
The adapter at projects/oxygen/index.js exports TVL: 
        
 
 
    solana                    25.83 M
    
    total                    25.83 M

waynebruce0x:
What extra TVL is included in this endpoint? is the API open source?

at-convexitydmcc:
We've recently launched Oxygen Light version of the protocol light.oxygen.org . The TVL is a sum of TVL (app.oxygen.org) & TVL (light.oxygen.org)

waynebruce0x:
Are they separate products/contracts? I'm just wondering if there would be any double count

at-convexitydmcc:
They are separate, different Solana programs. There is no double count.

PR #2204, opened 2022-04-07
DeversiFi: Include TVL from wrapped aust on ethereum

plutoegg:


llamatester:
Error while running adapter at projects/deversifi/index.js: 
        
 
 
    
            module.exports = {
              methodlogy: '...',
              ethereum: {
                tvl: 
                pool2:
              }
            }

llamatester:
The adapter at projects/deversifi/index.js exports TVL: 
        
 
 
    ethereum                  77.50 M
    
    total                    77.50 M

PR #2213, opened 2022-04-08
Modify Parallel Finance endpoint url

alvaroparallel:

### DIFF
--- a/projects/parallel.js
+++ b/projects/parallel.js
@@ -4,11 +4,11 @@ const axios = require('axios');
 async function fetch() {
         const response = (
             await retry(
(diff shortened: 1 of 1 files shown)


alvaroparallel:
We're providing the new url for this endpoint. This new endpoint include all our products tvl (Crowdloan TVL, Amm TVL,  Money Market TVL, etc.)

As a plus If you're able to help us to change the current name to **"Parallel DeFi Super App"** instead of **"Parallel Liquid Crowdloan"** would be great, Cuz at this moment we're not only showing the liquid crowdloan's TVL.

llamatester:
The adapter at projects/parallel.js exports TVL: 
        
 
 
    
    total                    650.68 M

PR #2237, opened 2022-04-10
feat(primitive-rmm): adds weth-usdc rmm tvl to primitive adapter

Alexangelj:
#### Description of changes:

Adds the WETH-USDC pair TVL using `balanceOf` calls for WETH and USDC. 

For each token pair, a PrimitiveEngine contract is deployed. Pools can be created within this PrimitiveEngine contract for that token pair. Additionally, the pair's tokens can be deposited into an internal balance. Thus, a single PrimitiveEngine holds multiple pools of the same tokens and internal balances. You can read more [here](https://library.primitive.xyz/faq/usage/creating).

This is the first pair of the Replicating Market Maker protocol.

In 1-2 weeks we will make another PR that will track all pairs and pools once the Primitive subgraph is in production.

### DIFF
--- a/projects/primitive/index.js
+++ b/projects/primitive/index.js
@@ -1,23 +1,14 @@
-const BigNumber = require('bignumber.js')
-const v1TVL = require('./v1')
+const rmmTVL = require('./rmm')
 
 async function tvl(timestamp, block) {
-  const [v1] = await Promise.all([v1TVL(timestamp, block)])
-
-  const tokenAddresses = Object.keys(v1)
-  const balances = Array.from(tokenAddresses).reduce(
-    (accumulator, tokenAddress) => {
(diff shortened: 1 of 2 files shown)


0xngmi:
could you change the adapter to keep v1? so tvl would be combination of v1 + rmm, that way we can also run it historically

llamatester:
The adapter at projects/primitive/index.js exports TVL: 
        
 
 
    ethereum                  42.60 k
    
    total                    42.60 k

Alexangelj:
> could you change the adapter to keep v1? so tvl would be combination of v1 + rmm, that way we can also run it historically

Yes that makes sense!

llamatester:
The adapter at projects/primitive/index.js exports TVL: 
        
 
 
    ethereum                  69.79 k
    
    total                    69.79 k

PR #2350, opened 2022-04-21
ENH: ignite-finance tvl

ash-ignite:
Variant12
Variant14


waynebruce0x:
is it possible to calculate TVL using token balances instead of reading totalSupply?

ash-ignite:
sTinder stands for staked TINDER. So we get exactly the same number. 

Do I need to change my code?

PR #2376, opened 2022-04-23
Update Orca adapter to include whirlpools TVL

philcchen:
Orca recently released a new set of concentrated liquidity pools and would like to update the adapter to reflect these new pools.


llamatester:
The adapter at projects/orca exports TVL: 
        
 
 
    
    total                    403.74 M

PR #2459, opened 2022-04-28
Wormhole - add Aurora TVL

justinschuldt:
Adds TVL data for the Aurora chain.


justinschuldt:
@waynebruce0x just adds data for the newest chain added to Wormhole.

PR #2506, opened 2022-05-03
Add new mappings for the new tokens on Blindex (due to the fact they are not yet on CoinGecko)

kazazor:
This PR only updates the already listed Blindex protocol, therefore I removed all the fields that are not needed again.
However, we do want to change the short description of Blindex, so adding the new one here again.

The core reason for this PR is the fact new tokens were added to the Blindex platform. Since it runs on RSK (which has difficulties being added to Coingecko at the moment because it lacks The Graph on the chain), the approach here is mapping the stables to the corresponding ones on CG.

That was the approach that was agreed upon here, so just updating it.

Blindex is a multi-asset stablecoin DeFi platform, on our journey to Tokenize Everything.
Mint, trade, invest and redeem any pegged asset.
All assets are partially collateralized, algorithmically stabilized and completely decentralized


llamatester:
The adapter at projects/blindex exports TVL: 
        
 
 
    rsk                       2.03 M
    
    total                    2.03 M

PR #2530, opened 2022-05-04
Fix tvl and add borrowed for gearbox

harsh-98:
##### Twitter Link:
[Gearboxprotocol](https://twitter.com/GearboxProtocol)

Before only available liquidity for gearbox protocol was accounted for, this ignored borrowed amount. Fix tvl and add total borrowed per pool. 

Tested.
Screenshot 2022-05-04 at 11 37 20 PM


waynebruce0x:
Can you revert the change to core TVL? We only count liquidity (supply - borrow), and then the borrowed funds can be added/removed by users choice

harsh-98:
@waynebruce0x reverted to old tvl method, this PR adds borrowed function.

llamatester:
The adapter at projects/gearbox exports TVL: 
        
 
 
    ethereum-borrowed         2.67 M
    ethereum                  12.63 M
    borrowed                  2.67 M
    
    total                    12.63 M

PR #2572, opened 2022-05-06
Updated the MahaDAO pools

senamakel:
We've added two more pools for which our stablecoin ARTH is also a part of the LP. 

(as per the definition of what should be counted in pool2, we take into account two kinds of LPs, one that uses ARTH (our stablecoin) and MAHA (our gov. token)).


llamatester:
The adapter at projects/mahadao exports TVL: 
        
 
 
    ethereum-pool2            286.23 k
    ethereum                  261.10661999999996
    bsc-pool2                 1.26 M
    polygon-staking           1.85 M
    bsc                       767.66 k
    polygon                   709.32 k
    polygon-pool2             417.61 k
    pool2                     1.97 M
    staking                   1.85 M
    

PR #2613, opened 2022-05-09
Including New Farm Pool

0xAces:
We recently deployed a new Curve YUSD LP Farm Pool.

Added the new pool to the adapter. :)


0xAces:
Will resubmit after resolve conflict

PR #2617, opened 2022-05-10
added deposited KSD to tvl

pureicis:
##### Short Description (to be shown on DefiLlama): Kokoa Finance is a cryptoasset-backed stablecoin platform, providing interest-free loans and low-risk savings products, native on Klaytn network.
Kokoa Stable Dollar(KSD) is the first and dominant stablecoin native to Klaytn network. 

##### methodology (what is being counted as tvl, how is tvl being calculated): tvl is calculated using the total collateral value locked in the protocol plus the deposited KSD. Staked tokens include staked kokoa value. Pool2 includes staked LP tokens

### DIFF
--- a/projects/kokoa-finance/DKSD.json
+++ b/projects/kokoa-finance/DKSD.json
@@ -1,0 +1,978 @@
+{
+  "address": "0x5e6215DFB33b1Fb71E48000A47Ed2eBb86d5Bf3d",
+  "abi": [
+    {
+      "anonymous": false,
+      "inputs": [
+        {
+          "indexed": false,
+          "internalType": "address",
+          "name": "previousAdmin",
+          "type": "address"
+        },
+        {
+          "indexed": false,
+          "internalType": "address",
+          "name": "newAdmin",
+          "type": "address"
+        }
+      ],
+      "name": "AdminChanged",
+      "type": "event"
+    },
+    {
+      "anonymous": false,
+      "inputs": [
+        {
+          "indexed": true,
+          "internalType": "address",
+          "name": "implementation",
+          "type": "address"
+        }
+      ],
+      "name": "Upgraded",
+      "type": "event"
+    },
+    {
+      "stateMutability": "payable",
+      "type": "fallback"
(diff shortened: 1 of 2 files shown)


llamatester:
The adapter at projects/kokoa-finance exports TVL: 
        
 
 
    klaytn-staking            4.55 M
    klaytn                    31.43 M
    klaytn-pool2              4.12 M
    staking                   4.55 M
    pool2                     4.12 M
    
    total                    31.43 M

waynebruce0x:
The collateral counted in fetchCollateral is backing the minted KSD - is that correct?

pureicis:
Yes it is, but it is also locked up again as a deposit feature. (if users
choose to do so)
It's like Makerdao and Anchor protocol in one, so we think the deposited
KSD should also be counted as tvl just as deposited UST is considered tvl
of Anchor! :)

2022년 5월 10일 (화) 오후 6:59, waynebruce0x ***@***.***>님이 작성:

waynebruce0x:
Counting the minted KSD would be double counting the underlying collateral. It's effectively repackaging and recounting the same coins, so we cant do that

pureicis:
Yes, I do understand the concerns for double counting,
but please do take into account that we are not simply recounting all the
KSDs that have been minted based on the collaterals, but only the one's
deposited.

Our "Earn" Product(which is a separate product from our BorrowVault) offers
savings feature of dynamic rate(currently above 16%).
Users who have minted KSD(or bought them from the market) are voluntarily
locking them for the additional yields that we provide.


Moreover, apart from the TVL calculation, could you update the Logo of
Kokoa Finance to :
https://drive.google.com/file/d/1Z8mAm5nL4gbGnkS_jTvnlekKxfoFkhA4/view?usp=sharing
and also our representing Token from $KSD to $KOKOA?

Thanks!

waynebruce0x:
Yes we will update the logo but I'm afraid we can't merge this PR because it is a double count of the collateral - if you want to discuss this further please message us on discord

PR #2619, opened 2022-05-10
feat: keep3r network adapter

wei3erHase:
ey llamas!

we're pushing this changes to fix the issue with 'Coingecko not listing ibX tokens', making all of ibEUR, ibJPY, ibGBP, ... to appear with 0 value locked.

the team has managed to make a `fetch` method, to quote each token from it's correspondant Curve pool, but we've faced some issues with the `borrow` and `staked` methods, as we don't finally see how to implement the same `fetch` mechanism for these metrics. a hand here could come quite handy `<3`

also, we moved the code to it's own folder for better repo hygiene, and implemented some ts-check on the way for easier DX

please just ping us if there's something odd with any,
THANKS for your all your work!


llamatester:
The adapter at projects/keep3r exports TVL: 
        
 
 
    ethereum-staking          18.55 M
    ethereum-borrowed         126.20 M
    staking                   18.55 M
    borrowed                  126.20 M
    
    total                    581.72 M

PR #2674, opened 2022-05-13
[Muuu Finance] fix: loop logic

shuuuting95:
- Add new pool
- Bug Fix

We expected this result as Array, but the actual one is Object. Thus we fixed loop logic that uses array.length
```
await sdk.api.abi.call({
          target: REGISTRY_ADDRESS,
          block: chainBlocks["astar"],
          chain: "astar",
          abi: ABI.get_coins,
          params: pool,
        })
```

Besides, the Audit is completed. Could you update the status?
URL is here.
https://github.com/muuu-finance/protocol/blob/main/docs/audit/PeckShield-Audit-Report-Muuu-v1.0.pdf

### DIFF
--- a/projects/muuu/tvl.js
+++ b/projects/muuu/tvl.js
@@ -33,6 +33,9 @@ const TOKENS = {
   // BAI -> DAI(TMP)
   "0x733ebcC6DF85f8266349DEFD0980f8Ced9B45f35":
     "0x6b175474e89094c44da98b954eedeac495271d0f",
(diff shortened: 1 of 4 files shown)


llamatester:
The adapter at projects/muuu exports TVL: 
        
 
 
    astar                     6.57 M
    
    total                    6.57 M

llamatester:
The adapter at projects/muuu exports TVL: 
        
 
 
    astar                     6.69 M
    
    total                    6.69 M

PR #2742, opened 2022-05-20
Add ALGO-USDC LP token as collateral

stefanofa:
Added ALGO-USDC LP Tinyman and ALGO-USDC LP Pact tokens, which can be used as collateral on Folks Finance.


llamatester:
The adapter at projects/folks-finance exports TVL: 
        
 
 
    algorand-borrowed         940.38 k
    algorand                  11.75 M
    borrowed                  940.38 k
    
    total                    11.75 M

PR #2787, opened 2022-05-26
Mars protocol shutdown

larry0x:
we have nuked our Terra contracts and returned all funds to users. therefore the TVL is zero for now


llamatester:
        
 
 
    terra-borrowed            0
    terra                     0
    borrowed                  0
    
    total                    0

0xngmi:
why is it currently reporting 2M?

larry0x:
> why is it currently reporting 2M?

shutting down process still ongoing rn

PR #2499, opened 2022-05-03
added i4i-finance

pureicis:
##### methodology (what is being counted as tvl, how is tvl being calculated): tvl is calculated using the total value of protocol's liquidity pool. Staked tokens include staked EYE values. Pool2 includes staked lp tokens eligible for EYE emissions

PR #2792, opened 2022-05-27
Revert "added i4i-finance"

waynebruce0x:
Reverts DefiLlama/DefiLlama-Adapters#2499


waynebruce0x:
reverting because I was gonna edit this but we've removed the caver dependency so I'm gonna do some more digging into best way to calc on chain

PR #2811, opened 2022-05-29
added support to calculate TVL for non-native assets locked in Oswap

pmiklos:
Currently, the live Oswap adapter version only counts TVL of the native asset (bytes) locked in, basically only one side of each liquidity pool. This pull request calculates TVL for all non-native assets as well that are locked in on Oswap, but only those not issued by the pool itself.  These non-native assets can either be outputs of other protocols or user issued custom assets. 

This pull request will basically double the calculated TVL of Oswap compared to what is live right now.

Also, updated methodology to reflect the new calculation method.

### DIFF
--- a/projects/helper/obyte.js
+++ b/projects/helper/obyte.js
@@ -1,29 +1,118 @@
 const utils = require('./utils')
 
 /**
- * @param {number} timestamp - unix timestamp in seconds from epoch of the moment in time for which the TVL is calculated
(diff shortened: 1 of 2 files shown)


llamatester:
The adapter at projects/oswap exports TVL: 
        
 
 
    obyte                     236.11 k
    
    total                    236.11 k

PR #2812, opened 2022-05-30
Update EverRise

beneverrise:
* Update EverRise pool2 owners to DAO contracts
* Include EverMigrate service TVL

##### Current TVL:
```
------ TVL ------
bsc-staking               14.47 M
fantom-staking            1.29 M
ethereum-staking          2.94 M
polygon-staking           3.29 M
avax-staking              1.40 M
bsc-pool2                 899.27 k
ethereum-pool2            851.71 k
avax                      70.49 k
avax-pool2                633.72 k
bsc                       1.51 M
ethereum                  583.47 k
fantom                    43.63 k
fantom-pool2              585.93 k
polygon                   56.40 k
polygon-pool2             731.12 k
staking                   23.38 M
pool2                     3.70 M

total                    2.26 M
```

Tokens migrating or merging contracts (e.g. v1 -> v2) use EverMigrate to provide escrowed migration where the EverMigrate contract holds the new tokens and holders exchange their old tokens for the new tokens at given consolidation rate. The EverMigrate TVL being counted is the new tokens still held under escrow and yet to be migrated to.


llamatester:
The adapter at projects/everrise exports TVL: 
        
 
 
    avax-staking              1.40 M
    polygon-staking           3.29 M
    bsc-staking               14.47 M
    ethereum-staking          2.94 M
    bsc-pool2                 899.27 k
    avax                      70.49 k
    fantom-staking            1.29 M
    avax-pool2                633.72 k
    ethereum                  583.48 k
    ethereum-pool2            851.72 k
    bsc                       1.51 M
    fantom                    43.63 k
    fantom-pool2              585.93 k
    polygon-pool2             731.12 k
    polygon                   56.40 k
    staking                   23.38 M
    pool2                     3.70 M
    
    total                    2.26 M

waynebruce0x:
So the tokens locked in everMigrate contracts haven't yet been in circulation? is that right?

beneverrise:
> So the tokens locked in everMigrate contracts haven't yet been in circulation? is that right?

If the migrate is open (has tokens) then their previous token they are migrating from will be out of circulation. Usually trading stopped and LP moved to new contract; so they have been sorta in circulation but via previous contract.

However the new tokens haven't been released as its done via a single txn exchange; old tokens come in (and given to the previous contract's deployer) and new tokens given to the holder doing trading them in.

The principle is to avoid airdrops and have a single verifiable onchain transaction for the upgrade. The previous tokens will have no value at the point the new tokens have value (start showing up in the TVL). However its up to the holders when they do the upgrade (and pay the gas for it)

beneverrise:
> So the tokens locked in everMigrate contracts haven't yet been in circulation? is that right?

If the migrate is open (has tokens) then their previous token they are migrating from will be out of circulation. Usually trading stopped and LP moved to new contract; so they have been sorta in circulation but via previous contract.

However the new tokens haven't been released as its done via a single txn exchange; old tokens come in (and given to the previous contract's deployer) and new tokens given to the holder doing trading them in.

The principle is to avoid airdrops and have a single verifiable onchain transaction for the upgrade. The previous tokens will have no value at the point the new tokens have value (start showing up in the TVL). However its up to the holders when they do the upgrade (and pay the gas for it).

If trading on the old tokens can't be stopped; then the migrate will normally be closed* prior to adding value to the new tokens; and they won't show up with value in the TVL.

*This is due to for example if the old token had liquidity locked; then the old tokens given to deployer in the exchange need to be sold to extract the value from the LP, to move it to the new token, and can't have people picking up the deployer sold tokens and rock bottom price and using them to migrate a second time.

beneverrise:
Example transaction https://etherscan.io/tx/0x302d63f29c8976a25b3360c2c2e86ccebc5db562de9f7c4d82ee8046363ac89f

beneverrise:
> So the tokens locked in everMigrate contracts haven't yet been in circulation? is that right?

Or set of migration txns https://etherscan.io/token/0x6107f4901476c3449403a1d908e19148950c2bc5?a=0x429ca183c5f4b43f09d70580c5365a6d21cccd47

The equvilent previous tokens that are being exchanged needed to be in circulation to be available in the migrate process (their is no point in the dev adding the new tokens they already hold or the LP tokens for example, only the tokens that can/should be migrated will be added to the migrate contract)

beneverrise:
The TVL for any token will drop over time (very fast at start) as the long tail of holders migrate (above set has been open for 103 days and still are happening on and off); but it means the deployer of new contract doesn't have to deal with all the book keeping and admin of snapshotting holder lists and accuracy required for manual airdrops (or pay the gas for every txn); and the holder doesn't need to worry about same name fake tokens being aridropped to them; or whether they will receive an airdrop after handing in their tokens as they have a txn record or what the correct token is.

beneverrise:
Website info for the EverMigrate dApp https://www.everrise.com/evermigrate/

image

waynebruce0x:
I dont think we should count everMigrate TVL because it sounds like the locked tokens are all unissued.
How do LPs get into the EverOwn DAO? Are they deposited by users?

beneverrise:
> I dont think we should count everMigrate TVL because it sounds like the locked tokens are all unissued.

They are effectively issued tokens?

The 3rd party protocol that uses the service has issued the tokens and locked their value in the EverRise protocol. Those tokens are no longer under the control of that 3rd party protocol but locked in the EverRise protocol and available to their holders at no cost other than exchanging their previously issued, now otherwise worthless, tokens. Their holders have the previous circulating tokens to redeem that value.

Otherwise, is value lost from the system while locked? If their previous issued tokens have no value, and the escrowed tokens have no value, then a large amount of value disappears, only to reappear on the redemption exchange.

The service for 3rd party protocols is operating as a stable coin dex LP with an N:1 OLD/NEW pair? (though it goes only one way)

I assume all the tokens in dex LPs count as issued tokens for TVL even though they haven't been bought (as the number can never go to zero as the price would be infinite).

> How do LPs get into the EverOwn DAO? Are they deposited by users?

Yes, the 3rd party protocol deposits them; and then held under vote lock by the holders of that protocol's governance token. EverOwn is a DAO service for other protocols (although EverRise also uses it for their LPs in this change). To release the LP, more than 51% of holdings that vote need to agree to the release (to enable LP removal); it's a governance-based lock rather than a timelock etc.

EverMigrate and EverOwn are part of EverRise's b2b/c security services (or protocol to protocol/holder), where a protocol locks value into the EverRise ecosystem as a neutral 3rd party escrow agent on assets, the protocol has given over to holder governance/ownership.

waynebruce0x:
Have the tokens in the LPs previously been in circulation?

waynebruce0x:
Cool, in that case we should include the LP TVL but not the evermigrate TVL

beneverrise:
> Cool, in that case we should include the LP TVL but not the evermigrate TVL

Change for LP only https://github.com/DefiLlama/DefiLlama-Adapters/pull/2863

beneverrise:
> Cool, in that case we should include the LP TVL but not the evermigrate TVL

Will follow up in discord on evermigrate and other services, might be easier :)

PR #2829, opened 2022-05-30
Include staked BAI in AstridDAO TVL

KayAstrid:
`bai-stablecoin` has become available on CoinGecko. Correct the price listing of BAI (which was temporarily set to DAI as a hack), and include staked BAI into AstridDAO's TVL.


llamatester:
The adapter at projects/astriddao exports TVL: 
        
 
 
    astar                     9.49 M
    
    total                    9.49 M

PR #2861, opened 2022-06-01
Support new version of mean finance

0xsambugs:
Hey Llamas!

About a month ago, we had a vulnerability report in [Mean Finance](https://defillama.com/protocol/mean-finance). Because of that, we had to redeploy our contracts (with the fix of course 😆 ). Right now, the adapter is only reporting the funds of the vulnerable version.

While the users migrate their positions, we thought it made sense to count the funds in both versions. Once all users have migrated, we can remove the old version. 

Also, we added support for Arbitrum, which has no funds yet (we haven't announced it). But if we add it now, we can avoid a future pull request

That's it from us, keep up the amazing work!


llamatester:
The adapter at projects/meanfinance exports TVL: 
        
 
 
    arbitrum                  0
    ethereum                  32.60 k
    optimism                  109.76 k
    polygon                   24.04 k
    
    total                    166.41 k

PR #2879, opened 2022-06-03
Add optimism to beethoven-x

franzns:
This is addition to the already present beethoven-x project. Do you need below info nevertheless?

Also switching to thegraph subgraphs since we plan to shut down our local one.

### DIFF
--- a/projects/beethovenx/index.js
+++ b/projects/beethovenx/index.js
@@ -1,32 +1,47 @@
-const { request, gql } = require("graphql-request");
-const {toUSDTBalances} = require("../helper/balances");
+const { GraphQLClient, gql } = require("graphql-request");
+const { toUSDTBalances } = require("../helper/balances");
+const { getBlock } = require("../helper/getBlock");
+async function getTVL(subgraphName, block) {
+  // delayed by 50 blocks to allow subgraph to update
+  block -= 50;
+  var endpoint = `https://api.thegraph.com/subgraphs/name/beethovenxfi/${subgraphName}`;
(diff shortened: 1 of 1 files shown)


llamatester:
Error while running adapter at projects/beethovenx: 
        
 
  
    
    
    Failed to decode `block.number` value: `subgraph Qmdky7YefnbBCWXq4SimxiiZJThGBitbA4FRBn3exrdhtd has only indexed up to block number 39697207 and data for block number 39697414 is therefore not yet available`
    Truncated error stack:
        at /home/runner/work/DefiLlama-Adapters/DefiLlama-Adapters/node_modules/graphql-request/dist/index.js:340:31

llamatester:
        
 
 
    fantom                    83.93 M
    optimism                  4.03 M
    
    total                    87.96 M

waynebruce0x:
Is there any way we can collect this data using onchain calls instead? We get a lot of trouble from thegraph

franzns:
Unfortunately we can't, since we can't source token prices from coingecko for the interest bearing tokens in our boosted pools.
We could use our backend but there would not be a way to get historical TVL. And that will take some time to implement as well.

waynebruce0x:
Can we map the ib token addresses to their underlying tokens and use that? Can the conversion rate be foun with a function call?

franzns:
What I could do is what we had before, that if we want TVL for the a current timestamp we can use our backend and if we need historical data, we use the subgraph. Could we merge this now and come up with another solution after? Any other solution will need quite some effort to implement.

PR #2894, opened 2022-06-06
update crema api

Pisces0818:


llamatester:
The adapter at projects/crema.js exports TVL: 
        
 
 
    
    total                    13.86 M

waynebruce0x:
Do you know where the additional $3M has come from?

Pisces0818:
The current adapter is using our old api, which doesn't include a few new pools. New api has been updated.

> Do you know where the additional $3M has come from?

The current adapter is using our old api, which doesn't include a few new pools. New api has been updated.

PR #2912, opened 2022-06-07
feat: update ankr tvls to use official endpoint

ittikorns:


waynebruce0x:
We'd prefer to keep the adapter using onchain calls

ittikorns:
@waynebruce0x There are inconsistencies in the data we fetched from on chains due to the architecture of our token is certificate type which can be claimed and unclaimed and spread out on multiple chains. Would it be possible to use this method to sort out the on-chain data while we're working out? You can verify the TVLs here https://www.ankr.com/staking/stake/

waynebruce0x:
Since aXXXb tokens are mint/burnt 1:1 with deposits, can't we count this instead? Also, since MATIC staking all takes place on ethereum, on other projects we have counted it as ethereum TVL.

llamatester:
The adapter at projects/ankr exports TVL: 
        
 
 
    bsc                       7.20 M
    ethereum                  69.21 M
    fantom                    35.99 k
    avalanche                 204.43 k
    
    total                    76.65 M

ittikorns:
@waynebruce0x There are complications on aETHb total supply which doesn't match with the total number of ETH staked on eth2. I've removed polygon tvl and merged it with Ethereum.

PR #2919, opened 2022-06-08
Add optiondance tvl

cyrto:
OptionDance is a decentralized platform for users to trade options in simplified strategies.


cyrto:
> hi @cyrto
> 
> Thanks for the PR, can you update the adapter to fetch the info from on chain calls instead?

Hi, The Mixin Network is not support contract so far, so there is no on chain calls available.

PR #2935, opened 2022-06-10
change back to backend graph, thegraph as fallback

franzns:
Change back the logic to what was done by @0xngmi here https://github.com/DefiLlama/DefiLlama-Adapters/blob/0d6bcfb2d66d9adb50c20ce6c70a5c257a69a6f6/projects/beethovenx/index.js

We are using our backend to track current number of TVL, which takes all nested assets into account. If the timestamp is older, we use thegraph API as fallback.


llamatester:
        
 
 
    fantom                    97.34 M
    optimism                  4.86 M
    
    total                    102.21 M

0xngmi:
main issue i see with it is that the subgraph could be arbitrarily delayed
eg; we could be ingesting data that is days old with your pr

waynebruce0x:
By passing a blockheight into subgraph queries we can easily see when data is outdated, and debug the issue, but with this implementation we wont be able to debug anything

franzns:
> By passing a blockheight into subgraph queries we can easily see when data is outdated, and debug the issue, but with this implementation we wont be able to debug anything

Would you like to see if data is outdated on thegraph? An option would be to query the meta information that returns the current block it has synced to. Data on the backend is considered current and up to date because it uses a combination of subgraph and onchain calls.

waynebruce0x:
Yeah I think that'd work, thanks

llamatester:
        
 
 
    fantom                    77.51 M
    optimism                  2.70 M
    
    total                    80.20 M

llamatester:
        
 
 
    optimism                  2.70 M
    fantom                    77.64 M
    
    total                    80.34 M

franzns:
@waynebruce0x as discussed, added the getLatestSyncedBlock function for both chains.

franzns:
@waynebruce0x anything I can do to speed this up?

PR #2981, opened 2022-06-15
fix astarfarm tvl

lilpacy:
## Summary
Fixing higher than the actual tvl value

## Details
The calculation was incorrect and displayed higher than the actual tvl, so it was corrected to display the correct value.  
The tvl has been lowered to the correct value due to this.

### DIFF
--- a/projects/astarfarm/abi.json
+++ b/projects/astarfarm/abi.json
@@ -1,13 +1,15 @@
-[{
-  "inputs": [],
-  "name": "staked",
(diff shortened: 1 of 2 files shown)


llamatester:
The adapter at projects/astarfarm exports TVL: 
        
 
 
    astar                     7.39 k
    
    total                    7.39 k

lilpacy:
I really apologize for bothering you all so many times🙏

PR #3045, opened 2022-06-21
Premia - Use base asset from subgraph query instead of DAI

Lumyo:
Fantom PUT pools are using USDC instead of DAI, and their TVL are therefore currently not accounted for, as DAI is hardcoded as base asset for all pools.


llamatester:
The adapter at projects/premia exports TVL: 
        
 
 
    fantom                    839.20 k
    ethereum-staking          4.52 M
    arbitrum                  2.77 M
    bsc                       8.86 k
    ethereum                  2.32 M
    staking                   4.52 M
    
    total                    5.94 M

PR #3079, opened 2022-06-24
Added Tenset TVL

pkostrzewski:


waynebruce0x:
Can we use onchain calls instead?

PR #3135, opened 2022-07-01
MuesliSwap: Hybrid DEX upgrade - Use combined pool + order book 

MuesliSwapTeam:
Twitter: https://twitter.com/muesliswapteam

This uses a different URL to query the MueslISwap order book. This order book is enriched with pseudo orders generated from the hybrid AMM (based on Liquidity Pools) that MuesliSwap features. The combined liquidity in the pseudo orders is equal to the TVL of the liquidity pool that generated it, so it should be fair to simply count this as by the previous approach.

We can also think of using two URLs, one for the raw order book and one for the liquidity pools and then independently sum the results and are open for feedback.

### DIFF
--- a/projects/muesliswap/index.js
+++ b/projects/muesliswap/index.js
@@ -44,7 +44,7 @@ async function adaTvl(){
         const totalSell = totalAmountOtherToken * topPrice
         totalAda += totalBuy + totalSell
     }))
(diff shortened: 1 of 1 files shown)


llamatester:
        
 
 
    milkomeda                 2.07 M
    smartbch-staking          1.93 k
    cardano-staking           1.88 M
    cardano                   4.86 M
    smartbch                  11.16 k
    staking                   1.88 M
    
    total                    6.94 M

PR #3160, opened 2022-07-04
Update angle adapter

tutacrypto:
### Angle adapter update

This is an update of Angle adapter to take into account [Angle Borrowing module](https://docs.angle.money/angle-borrowing-module/borrowing-module) on Ethereum mainnet and on Polygon. 
Some Polygon TVL is not displayed when testing, but this might come from very low balances (< $1k). 

Current DeFillama page: https://defillama.com/protocol/angle

### DIFF
--- a/projects/angle/index.js
+++ b/projects/angle/index.js
@@ -41,12 +41,26 @@ const poolManagers_abi = {
 };
 
 // get Borrowing module vault managers list
-async function getVaultManagersFromAPI() {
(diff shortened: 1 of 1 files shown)


llamatester:
        
 
 
    ethereum-staking          2.12 M
    ethereum                  99.07 M
    polygon                   173.079
    staking                   2.12 M
    
    total                    99.07 M

PR #3318, opened 2022-07-21
Fixed onAVAX TVL

rorrieswap:
Fixed onAVAX TVL Calculations

**ONLY UPDATE TVL Code**
Should be around 745k


llamatester:
The adapter at projects/onavax exports TVL: 
        
 
 
    avax                      742.24 k
    
    total                    742.24 k

PR #3326, opened 2022-07-22
Add liquidation adapter for Compound

0xStrobe:
🤫

### DIFF
--- a/liquidations/aave-v2/index.ts
+++ b/liquidations/aave-v2/index.ts
@@ -21,6 +21,11 @@ const query = gql`
         currentTotalDebt
       }
     }
+    _meta {
+      block {
+        number
+      }
+    }
   }
 `;
 
--- a/liquidations/aave-v2/index.ts
+++ b/liquidations/aave-v2/index.ts
@@ -1,99 +1,102 @@
-import { gql, request } from "graphql-request"
-import { getPagedGql } from "../utils/gql"
+import { gql, request } from "graphql-request";
+import { getPagedGql } from "../utils/gql";
 
 const query = gql`
-query users($lastId: String) {
-    users(first: 1000, where:{
-        borrowedReservesCount_gt: 0,
-        id_gt: $lastId
-      }) {
-        id
-        reserves{
-          usageAsCollateralEnabledOnUser
-          reserve{
-            symbol
-            usageAsCollateralEnabled
-            underlyingAsset
-            price {
-                priceInEth
-            }
-            decimals
-            reserveLiquidationThreshold
+  query users($lastId: String) {
+    users(first: 1000, where: { borrowedReservesCount_gt: 0, id_gt: $lastId }) {
+      id
+      reserves {
+        usageAsCollateralEnabledOnUser
+        reserve {
+          symbol
+          usageAsCollateralEnabled
+          underlyingAsset
+          price {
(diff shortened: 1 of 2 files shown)


liquidations/aave-v2/index.ts
```diff
@@ -1,99 +1,102 @@
...
-          currentATokenBalance
-          currentTotalDebt
+          decimals
+          reserveLiquidationThreshold
         }
+        currentATokenBalance
+        currentTotalDebt
+      }
     }
-    _meta{
```
0xngmi, line 27: better not to remove this, need to use it to make sure that subgraph is not stale

liquidations/aave-v2/index.ts
```diff
@@ -1,99 +1,102 @@
...
     }
-}
-`
+  }
+`;
 
-const subgraphUrl = "https://api.thegraph.com/subgraphs/name/aave/protocol-v2"
+const subgraphUrl = "https://api.thegraph.com/subgraphs/name/aave/protocol-v2";
+
+const positions = async () => {
```
0xngmi, line 37: has anything been changed here? if no change has been made better not to change the code since its a pain to review

liquidations/compound-v2/index.ts
```diff
@@ -0,0 +1,159 @@
...
+          exchangeRate
+          reserveFactor
+          underlyingDecimals
+          underlyingAddress
+        }
+        borrowBalanceUnderlying
+        supplyBalanceUnderlying
+        enteredMarket
+      }
+    }
```
0xngmi, line 35: need block check

liquidations/compound-v2/index.ts
```diff
@@ -0,0 +1,159 @@
...
+        return {
+          debt,
+          price,
+          token: token.market.underlyingAddress,
+          totalBal: token.supplyBalanceUnderlying,
+          decimals,
+        };
+      });
+
+    const liquidablePositions = debts
```
0xngmi, line 124: i see, i guess pretty similar to aave adapter?

liquidations/aave-v2/index.ts
```diff
@@ -1,99 +1,107 @@
...
         }
+        currentATokenBalance
+        currentTotalDebt
+      }
     }
-    _meta{
-        block {
-            number
-        }
+    _meta {
```
0xStrobe, line 24: added it back

liquidations/aave-v2/index.ts
0xngmi: its not, i added the meta bit but never got around to adding the check against current block (bug in adapter)

the way its handled on defillama adapters its by querying subgraph at a specific block, if subgraph is not synced then query just fails

i wanted to check against block here instead because liquidation data is much more time sensitive

liquidations/aave-v2/index.ts
0xngmi: yeah np, if reviewing i prefer to make a PR with changes, and then after its merged make another PR running prettier, which can be quickly merged

but tbh it doesnt really matter

liquidations/aave-v2/index.ts
0xStrobe: got it, will implement in the future in a diff pr

PR #3403, opened 2022-07-30
update spicyswap API

arrijabba:
switch from backup API (B) to main API (A)


llamatester:
Error while running adapter at projects/spicyswap: 
        
 
  
    
    
    Error: 
    
            module.exports = {
              methodlogy: '...',
              ethereum: {
                tvl: 
                pool2:
              }
            }

PR #3489, opened 2022-08-09
add more pools from Mercurial Finance.

siong1987:
Added new pools from Mercurial Finance and deprecated some pools (UST).

https://github.com/mercurial-finance/stable-swap-n-pool-js


llamatester:
Error while running adapter at projects/mercurial.js: 
        
 
  
    
    
    Error: 
    
            module.exports = {
              methodlogy: '...',
              ethereum: {
                tvl: 
                pool2:
              }
            }

PR #3507, opened 2022-08-10
Switching to v2

franzns:
We introduced our new frontend which is also supported by a fully revamped backend. 
Changing the adapter to use the new backend.


llamatester:
        
 
 
    optimism                  3.09 M
    fantom                    72.67 M
    
    total                    75.76 M

PR #3678, opened 2022-08-24
feat: add hodl

yiplee:

hodl logo


yiplee:
@g1nt0ki hello, please update the category of HODL to be Staking thanks.

PR #3846, opened 2022-09-09
added suport for the new version of Prophet

pmiklos:
A new version of Prophet Prediction Markets was rolled out this week, to support the new prediction markets based on the new version I added support to pull TVL from this new version.


llamatester:
The adapter at projects/prophet exports TVL: 
        
 
 
    obyte                     1.09 k
    
    total                    1.09 k

PR #3863, opened 2022-09-12
add gmx volume adapter

xyse:
0xngmi told me you will add this once the UI for perp dexes is added so I wrote it


0xtawa:
Hi, could you return only the volume of the swaps? Currently not including other types of volume in the dashboard

xyse:
you can easily change what volume is included by removing the changing the gql query. i changed it it now only includes swaps.

PR #3888, opened 2022-09-14
Minto update

semdestroyer:
Hi! In our project tvl and staking have the same values, we need to take the value of staking  as tvl


realdealshaman:
sorry, we can't count it this way, it would not be inline with the methodology applied to all other protocols, including others with similar stats. Also this would cause your tvl to 2X whenever users toggled on the "staking" filter

semdestroyer:
So, we can't count only staking values as tvl without staking filter? Just zero main tvl is not  what we need.

realdealshaman:
not every user sees staking tvl as tvl, we want to be inclusive of all projects so we added the filters so that we could also list projects that only have a staking tvl

realdealshaman:
The value that is staked will always be visible when the staking toggle is on so no need to count it twice. Can we close this PR?

PR #3989, opened 2022-09-23
Kdswap adapter

KdWarlordz:
Solution for breaking price information Kdswap. 

explanation: if the balance check of 1 token cannot be retrieved does not mean, all should fail it should just fallback to 0


llamatester:
The adapter at projects/kdswap exports TVL: 
        
 
 
    kadena                    1.27 M
    
    total                    1.27 M

llamatester:
The adapter at projects/kdswap exports TVL: 
        
 
 
    kadena                    1.26 M
    
    total                    1.26 M

PR #3998, opened 2022-09-23
Feature/tvl fix

chaoticgoodpanda:
Fixed tvl calculations to accurately reflect totalSupply. Older work used underlyings for totalSupply calculations which are inaccurate. Using totalSupply for totalSupply calculations is accurate.


llamatester:
The adapter at projects/0vix exports TVL: 
        
 
 
    polygon-borrowed          1.76 M
    polygon                   6.12 M
    borrowed                  1.76 M
    
    total                    6.12 M

llamatester:
The adapter at projects/0vix exports TVL: 
        
 
 
    polygon-borrowed          3.06 M
    polygon                   5.63 M
    borrowed                  3.06 M
    
    total                    5.63 M

g1nt0ki:
I have refactored the code to use compound adapter, noticed that you were including borrowed as part of tvl, sorry but borrowed value should not be included in lending protocol tvl, I have manually verified that all tokens in lending contracts are included correctly here.

chaoticgoodpanda:
Hmm, okay, but can you revert back to my commit _before_ changing TVL = totalSupply + totalBorrows?

This is because the compound calculation is incorrect. You can check it on app.0vix.com right now: 
Screen Shot 2022-09-26 at 4 48 44 PM


TVL just totalSupply alone is $8.4 MM, but defi-llama's calculations have it at $5.4 MM, which is incorrect: https://defillama.com/protocol/0vix.

g1nt0ki:
@chaoticgoodpanda TVL is the token value in your contracts, hence we are not counting the borrowed tokens. 

When you count the token value in all your contracts, is it 5.6M or 8.4M ?

PR #4004, opened 2022-09-23
Add premium and notional

paulvaden:
Instead of returning totalVolume/dailyVolume, the Lyra volume adapter now returns:

totalNotionalVolume
totalPremiumVolume
dailyNotionalVolume
dailyPremiumVolume

For Options dexes it makes sense to show both premiums and notional IMO.

### DIFF
--- a/volumes/helper/getLyraSubgraphVolume.ts
+++ b/volumes/helper/getLyraSubgraphVolume.ts
@@ -1,10 +1,10 @@
 import { Chain } from "@defillama/sdk/build/general";
 import { BigNumber } from "ethers";
 import { request, gql } from "graphql-request";
-import { getBlock } from "../../projects/helper/getBlock"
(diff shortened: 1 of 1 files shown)


dtmkeng:
@0xtawa

0xtawa:
Hi @paulvaden, please, keep also `totalVolume` and `dailyVolume` attributes, will merge then. ty

paulvaden:
@0xtawa is it possible to keep them as two separate volume metrics for Options DEXes? Both premium volume and notional volume are important to see separately.

0xtawa:
Hi @paulvaden, indeed. I meant to also add the attributes `totalVolume` and `dailyVolume` with the others, prefer to do that to keep same structure until derivatives attributes are defined. Just added them, merging

ty

PR #4011, opened 2022-09-24
Fix DUAL TVL

brittcyr:
Update the dual.finance tracker now that it has switched to using soETH


llamatester:
        
 
 
    solana                    1.27 k
    
    total                    1.27 k

PR #4021, opened 2022-09-26
Track Klima DAO wallet

kargakis:
Track the Klima DAO wallet which is another place
apart from the treasury where DAO funds are kept.

https://docs.klimadao.finance/developers/contracts/dao-wallet


llamatester:
The adapter at projects/klima-dao exports TVL: 
        
 
 
    polygon-staking           20.00 M
    polygon                   44.97 M
    staking                   20.00 M
    
    total                    44.97 M

PR #4157, opened 2022-10-05
Use new muesliswap price api for determining the price of AADA

MuesliSwapTeam:
MuesliSwap is deprecating the old orderbook contract, so the price of AADA on the old contract is not representative anymore. Instead, the new price API should be used.


llamatester:
The adapter at projects/aada exports TVL: 
        
 
 
    cardano-staking           419.39 k
    cardano                   340.24 k
    staking                   419.39 k
    
    total                    340.24 k

PR #4173, opened 2022-10-05
Update DUAL tvl

brittcyr:
Update tvl calculation for dual finance to include all value locked, including withdrawable tokens in expired vaults


llamatester:
        
 
 
    solana                    3.04 k
    
    total                    3.04 k

PR #4175, opened 2022-10-05
Update Phuture TVL

jn-lp:
- We've added a new type of product – SavingsVault on Ethereum, it has separate tvl function, which result is added to total eth tvl using `sdk.util.sumChainTvls`
- Fixed Avax logs subscription – now all functions use multicalls instead


llamatester:
The adapter at projects/phuture exports TVL: 
        
 
 
    avax                      285.46 k
    ethereum                  104.12 k
    
    total                    389.58 k

PR #4450, opened 2022-10-25
feat: adding eth call selling quote vault

smiti-maheshwari:
Adding vault in polynomial protocol value


llamatester:
The adapter at projects/polynomial exports TVL: 
        
 
 
    optimism                  1.07 M
    
    total                    1.07 M

PR #4452, opened 2022-10-25
Cega Defi Protocol: include collateral in tvls

vsai:
This commit will include the total the sum of collateral being held in the tvl.


llamatester:
The adapter at projects/cega exports TVL: 
        
 
 
    solana-borrowed           44.95 M
    solana                    10.07 M
    borrowed                  44.95 M
    
    total                    10.07 M

PR #4461, opened 2022-10-26
Update for DFX v2 pools.

pscarolus:
DFX Recently launched V2 of the AMM. This means liquidity has largely been migrated to new pool contracts.
Some of the stablecoin tokens on Polygon are also now native rather than bridged tokens for example CADC, XSGD and XIDR.


llamatester:
The adapter at projects/dfx exports TVL: 
        
 
 
    polygon                   4.07 M
    ethereum                  18.03 M
    
    total                    22.10 M

PR #4625, opened 2022-11-12
use new solend config endpoint

0xCactus:
Use new solend endpoint to get markets

Solend


llamatester:
The adapter at projects/solend exports TVL: 
        
 
 
    solana-borrowed           26.23 M
    solana                    36.84 M
    borrowed                  26.23 M
    
    total                    36.84 M

PR #4762, opened 2022-11-28
Youves: correction of pool2

spolel:
Correction of pool2 as the contracts involved do not represent value locked as described in "pool2" in https://docs.llama.fi/list-your-project/what-to-include-as-tvl. None of those contracts contain staked LP tokens from our own governance token ($YOU).

### DIFF
--- a/projects/youves/index.js
+++ b/projects/youves/index.js
@@ -83,11 +83,6 @@ async function tvl() {
     fetchBalance(balances, 'tzbtc-lp', engines.uDefitzBTCLPV2, 0, sharePrice),
   ])
 
-  return convertBalances(balances)
-}
-
-async function pool2() {
-  const balances = {}
   const fixBalances = await getFixBalances('tezos')
 
   const youvesLPs = dexes.map(i => i.contractAddress).filter(i => i)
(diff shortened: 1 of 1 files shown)


g1nt0ki:
@spolel strictly speaking, it is not a pool2, but this is not a normal dex either. Limited choice of tokens, and they are paired against youvesTokens whose backing is already counted towards tvl.

There is two solutions here, one create a separate listing for youves dex.
Other is count non youves mintedTokens in the pools towards tvl, what do you prefer?

g1nt0ki:
*the dex will be marked as double-counted and its tvl wont reflect on overall youves/tezos tvl.

youves would look similar to https://defillama.com/protocol/folks-finance

mlaeng:
> Other is count non youves mintedTokens in the pools towards tvl, what do you prefer?

You mean "count only the non youves Tokens in the pools towrds tvl?" (but not the youves minted tokens?)

mlaeng:
> youves would look similar to https://defillama.com/protocol/folks-finance

You mean having a tab for the youves dex and one for the lending part then being summed up on the left?

image

g1nt0ki:
that is a bug in the UI, if you look at algorand chain page: 
image
the liquid staking part is greyed out because it is marked as doublecounted: https://github.com/DefiLlama/DefiLlama-Adapters/blob/main/projects/folks-galgo/index.js#L9

In their case, we noticed that most of the lending tvl comes from their liquid staking protocol

g1nt0ki:
> > Other is count non youves mintedTokens in the pools towards tvl, what do you prefer?
> 
> Hey, thanks a lot for the super speedy feedback! You mean "count only the non youves Tokens in the pools towards tvl?" (but not the youves minted tokens?)

yeah, ideally only the tokens not belonging to the protocol should be counted towards tvl, so no uUSD/uDEFI

g1nt0ki:
closing this due to lack of activity

PR #4783, opened 2022-11-30
lympo: fix tvl

RomanG11:
Our ultimate goal is to reflect TVL data on CoinMarketCap (CMC)

But currently, SPORT token TVL shows under staking "toggle," and CMC is not pulling this data. How should we manage the adapter, so it pulls TVL from Polygon and not from Staking toggle.


llamatester:
The adapter at projects/lympo exports TVL: 
        
 
 
    polygon                   568.32 k
    
    total                    568.32 k

RomanG11:
https://api.llama.fi/tvl/lympo

services like CMC and coingecko are using this endpoint to show TVL. Right now it returns 0. How can we fix it?

PR #4900, opened 2022-12-17
[Update] Fetch TVL data from contract calls

strxfinance:
This pull request updates the API to fetch data from contract calls to get the precise TVL rather than checking balance only. The previous method was inaccurate as it relied on balance data, which can be changed or not reflect the true value locked in the contract.

By using contract calls to fetch the TVL data, we can ensure that the data displayed in the API is accurate and reflects the true value locked in the contract.

Please review and let me know if there are any questions or concerns.

Changes Made:
- Updated the API to fetch TVL data from contract calls rather than balance
- Tested the updated API to ensure it is working correctly and returning accurate data

Thank you for considering this pull request.


llamatester:
The adapter at projects/strxfinance exports TVL: 
        
 
 
    tron                      12.56 k
    
    total                    12.56 k

PR #4996, opened 2023-01-06
Include GSO in TVL for DUAL finance

brittcyr:


llamatester:
        
 
 
    solana                    480.90 k
    
    total                    480.90 k

PR #5030, opened 2023-01-12
refactor(sense): sum tvl via underlying tokens rather than target tokens

jparklev:
Updating the sense adapter with the following changes:
* Add in the current Space Factory address
* Calculate TVL via the underlying tokens rather than the Target tokens

For an example of that last point, with this change we no longer sum up the `maUSDC` balances we hold directly, but rather we first them to their unwrapped USDC value, and then sum those balances. We can do this because our protocol must track the exchange rate between these two tokens anyway for our rate splitting mechanism to work (`scale`). The motivation here is that the underlying tokens will have price much more often then the potentially exotic Targets (the IDLE Junior STETH tranche, for example)


llamatester:
The adapter at projects/sense exports TVL: 
        
 
 
    ethereum                  802.61 k
    
    total                    802.61 k

jparklev:
Ty for merging @g1nt0ki – I see the dashboard just updated with the latest data! Is it possible to backfill like the last month of the historical data with these changes as well?

PR #5069, opened 2023-01-16
fix hallmark inverse

Define101:
fix hallmark inverse


llamatester:
The adapter at projects/inverse exports TVL: 
        
 
 
    ethereum                  6.80 M
    
    total                    6.80 M

PR #5086, opened 2023-01-18
Hard code Kujira USK contract addresses due to migrations changing code IDs.

MikeR808:
Previous implementation has lowered TVL due to a contract migration when configs update and contract addresses not being included.

This change adds the contract addresses for USK vaults directly in.


llamatester:
The adapter at projects/kujira exports TVL: 
        
 
 
    kujira                    2.72 M
    
    total                    2.72 M

llamatester:
The adapter at projects/kujira exports TVL: 
        
 
 
    kujira                    2.72 M
    
    total                    2.72 M

g1nt0ki (changes requested):
instead of removing the old code, can you add this list to owners, I think (and am hoping) that `sumTokens` method ignores duplicates

PR #5137, opened 2023-01-25
Pirex TVL Updates

swheel33:
Update Pirex TVL to use locked CVX in convex locker instead of using supply of pxCVX. This is a different number since the pxCVX can be burnt for various reasons (ie Expert mode) but the CVX is still locked


llamatester:
The adapter at projects/pirex exports TVL: 
        
 
 
    ethereum                  9.82 M
    
    total                    9.82 M

PR #5350, opened 2023-02-16
MuesliSwap: Add support for token/token pairs

MuesliSwapTeam:
MuesliSwap has recently launched stablecoin pools that pair two native tokens. This was currently not supported by the adapter but only a trivial change is necessary to support it.

Reference: https://medium.com/@muesliswap/iusd-djed-concentrated-lp-yield-farm-56e95b2f881c


llamatester:
        
 
 
    smartbch-staking          1.19 k
    smartbch                  7.68 k
    milkomeda                 576.84 k
    cardano-staking           1.79 M
    cardano                   8.78 M
    staking                   1.80 M
    
    total                    9.37 M

PR #5408, opened 2023-02-22
feat: update Reserve adapter to dynamically fetch TVL 

larrythecucumber321:
This implementation relies on listening for the [deployer](https://etherscan.io/address/0xfd6cc4f251eae6d02f9f7b41d1e80464d3d2f377) contract's `RTokenCreated` event to populate the `owners` array. The relevant tokens are then subsequently queried

closes https://github.com/reserve-protocol/protocol/issues/423


llamatester:
The adapter at projects/reserve exports TVL: 
        
 
 
    ethereum                  28.92 M
    
    total                    28.92 M

g1nt0ki:
@larrythecucumber321 thanks for the PR, I updated it to cache logs and blacklisted RSR from being counted towards tvl

larrythecucumber321:
Thank you for the upgrades @g1nt0ki 🙏 I was wondering if it would be possible to remove RSR from the blacklist, because users stake it to provide insurance for the protocol, just like how AAVE is used for that purpose (and I see it's being counted towards TVL).

g1nt0ki:
sorry, we are more strict/conservative when it comes to CDP, for example: in the daI-USDC arrakis LP of collateral, we only count the USDC part, same for curve LP with FRAX in it for FRAX

larrythecucumber321:
Thanks for that explanation. Would RSR staking be more appropriate under the `staking` category instead? I can create another PR for that!

g1nt0ki:
yup, that is correct

g1nt0ki:
hope this is correct: https://github.com/DefiLlama/DefiLlama-Adapters/commit/15520b21824f3731fa5ac575b0eee8f024aa843a

larrythecucumber321:
Beauty! Sniffing the creation event for `stRSR`... 🤌
image

PR #5500, opened 2023-03-04
Llama Airforce: Track new cvxCrv, auraBal & cvx vaults. Fix FXS vault TVL calculation.

benber86:
This fixes some issues with the current TVL calculation for Llama Airforce and tracks new vaults:
- The cvxCrv vault that is currently tracked has been deprecated and only holds a fraction of the TVL of the new one. The TVL of both vaults is now tracked.
- The calculation of the TVL for the FXS vault assumed the underlying was FXS when it is an LP token for the cvxFXS/FXS pool on Curve. This PR takes this into consideration when calculating the TVL
- The CVX and auraBal vaults are currently not tracked, this PR would add both of them

### DIFF
--- a/projects/llama-airforce/abi.json
+++ b/projects/llama-airforce/abi.json
@@ -1,4 +1,7 @@
 {
   "totalUnderlying": "uint256:totalUnderlying",
-  "priceOracle": "uint256:price_oracle"
+  "totalAssets": "uint256:totalAssets",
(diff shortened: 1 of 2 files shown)


llamatester:
        
 
 
    ethereum                  32.41 M
    
    total                    32.41 M

llamatester:
        
 
 
    ethereum                  32.10 M
    
    total                    32.10 M

PR #5501, opened 2023-03-04
AnyHedge - fix data timing

A60AB5450353F40E:
Discussed on Discord:

>I'd rather your app display the data with correct dates and just not show the last 30d: here's how I plot the stats, you'll notice the gap on the volume opened&closed chart: https://gitlab.com/0353F40E/anyhedge-stats/-/blob/master/plots/plots.md

>What if I make the adapter throw an error when querying date range of last 30d, will it still be trying to backfill where it left off?

>anyway, would it still work if the adapter throws an error for queries within last 30d range?

Got agreement from 0xngmi & ulysses

If you're wondering why we can't show current data - it's because it will always be 0, since opened-not-yet-settled contracts are hidden, you can't detect those on-chain. We can't see active contracts into which settled contracts rolled over.

Every contract has 2 transactions: 1 for when it's opened, 1 for when it's settled. The 1st TX will be mined on creation, but when you scrape on-chain data there's no way to tell it apart from other UTXOs since it is P2SH and it could be anything, maybe a HTLC, maybe a multisig, maybe a custom covenant, you have to wait for the 2nd TX (one that spends the UTXO) to be able to find the 1st TX.
So later, when contract matures, either party will submit the 2nd TX which reveals the code + adds the oracle message, and then the contract pays it out - at which point it can be detected and my scanner detects these settlements.

This is similar to LN on BTC chain - you can't really know that an UTXO is a LN channel, until it is actually closed.

### DIFF
--- a/projects/anyhedge/index.js
+++ b/projects/anyhedge/index.js
@@ -14,12 +14,20 @@ async function GetDailyHistory() {
 }
 
 async function getTVLAnyHedge(timestamp) {
-  const day = new Date((timestamp - 30*24*3600) * 1000).toISOString().slice(0,10) // tvl data lags by contract duration since contracts are secret until settled
+  const day = new Date(timestamp * 1000).toISOString().slice(0,10)
   return dayHistory[day]
 }
 
 async function tvl(timestamp) {
   let tvlAnyHedge
+
+  // tvl data lags by contract duration since contracts are secret until settled
+  // so tvl at current time will always be 0, and only later when contracts are revealed
+  // can it be calculated in retrospect and stats back-filled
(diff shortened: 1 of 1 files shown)


llamatester:
Error while running adapter at projects/anyhedge: 
        
 
  
    
    
    Data for the date is incomplete, awaiting contract reveals.

A60AB5450353F40E:
obviously, the `test.js` will now fail... when testing locally I just changed `test.js` to query older date:
```
-  const unixTimestamp = Math.round(Date.now() / 1000) - 60;
+  const unixTimestamp = Math.round(Date.now() / 1000) - 60 - 31*86400;
```

PR #5571, opened 2023-03-14
Update Mars contract query

larry0x:
The `{ markets: {} }` query by default [only returns the first 5 assets](https://github.com/mars-protocol/red-bank/blob/master/contracts/red-bank/src/query.rs#L23-L24). We now have 6 supported assets so one is missed.

This PR adds an additional parameter `limit: 10` so it returns the first 10.

The contract only supports a maximum of 10 per query. If we do have more than 10 assets in the future, we will need to implement a loop to fetch all assets, 10 at a time. I can do a PR later when this becomes necessary.

### DIFF
--- a/projects/mars/index.js
+++ b/projects/mars/index.js
@@ -7,7 +7,7 @@ const chain = 'osmosis'
 const contract = 'osmo1c3ljch9dfw5kf52nfwpxd2zmj2ese7agnx0p9tenkrryasrle5sqf3ftpg'
 
 async function borrowed() {
-  const res = await queryContract({ contract, chain: 'osmosis', data: { markets: {} } })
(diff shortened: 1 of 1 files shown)


llamatester:
        
 
 
    terra                     0
    osmosis-borrowed          4.74 M
    osmosis                   7.32 M
    borrowed                  4.74 M
    
    total                    7.32 M

larry0x:
> The adapter at projects/mars exports TVL:
> 
> ```
> terra                     0
> osmosis-borrowed          4.74 M
> osmosis                   7.32 M
> borrowed                  4.74 M
> 
> total                    7.32 M 
> ```

This result is correct. Currently on defillama dashboard it shows only $2.21M borrowed

PR #5590, opened 2023-03-15
Streamflow: use different api endpoint

Yolley:
Hey everyone, we've had some stability issues with the previous solution that provided the API results so we'd like to set up a different API source as you can see in the PR.

The API source is an internal Metabase storage, it polls on-chain data, token prices and expose that through an API endpoint provided.

Here is a run-through of the Metabase setup: https://www.loom.com/share/2f41c0da320b4be9809949c2da4337dd

If you'd like we can provide you with access and run through the details of how it works let us know.

Thank you in advance.

### DIFF
--- a/projects/streamflow/index.js
+++ b/projects/streamflow/index.js
@@ -1,27 +1,22 @@
 const { getCache } = require('../helper/http')
 
-const TVL_KEY = "TVL (unlocked) (L1+stables)";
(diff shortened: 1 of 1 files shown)


llamatester:
The adapter at projects/streamflow exports TVL: 
        
 
 
    solana-vesting            147.92 M
    solana                    164.50 k
    vesting                   147.92 M
    
    total                    164.50 k

PR #5604, opened 2023-03-16
Add Polygon for ReHold

bifot:
We already have a listing on DefiLlama, just want to add a new blockchain for us. Same contract address, same way for calculating TVL. Thank you!


llamatester:
The adapter at projects/rehold exports TVL: 
        
 
 
    polygon                   32.88 k
    bsc                       168.87 k
    
    total                    201.76 k

PR #5640, opened 2023-03-19
Change links to "llama.fi" domain

Avedith:
As what the title says.


Avedith:
Don't merge as discord subdomain doesn't exist. No idea if this will exist eventually or not.

PR #5651, opened 2023-03-20
updates grizzly fi TVLS

rogsta-dev:
This PR updates the TVL of the Grizzly.fi project. It contains all new pools on the BSC version to be summed to the TVL. These pools are aggregator contracts, which provide liquidity in different protocols and handle the LP Tokens.

The following pools were added:

- 3 Biswap hives
- 2 GHNY Farm contracts
- 7 Thena Hives (yearn vault forks)
- 3 Pancakeswap Stable Hives

The Biswap, The Farms, and the Thena hives are aggregator contracts containing UniswapV2-like LP tokens so they were unwrapped using unwrapUniswapLPs helper function. The Pancakeswap Stable Hives were unwrapped using a custom function unwrapStablePcsLPs. As these Pancakeswap Stable Hives are based on curve stableSwap, this function extracts the underlying coins and their balances, the total supply, and calculates the underlying token prices.

In addition, a bug from the previous version was fixed. The Pancakeswap hives and the Biswap Hives contain 3 different strategies (standard, stable, grizzly). The contracts track the LP values separately so 3 calls per hive need to be executed for each strategy and then summed up to get the total value locked in LP tokens.

The resulting TVL can be compared to the TVL on the Webapp of Grizzly.fi: https://app.grizzly.fi/en
In the top left corner, there is the project's internal calculation of the TVL shown which is approximately the same as the script for DefiLlama.

### DIFF
--- a/projects/grizzlyfi/index.js
+++ b/projects/grizzlyfi/index.js
@@ -2,11 +2,21 @@ const sdk = require("@defillama/sdk");
 const { transformBscAddress } = require("../helper/portedTokens");
 const { stakings } = require("../helper/staking");
 const { pool2 } = require("../helper/pool2");
-const { unwrapUniswapLPs } = require("../helper/unwrapLPs");
+const { unwrapUniswapLPs, unwrapLPsAuto } = require("../helper/unwrapLPs");
+const { BigNumber } = require("bignumber.js");
 
-const abi = "uint256:grizzlyStrategyDeposits"
+const abiGrizzly = "uint256:grizzlyStrategyDeposits"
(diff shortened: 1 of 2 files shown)


llamatester:
The adapter at projects/grizzlyfi exports TVL: 
        
 
 
    bsc-staking               9.72 M
    bsc                       16.33 M
    staking                   9.72 M
    
    total                    16.33 M

llamatester:
The adapter at projects/grizzlyfi exports TVL: 
        
 
 
    bsc-staking               9.73 M
    bsc                       16.30 M
    staking                   9.73 M
    
    total                    16.30 M

PR #5723, opened 2023-03-27
fix: add ethereum and aggregator vault floats

0xSaga:
Sector Finance

add ethereum factory

add aggregator vault floats, which are not included in TVL calculation if only using strategy tvls


llamatester:
The adapter at projects/sector-fi exports TVL: 
        
 
 
    ethereum                  26.49 k
    moonriver                 94.93 k
    arbitrum                  2.66 M
    optimism                  493.01 k
    
    total                    3.27 M

PR #5767, opened 2023-03-31
Magnum - Sharpe Labs

Anirudha619:
MagETH is a vault that allows users to leverage their staked ETH position on Lido stETH through Aave Flashloans and p2p lending infrastructure Morpho, maximising user returns while ensuring healthy collateral and leverage.

Above script calculates the TVL of vault.


llamatester:
The adapter at projects/SharpeLabs exports TVL: 
        
 
 
    ethereum                  355.58163650071145
    
    total                    355.58163650071145

PR #5779, opened 2023-04-01
Add new uCRV pounder

0xAlunara:
* Adds support for the new Llama Airforce uCRV pounder.


llamatester:
        
 
 
    ethereum                  27.54 M
    
    total                    27.54 M

PR #5859, opened 2023-04-08
list zks-miner project

ZksMiner:
Zks Miner is the first protocol that offers the best utility - PASSIVE INCOME from all Staking protocols. We looked at every single protocol created and designed the perfect model. We want to be the leader in Defi with the best PASSIVE INCOME protocol on the Zksync ecosystem that has ever existed. We have combined a high fixed withdrawal amount and token appreciation to make this protocol sustainable and long-lasting.

Zks Miner provides ETH Staking right in your wallet, the token is always in your wallet so there is no need to put it in a centralized authority. All you need to do is buy, hold and automatically get rewards in your wallet so there is no more complicated staking process.

In addition, the Gamefi, NFT models will be updated and will create a strong and dynamic DeFi ecosystem in the zkSync era, creating an era of financial freedom. Every transaction at Zks Miner is unjust and Guaranteed with Zksync  oversight, transparency and clarity

### DIFF
--- a/projects/zks-miner/index.js
+++ b/projects/zks-miner/index.js
@@ -1,0 +1,47 @@
+const sdk = require('@defillama/sdk');
+
+const SILVER_MINER = "0x0A23aF664eFA875B36f8e82e1D52FBaE34607928";
+const DIAMOND_MINER = "0xF375De0ceeA7ED48D1074b777Fc37e7978f7Ba80";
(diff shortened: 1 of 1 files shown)


llamatester:
The adapter at projects/zks-miner exports TVL: 
        
 
 
    era                       55.7445
    
    total                    55.7445

PR #5890, opened 2023-04-11
updating methodology with correct address

TJakubek:
Just updating the methodology to have the correct factory address. Previously V1 was shown, V2 has more liquidity and is the one currently used


llamatester:
The adapter at projects/flarex exports TVL: 
        
 
 
    songbird                  2.49 M
    
    total                    2.49 M

PR #5939, opened 2023-04-14
Fix TVL calculation for rocifi-v2

katamarinaki:
One unfortunate mistake led to an incorrect TVL calculation. This PR fixes it.


llamatester:
The adapter at projects/rocifi-v2 exports TVL: 
        
 
 
    polygon-borrowed          8.03 k
    polygon                   55.94 k
    borrowed                  8.03 k
    
    total                    55.94 k

PR #5951, opened 2023-04-16
Polynomial: split earn and trade

gdeschryver:
Trying to display the two sources as two separate tabs, much like ribbon earn/lend or the different versions of aave.
Not sure this is the correct way.


llamatester:
The adapter at projects/polynomial-trade exports TVL: 
        
 
 
    optimism                  2.29 M
    
    total                    2.29 M

PR #6115, opened 2023-04-28
Use Illuminate PTs total supply as proxy for TVL

sourabhmarathe:
To follow up our PR from last week which properly valued the tokens within our Pools, we've also included the same logic to value the remaining TVL that has been deposited to Illuminate but not committed as liquidity.

---------------------------------

The current implementation:
1. Calculates the TVL of a pool's underlying balance (yearn shares)
2. Calculates the price of a pool's iPT
3. Calculates the TVL of a pool's iPT balance using iPT price

Extending this, the TVL of the wider protocol would be the pool's underlying TVL (shares balance) + (iPT price * iPT total supply), as opposed to solely the balance of a given pool.

----------------------------------

To clarify --

Illuminate works by allowing users to deposit an external Principal Token (Notional, Pendle, etc.,), which then mints an equivalent amount of iPTs. Typically, iPTs are sold at a discount on a YieldSpace AMM, and this discount represents the fixed-rate yield derived from providing the underlying asset until maturity. 

So the minting process of iPTs establishes the core of our TVL, while any additional TVL would come in the form of underlying capital committed to a pool's liquidity

This PR accurately reflects this logic by retaining the previous logic, and extending iPT pricing to `iPT.totalSupply()` as opposed to `iPT.balanceOf(pool)` -- better representing the total amount of liquidity being deposited into our protocol.


llamatester:
The adapter at projects/illuminate-fi exports TVL: 
        
 
 
    ethereum                  288.36 k
    
    total                    288.36 k

PR #6150, opened 2023-05-03
Update ergodex.js

yasha-black:
We have updated API method, so there is no need to divide value by 100 anymore


llamatester:
The adapter at projects/ergodex.js exports TVL: 
        
 
 
    ergo                      1.76 M
    
    total                    1.76 M

PR #6215, opened 2023-05-10
Update Gamma Adapter for Polygon zkEVM

StrategicReserve:


llamatester:
Error while running adapter at projects/visor: 
        
 
  
    
    
    Error: 
        Unknown chain(s): zkEVM
        Note: if you think that the chain is correct but missing from our list, please add it to 'projects/helper/chains.json' file

StrategicReserve:
oops! fixing name

PR #6242, opened 2023-05-11
Fix balance key in Caviar adapter

0xmichalis:


llamatester:
The adapter at projects/caviar exports TVL: 
        
 
 
    ethereum                  1.63 M
    
    total                    1.63 M

PR #6286, opened 2023-05-17
Algomint update to monitor new ETH & BTC baskets

clooneyr:
This is an update to our current TVL monitor. We have recently launched our new BTC and ETH baskets. This update now monitors the TVL held in the basket address's.


llamatester:
The adapter at projects/algomint exports TVL: 
        
 
 
    algorand                  0
    
    total                    0

PR #6291, opened 2023-05-17
fix:support  ot v2

HAYWAEL:
We have released v2 version, currently both v1 and v2 are running on our website, so please add support for it；


llamatester:
The adapter at projects/OT exports TVL: 
        
 
 
    era                       543.84 k
    
    total                    543.84 k

PR #6303, opened 2023-05-18
Remove hardcoded balances from Risk Harbor

MikeR808:
Remove hardcoded LUNA balances from Risk Harbor and switch to querying the vaults wrapped token balance directly.

Current code is checking a config value on each vault which has hardcoded values to ~5m. TVL did not update after a massive transfer of LUNA out of the protocol today to return to community pool so have switched it to use only actual balances.

### DIFF
--- a/projects/risk-harbor/index.js
+++ b/projects/risk-harbor/index.js
@@ -108,6 +108,7 @@ const networks = {
     ],
   },
   terra2: {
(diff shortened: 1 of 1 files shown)


llamatester:
The adapter at projects/risk-harbor exports TVL: 
        
 
 
    fantom                    26.123670585285
    arbitrum                  15.95 k
    terra2                    186.73214
    terra                     0
    aurora                    3.8221398300539997
    avax                      55.42 k
    ethereum                  18.61 k
    optimism                  8.99 k
    
    total                    99.19 k

PR #6366, opened 2023-05-24
update TokenPocket staking value

ycg5250:
We have two types of staking methods. Previously, the total staked amount was the sum of staking at least 32 ETH in a single transaction. Now, we have also included the total staked amount from small stakes. The combined staked amount from these two methods represents the total number of ETH we have currently staked.

### DIFF
--- a/projects/TokenPocket/index.js
+++ b/projects/TokenPocket/index.js
@@ -11,7 +11,19 @@ async function tvl() {
       }
     )
   )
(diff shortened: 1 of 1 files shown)


llamatester:
The adapter at projects/TokenPocket exports TVL: 
        
 
 
    ethereum                  8.16 M
    
    total                    8.16 M

PR #6417, opened 2023-05-30
Update GLIF tvl project post-launch

Schwartz10:
This is an update to an existing project - GLIF / Infinity Pool. Up until Monday the 29th, the Infinity Pool was only open to depositors, and not borrowers. On Monday, the borrow side opened up to Filecoin Storage Providers, and now, we need to update our TVL calculations to include the total funds locked by storage providers as borrowing collateral. We explicitly reduce tvl by the amount of funds the pool has issued to SPs to avoid double count.


llamatester:
        
 
 
    filecoin                  5.45 M
    
    total                    5.45 M

PR #6447, opened 2023-06-01
Listing of Tropykus in zkEVM

dacarva:
The reason for this PR is to add the deployment of Tropykus in zkEVM. There is an importante change between v1  in RSK and v2 in zkEVM. 
In v2 we forked the AAVE v2 contracts, whereas in v1 we forked and modified the contracts from Compound

### DIFF
--- a/projects/tropykus/index.js
+++ b/projects/tropykus/index.js
@@ -1,7 +1,22 @@
+const sdk = require('@defillama/sdk');
+const { aaveChainTvl } = require('../helper/aave');
 const { usdCompoundExports } = require('../helper/compound')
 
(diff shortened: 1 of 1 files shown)


llamatester:
The adapter at projects/tropykus exports TVL: 
        
 
 
    polygon_zkevm-borrowed    5.31 k
    polygon_zkevm             2.31 k
    rsk-borrowed              822.91 k
    rsk                       1.40 M
    borrowed                  828.22 k
    
    total                    1.41 M

dacarva:
Thanks for the response. Do you suggest we create another folder for Tropykus v2?

PR #6554, opened 2023-06-15
Fixing silver price calculation

aetherplex:
Might be easier for one of the Llamas to make this change since it looks like my formatting is making it hard to see the diff. Just need to change `grams: 1,` to `grams: ozToGrams` because I've since realised the silver price is returned in ounces not grams, which is inflating our TVL. Line 22 of `projects/meld-gold/index.js`


llamatester:
The adapter at projects/meld-gold exports TVL: 
        
 
 
    algorand                  333.15 k
    
    total                    333.15 k

PR #6716, opened 2023-07-04
readd cornucopias and update cardano

eserilev:
Hello, My name is Bill and I work for chainport

There have been some issues with TVL on defillama that have severly impacted the way people view our breidge New clients think we only have $9 million in TVL when in fact we have well over $100 million in TVL. This impacts our business as it makes us seem like a much smaller bridge than we actually are.

The main discrepancy is because you stopped tracking Cornucopias token and the OBI token

Cornucopias is one of our longest partners, they initially bridged their whole market cap from ETH to BNB using the Chainport bridge. They then decided to move half of their token volume from BNB to Cardano. Our bridge has minting permissions on both BNB and Cardano for their token, AND their native tokens are still locked on our bridge on ETHEREUM. I dont understand why you guys decided to comment out the COPI token from this adapter, and this merge request should rectify this issue.

There is a similar issue for the OBI token, though it is much less in terms of TVL

I also added our relevant cardano bridge addresses to this PR

### DIFF
--- a/projects/chainport/index.js
+++ b/projects/chainport/index.js
@@ -39,8 +39,8 @@ const config = {
         ADDRESSES.ethereum.BUSD,
         "0x8B3870Df408fF4D7C3A26DF852D41034eDa11d81",
         "0x0000000DE40dfa9B17854cBC7869D80f9F98D823",
-        // "0x42Baf1f659D765C65ADE5BB7E08eb2C680360d9d", //CORNOPIA 
-        // "0xBb3A8FD6Ec4bF0FDc6Cd2739b1e41192D12B1873", OBI 
+        "0x42Baf1f659D765C65ADE5BB7E08eb2C680360d9d", // CORNUCOPIAS 
+        "0xBb3A8FD6Ec4bF0FDc6Cd2739b1e41192D12B1873", // OBI 
         "0x3496B523e5C00a4b4150D6721320CdDb234c3079",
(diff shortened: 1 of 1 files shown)


llamatester:
The adapter at projects/chainport exports TVL: 
        
 
 
    cardano                   43.69 k
    bsc                       668.20 k
    ethereum                  126.59 M
    fantom                    27.967475911891253
    polygon                   2.77 k
    
    total                    127.30 M

eserilev:
@g1nt0ki hi, this is Bill. we spoke on Twitter regarding Chainports TVL on Cardano. Please see my comments above

0xngmi:
The change you are suggesting would count 114.20 M in TVL coming from copi (which would be ~90% of your TVL), however COPI's marketcap is only 23M according to https://www.coingecko.com/en/coins/cornucopias, so what you'd be counting as TVL would be the whole supply of COPI, including all coins that are locked. This goes against one of the core rules of our methodology, which you can find on https://docs.llama.fi/list-your-project/readme:
> We don't count any tokens that are not circulating or are yet to be issued

We also apply the same methodology to cross-chain tokens on multichain, where when the whole supply is bridged we cap what is counted as TVL, and for axelar, another bridge protocol, we don't count AXL, which is also a cross-chain token and would make up the majority of their TVL, so the same methodology is applied to all protocols on defillama. If you spot a protocol where we're not applying this methodology please let us know and we'll fix it.

PR #6725, opened 2023-07-05
marinade.finance: Show in native SOL

aankor:
Marinade.finance showed tokens distributed between native SOL and wSOL tokens because of a combination of using the sumTokens2 function counting native SOL balance as wSOL token on solana and custom balance entry using the plain solana key. This change will remove sumTokens2 usage and all balances will be summed into a single solana balance to show 100% TVL as a native solana token


llamatester:
        
 
 
    solana                    124.02 M
    
    total                    124.02 M

PR #6824, opened 2023-07-17
feat: add gauge staked lp token for crv pools

camolezi:
Context: HaloFi CRV pools stake the LP token during ongoing games. These pools were not accounted for in the DeFiLlama TVL calculation. The HaloFi API now returns the address of the gauge LP token for Curve pools that support gauge staking.

This is a small PR that incorporates the staked LP token in the TVL calculation.


llamatester:
The adapter at projects/good-ghosting exports TVL: 
        
 
 
    polygon                   112.91 k
    celo                      313.54 k
    
    total                    426.45 k

PR #6921, opened 2023-07-27
Fix tvl double subtract bug

Schwartz10:
Hi! We had a unknown bug occur on the DeFiLlama TVL because our server-side miner collaterals was already subtracting issued FIL, and the Adapter here then subtracts that issued FIL value _again_, so our TVL was showing up far too low.

You can see from our website that the numbers now align: https://glif.io


llamatester:
        
 
 
    filecoin                  12.77 M
    
    total                    12.77 M

PR #6965, opened 2023-08-01
Add Sobal to `Base` (Coinbase) network

joehquak:
This PR adds an additional network to an already existing listing for `Sobal`

Network added: `base`

Please could we also update our logo to: `https://github.com/Sobal/media/blob/main/logo/400px/png/logo-square-normal-400.png?raw=true`

Thank you!


joehquak:
Info missing while subgraph is building data, this is expected

PR #6995, opened 2023-08-02
Add Neutron TVL to Mars protocol

elfedy:
Add Neutron TVL (Red Bank only) using the same methodology as Osmosis for the Red Bank (Token balances as a good approximation of deposits - borrows).


llamatester:
        
 
 
    terra                     0
    neutron                   111.99 k
    osmosis                   8.44 M
    
    total                    8.55 M

PR #7325, opened 2023-08-30
Fix overidden asset quantity

anhtv-teko:


llamatester:
The adapter at projects/danogo exports TVL: 
        
 
 
    cardano                   435.51 k
    
    total                    435.51 k

PR #7331, opened 2023-08-30
Update subgraph url for term-finance

robotoer:
Updates subgraph URL to maintained term-finance graph node


llamatester:
The adapter at projects/term-finance exports TVL: 
        
 
 
    ethereum                  979.05 k
    
    total                    979.05 k

PR #7553, opened 2023-09-25
Add StakeStone TVL

luoyhang003:

TVL of StakeStone includes all ethers deposited by the users. It can be calculated as the sum of unallocated assets in the vault and deployed assets in the strategies.

##### Github org/user (Optional, if your code is open source, we can track activity): stakestone


llamatester:
The adapter at projects/stakestone exports TVL: 
        
 
 
    ethereum                  487.74 k
    
    total                    487.74 k

PR #7558, opened 2023-09-25
fix: tholgar TVL now counts redeemed AURA/CVX

0xtekgrinder:
Update Tholgar TVL as it was too big compared to the actual value because it counted the amount  of cvx / aura being redemed.


llamatester:
The adapter at projects/tholgar exports TVL: 
        
 
 
    ethereum                  10.27 k
    
    total                    10.27 k

PR #7670, opened 2023-10-10
Add wildbase

legend524:


legend524:
The tvl is correct!
but what about the base-pool2 and base-staking?

PR #7673, opened 2023-10-10
[Mars Protocol] - Update computations for marsv2 on Osmosis and add recent hallmarks

elfedy:
Since v2 launch on Osmosis, there are two updates to tvl:
1. Credit Manager does not expose a total vault coin balance query so the vault needs to be queried directly instead.
2. Now users can deposit to the credit manager. The token balances on credit manager are used to approximate these values (similar to what is done on the Red Bank).

Added two hallmarks as well (launch on neutron and mars v2 upgrade)

### DIFF
--- a/projects/mars/index.js
+++ b/projects/mars/index.js
@@ -21,6 +21,7 @@ const addresses = {
 async function osmosisTVL() {
   let balances = {};
   await addRedBankTvl(balances, 'osmosis');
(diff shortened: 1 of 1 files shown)


llamatester:
        
 
 
    terra                     0
    neutron                   99.78 k
    osmosis                   4.14 M
    
    total                    4.24 M

PR #7735, opened 2023-10-17
Add folks finance ultraswap

stefanofa:
Added [Ultraswap](https://app.folks.finance/ultraswap) tvl calculation, new feature under parent protocol Folks Finance.

Require https://github.com/DefiLlama/defillama-server/pull/4631


llamatester:
The adapter at projects/folks-ultraswap exports TVL: 
        
 
 
    algorand                  3.05 k
    algorand-borrowed         1.22 k
    borrowed                  1.22 k
    
    total                    3.05 k

stefanofa:
hey @g1nt0ki, thanks for the merge!

I had already tried to do this:
```ts
let data

async function getUltraswapData() {
  if (!data) {
    data = _getUltraswapData();
  }
  return data;
}
```
It seemed to me, however, that `tvl` and `borrowed` were called in parallel and thus neither of the 2 eventually found the data already cached

stefanofa:
> I think `data = await _getUltraswapData()` wouldnt work as it is async, but `data = _getUltraswapData()` should
> 
> Just added a log message and confirmed, it works (prints the debug message only once)

yes makes sense

PR #7773, opened 2023-10-19
fix: temp fix for vault tvl

0xKMG:
Since api.call and api.sumTokens is not return the token price, add code to fetch the price to calculate tvl


llamatester:
The adapter at projects/vaultka exports TVL: 
        
 
 
    arbitrum                  6.64 M
    
    total                    6.64 M

PR #7798, opened 2023-10-20
Update typus-finance v2

WayneAl:
4. Please fill the form below  **only if the PR is for listing a new protocol** else it can be ignored/replaced with reason/details about the PR

Update the new calculation method for v2 contract.


llamatester:
The adapter at projects/typus-finance exports TVL: 
        
 
 
    sui                       113.96 k
    
    total                    113.96 k

PR #7813, opened 2023-10-23
Muesli: Migrated to new price ep, as we are deprecating the old one

MuesliSwapTeam:
We are in the process of deprecating the list EP as it is designed in a way that introduces a lot of traffic costs on our side.

As defillama relies on this data, we have created a new EP that only returns the necessary price data for tokens that have a non-zero price, therefore keeping traffic at a minimum.

### DIFF
--- a/projects/muesliswap/index.js
+++ b/projects/muesliswap/index.js
@@ -33,13 +33,7 @@ async function adaTvl() {
     let totalAda = 0
 
     // fetch the prices of each traded token first
-    const tokenlistV2 = (await fetchURL("https://api.muesliswap.com/list?base-policy-id=&base-tokenname=")).data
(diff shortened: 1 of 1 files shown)


llamatester:
        
 
 
    milkomeda                 95.50 k
    cardano-staking           7.38 M
    cardano                   10.13 M
    staking                   7.38 M
    
    total                    10.23 M

MuesliSwapTeam:
> > We are in the process of deprecating the list EP as it is designed in a way that introduces a lot of traffic costs on our side.
> > As defillama relies on this data, we have created a new EP that only returns the necessary price data for tokens that have a non-zero price, therefore keeping traffic at a minimum.
> 
> thanks for creating a new endpoint for us and the this PR

Thanks for the fast response :)

PR #8233, opened 2023-11-30
fix Revest Finance Adapter for LPs not tracked by covalent

jhweintraub:
The Covalent API that Defillama uses doesn't always track the value of the LPs locked within the Revest locking system. As a result our TVL can fluctuate between millions and zero on a daily basis. This fix remedies that by manually tracking and unwraping the value of those LP tokens which are not on the Covalent token list. 

For our token vault v2 the balance of the LP is already tracked, we only added a single `balanceOf` call for our legacy tokenVault, and then a call to `unwrapUniswapLPs()`

### DIFF
--- a/projects/revest/index.js
+++ b/projects/revest/index.js
@@ -1,5 +1,5 @@
 const ADDRESSES = require('../helper/coreAssets.json')
-const { sumTokens2, } = require("../helper/unwrapLPs")
+const { sumTokens2, unwrapUniswapLPs } = require("../helper/unwrapLPs")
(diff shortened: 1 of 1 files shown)


llamatester:
The adapter at projects/revest exports TVL: 
        
 
 
    arbitrum                  0
    ethereum-staking          8.53 M
    avax                      16.0261
    optimism                  0
    fantom                    7.22 k
    polygon                   1.54627965
    ethereum                  1.55 M
    staking                   8.53 M
    
    total                    1.56 M

PR #8288, opened 2023-12-06
jexchange: no longer blacklist JEX

javier-nft:
We disabled our JEX distribution system (397M JEX remaining).
The TVL should be fine now.


llamatester:
The adapter at projects/jexchange exports TVL: 
        
 
 
    elrond                    118.09 k
    
    total                    118.09 k

PR #8567, opened 2024-01-04
Add staking data for lumi finance

stephen-lumi:
------ TVL ------
arbitrum-staking          4.68 M
arbitrum                  7.21 M
staking                   4.68 M

total                    7.21 M

### DIFF
--- a/projects/lumi-finance/index.js
+++ b/projects/lumi-finance/index.js
@@ -6,12 +6,11 @@ const Arb2CRVGauge = "0xCE5F24B7A95e9cBa7df4B54E911B4A3Dc8CDAf6f";
 const Arb2CRVLUAUSDLP = "0xD2239B95890018a8f52fFD17d7F94C3A82f05389";
 
 const staking = sumTokensExport({ ownerTokens: [[[Arb2CRVGauge, Arb2CRVLUAUSDLP], StakeContract]] });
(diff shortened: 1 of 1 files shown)


realdealshaman:
Hey, @stephen-lumi thanks for the PR, staking tvl should not be counted within the tvl export, it only needs to be exported as staking

alberthuang24:
@realdealshaman This one is a pledge of our agreement, if we don't count our tvl will be severely underestimated

alberthuang24:
When a stake, we transfer the money from the AMM contract to the stake contract.

realdealshaman:
In this case it is getting counted twice, we have toggles that apply to all projects that have staking and once they are toggled on the staking gets added to the tvl, in this case it will be counted twice

stephen-lumi:
How do I turn this switch on?

realdealshaman:
its on our main page and will also be in the protocol page once it is live
image

Currently we only need to export staking as staking

llamatester:
The adapter at projects/lumi-finance exports TVL: 
        
 
 
    arbitrum-staking          4.86 M
    arbitrum                  5.45 M
    staking                   4.86 M
    
    total                    5.45 M

realdealshaman:
@stephen-lumi I think I misunderstood the adapter at first. But what I want to convey is that we must be careful not to double-count the TVL

llamatester:
The adapter at projects/lumi-finance exports TVL: 
        
 
 
    arbitrum-staking          4.88 M
    arbitrum                  2.59 M
    staking                   4.88 M
    
    total                    2.59 M

PR #8621, opened 2024-01-10
Moved Boosted stake from borrowed to actual value inside the contract

elRaulito:
Moved the values from "Borrowed" to "Active" because the ADA is actually locked in the contract and not sent to the users


llamatester:
The adapter at projects/fluidtokens exports TVL: 
        
 
 
    cardano-borrowed          872.46 k
    cardano                   1.99 M
    borrowed                  872.46 k
    
    total                    1.99 M

PR #8661, opened 2024-01-15
fix the calculation of alp value

0xKMG:


llamatester:
The adapter at projects/vaultka exports TVL: 
        
 
 
    arbitrum                  29.62 M
    
    total                    29.62 M

PR #8873, opened 2024-02-06
morpho-blue markets: increasing the list of markets to be considered

tomrpl:
Hello,

I added few markets in the section. 
However I have 2 concerns by running the following command:

node test.js projects/morpho-blue/index.js

- 1. I think the wbIB01 token (https://etherscan.io/address/0xcA2A7068e551d5C4482eb34880b194E4b945712F), deployed 34 days ago is not taken into account: basically a wrapper of the ib01 token. 
- 2. I tried to remove the 'whitelisting ids' part of the script, however no 'CreateMarket' events are logging properly on my end. is there anything you can do? (basically consider all id from the 'CreateMarket' event to be taken into account in the tvl)

Thanks a lot and at you disposal

### DIFF
--- a/projects/morpho-blue/addresses.js
+++ b/projects/morpho-blue/addresses.js
@@ -1,12 +1,24 @@
-const wethWstethChainlinkAdaptiveCurveIRM945 =
+const wstethWethChainlinkAdaptiveCurveIRM945 =
   "0xc54d7acf14de29e0e5527cabd7a576506870346a78a11a6762e2cca66322ec41";
-const usdcWstethChainlinkAdaptiveCurveIRM860 =
+const wstethUsdcChainlinkAdaptiveCurveIRM860 =
   "0xb323495f7e4148be5643a4ea4a8221eef163e4bccfdedc2a6f4696baacbc86cc";
+const weethWethChainlinkAdaptiveCurveIRM860 =
+  "0x698fe98247a40c5771537b5786b2f3f9d78eb487b4ce4d75533cd0e94d88a115";
+const wbib01UsdcChainlinkAdaptiveCurveIRM965 =
+  "0x495130878b7d2f1391e21589a8bcaef22cbc7e1fbbd6866127193b3cc239d8b1";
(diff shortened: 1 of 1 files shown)


llamatester:
The adapter at projects/morpho-blue exports TVL: 
        
 
 
    ethereum-borrowed         24.84 M
    ethereum                  35.89 M
    borrowed                  24.84 M
    
    total                    35.89 M

g1nt0ki:
hi @tomrpl can you check if there is some bug in the event abi? I am getting this hash/topic (`0xa6d3dcbe752e0c7db0a44e056ff228aaaa08424d9d9edefd95523f9db69d41f3` ) with the event abi: `'event CreateMarket(bytes32 indexed id, (address loanToken, address collateralToken, address oracle, address irm, uint256 lltv) marketParams)'` 

but when I check etherscan, it matches nothing. wIB01 is a pain, it does not look like it is following any spec, we might need to add a manual coins adapter if you want to track it, but so far, not much tvl there?

tomrpl:
hey @g1nt0ki,

here is wha we are supposed to get:

-> cast sig 'CreateMarket(bytes32 id, (address,address,address,address,uint256) marketParams)'        
0xac4b2400 

image

However, weirdly enough I didn't manage to display any log from the getLogs function for the morpho blue contracts.
did you manage to get something on your side?

Thanks a lot for your reactivity

g1nt0ki:
@tomrpl I think I know the issue (dunno the solution though) when our library/ethers v6 tries to convert event abi string into topics hash, it messes up, mainly because it is not good at handling tuples? 

Didnt have time to dig deeper, so as a work around, passing correct topic as one of the arguments: https://github.com/DefiLlama/DefiLlama-Adapters/commit/9a3177b4c143e060caad4896b00c99ed0914543e 

Hopefully this adapter need never be updated again (we can create a new adapter for future versions/new products by morpho :) )

tomrpl:
Thanks a lot

PR #9057, opened 2024-02-24
Add new manta market on manta mainnet

ShoebillFinance:
Add manta lsd market on manta mainnet


llamatester:
Error while running adapter at projects/shoebillFinance-v2: 
        
 
  
    
    
    Error: Error: No RPCs available for wemix

PR #9071, opened 2024-02-26
Update TVL fix - Contract Finance

abdulrafay-contrax:
**NOTE**

We were getting less than real TVL, found that the token balances we were getting was from our vaults only. So we now used different contract function to get the balance, which is giving us correct balance including our governance etc.


llamatester:
        
 
 
    arbitrum                  17.06 k
    
    total                    17.06 k

PR #9306, opened 2024-03-12
add phar-v1 pools

0xShad0w:
Hi again! Another one! Ty

Also would it be possible to change Ramses logo as well?

image_2024-03-12_09-33-35

### DIFF
--- a/projects/pharaoh-exchange-v1/index.js
+++ b/projects/pharaoh-exchange-v1/index.js
@@ -1,0 +1,10 @@
+const {getUniTVL} = require('../helper/unknownTokens')
+const { staking } = require('../helper/staking')
+
+module.exports = {
+  misrepresentedTokens: true,
+  avax:{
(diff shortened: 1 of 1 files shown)


llamatester:
The adapter at projects/pharaoh-exchange-v1 exports TVL: 
        
 
 
    avax-staking              414.65 k
    avax                      87.57689311004012
    staking                   414.65 k
    
    total                    87.57689311004012

Define101:
@0xShad0w template information ....

0xShad0w:
Hey @Define101! My bad, I thought since they were all existing protocols that it would be easy for you to identify and I didn't have to fill in the template, but seems like it isn't the case and I should have at least provided some basic information. Will do it the right way next time, thank you. 

Would it be possible to rename them instead of V3 -> V2 and V2 -> V1 for all 3 projects? (NILE, PHAR and CLEO)

Thanks again!

Define101:
@0xShad0w Hello hello. Just informing us that we already track them is more than enought. If you also could provide the display name and twitter/website links that would be perfect 
Regarding ramses logo. Should be good within 2 hours

PR #9555, opened 2024-03-26
Update STFIL tvl

samuerio:
Due to an oversight in the previous TVL calculation method, borrowings from some users were not accurately accounted for. Hence, we have conducted this update.


llamatester:
The adapter at projects/stfil exports TVL: 
        
 
 
    filecoin                  97.37 M
    
    total                    97.37 M

PR #9606, opened 2024-03-29
feat: fix Staq wrong address

cokes123:
Sorry, I submitted an incorrect parameter before, which caused the tvl calculation error. 
This pre address is the stake amount of our validator, not the total stake amount.


llamatester:
The adapter at projects/bitstaq exports TVL: 
        
 
 
    map                       1.81 M
    
    total                    1.81 M

PR #10090, opened 2024-05-03
Use timestamp instead of block height

Schwartz10:
This PR uses timestamp instead of block height for purposes of backfilling historical data. GLIF project adapter had a bug in its historical calculation, so backfilling with this adapter is both (1) highly efficient and (2) precise and accurate. [Here's](https://discord.com/channels/823822164956151810/823822165538373696/1235992539350175866) the relevant conversation from the DefiLlama team in discord.

### DIFF
--- a/projects/glif/index.js
+++ b/projects/glif/index.js
@@ -7,10 +7,10 @@ module.exports = {
   methodology:
     "The GLIF Pools protocol is a liquid leasing protocol for Filecoin that requires borrowers to collateralize FIL in order to borrow for their storage providing operation. This TVL calculation adds the total amount of FIL deposited into the protocol, and the total amount of locked FIL collateral by borrowers, to arrive at TVL.",
   filecoin: {
-    tvl: async (_, height, _1, { api }) => {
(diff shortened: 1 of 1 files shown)


llamatester:
        
 
 
    filecoin                  221.31 M
    
    total                    221.31 M

projects/glif/index.js
```diff
@@ -7,10 +7,10 @@ module.exports = {
   methodology:
     "The GLIF Pools protocol is a liquid leasing protocol for Filecoin that requires borrowers to collateralize FIL in order to borrow for their storage providing operation. This TVL calculation adds the total amount of FIL deposited into the protocol, and the total amount of locked FIL collateral by borrowers, to arrive at TVL.",
   filecoin: {
-    tvl: async (_, height, _1, { api }) => {
+    tvl: async (timestamp, _, _1, { api }) => {
```
0xngmi, line 10: you cant use timestamp like that, instead you should get the timestamp from the api object

0xngmi (changes requested):

0xngmi:
PR applied in fd9cca4b8

PR #10151, opened 2024-05-09
Update Genius Yield adapter: added new address

4TT1L4:
**Problem:**
- The current implementation does not take v1.1 on-chain orders into account.

**Solution:**
- Take both v1.0 and v1.1 on-chain orders into account by calculating the TVL locked in both variants.


llamatester:
The adapter at projects/genius-yield exports TVL: 
        
 
 
    cardano-staking           3.68 M
    staking                   3.68 M
    cardano                   118.55 k
    
    total                    118.55 k

PR #10353, opened 2024-05-23
Fix processing CHAR and other Toucan tokens

0xmichalis:
399a3cf318e3a188b3d3232dd85b1242b0a2be8c was incorrectly setting supplies in
the destructured variables (bct, nct, char)
as for some networks (eg., base) not all
three tokens are found. Here I am fixing this
issue so now the final struct returned by
getCalculationMethod is created dynamically
and always contains the exact tokens with
the correct info.

### DIFF
--- a/projects/toucan-protocol/config.js
+++ b/projects/toucan-protocol/config.js
@@ -16,7 +16,20 @@ const CONFIG_DATA = {
         nct_bridge: "0xdC1Dfa22824Af4e423a558bbb6C53a31c3c11DCC"
     },
 };
+const TOKEN_DATA = {
+    bct: {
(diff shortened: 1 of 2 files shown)


llamatester:
The adapter at projects/toucan-protocol exports TVL: 
        
 
 
    polygon                   1.16 M
    celo                      198.92 k
    regen                     107.22 k
    base                      17.01 k
    
llamatester:
The adapter at projects/toucan-protocol exports TVL: 
        
 
 
    polygon                   1.16 M
    celo                      199.44 k
    regen                     107.28 k
    base                      17.09 k
    

PR #10464, opened 2024-06-02
Update bouncebit  tokens

leeamijo:
stBBTC is the Liquid Staking Derivative (LSD) of BBTC generated by staking nodes within the BounceBit chain.

### DIFF
--- a/projects/basin/index.js
+++ b/projects/basin/index.js
@@ -1,0 +1,34 @@
+const { getLogs2 } = require('../helper/cache/getLogs');
+
+// Aquifer
+const FACTORY = "0xba51aaaa95aeefc1292515b36d86c51dc7877773";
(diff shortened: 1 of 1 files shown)


Define101:
What is the difference betweet that token and 0x7F150c293c97172C75983BD8ac084c187107eA19

leeamijo:
> What is the difference betweet that token and 0x7F150c293c97172C75983BD8ac084c187107eA19

The address 0x8f083EaFcbba2e126AD9757639c3A1E25a061A08 corresponds to the Wrapped stBBTC Token. Using stBBTC directly in certain scenarios can cause issues with the calculation of BB Token rewards, so it needs to be wrapped. The asset value remains consistent before and after wrapping.

PR #10791, opened 2024-06-27
add sei to uni v3

Define101:


llamatester:
Error while running adapter at projects/uniswap: 
        
 
  
    
    
    Error: Llama RPC error! methos: undefined 
    - host: https://evm-rpc.sei-apis.com error: height is not available (requested height: 78353999, base height: 82909999)
    - host: https://evm-rpc.sei-apis.com/ error: height is not available (requested height: 78353999, base height: 82909999)

PR #10865, opened 2024-07-03
Update Vaults Steer - Contract Finance 

abdulrafay-contrax:
contrax steer vaults addresses updated


llamatester:
        
 
 
    arbitrum                  278.75 k
    
    total                    278.75 k

PR #10899, opened 2024-07-05
use proxy for reject request

0xpeluche:


projects/helper/proxyRequest.js
```diff
@@ -0,0 +1,56 @@
...
+  const proxyOptions = getProxyOptions();
+  const agent = new https.Agent({
+    rejectUnauthorized: false
+  });
+
+  const response = await axios({
+    method: 'get',
+    url: url,
+    httpsAgent: agent,
+    proxy: {
```
0xngmi, line 31: why copy the object like this and not just do `proxy:proxy`?

PR #10949, opened 2024-07-10
Fix: Xave-Finance update

0xpeluche:
Added permitFailure to the Data multicall for `CurveFactory` pools which sometimes returned null, as well as a checker to skip if data is null


llamatester:
The adapter at projects/xave-finance exports TVL: 
        
 
 
    polygon                   454.18 k
    avax                      320.03 k
    ethereum                  262.47 k
    
    total                    1.04 M

PR #10951, opened 2024-07-10
fix dopex-clamm update

0xpeluche:
Fix for the Dopex-clamm query, the request to the endpoint did not seem to work, slight refactor by separating the logics, and using recursive instead of do-while


llamatester:
The adapter at projects/dopex-clamm exports TVL: 
        
 
 
    arbitrum                  2.37 M
    
    total                    2.37 M

PR #10992, opened 2024-07-15
Fix stFil update

0xpeluche:
Fix stFil update: retrieve the information from the chain rather than the outdated API

image

5,329,478 * price (4.28) ~~ $22.8M


llamatester:
The adapter at projects/stfil exports TVL: 
        
 
 
    filecoin                  22.76 M
    
    total                    22.76 M

PR #11002, opened 2024-07-16
[ TVL update ] add ST-SBUCK saving vault TVL

Jarekkkkk:
add underlying BUCK TVL of saving-vault to current Strater TVL


llamatester:
The adapter at projects/strater exports TVL: 
        
 
 
    sui                       7.83 M
    
    total                    7.83 M

projects/strater/index.js
```diff
@@ -92,8 +94,18 @@ async function tvl(api) {
   const stapearlUSDTAmount =
     (stapearlLpAmount * stapearlReserveY) / stapearlLpSupply;
 
+  // saving vault
+  const savingVaultcObj = await sui.getObject(ST_SBUCK_VAULT_OBJECT_ID)
+  let savingVaultTVL = Number(savingVaultcObj.fields.free_balance) + Number(savingVaultcObj.fields.time_locked_profit.fields.locked_balance) + Number(savingVaultcObj.fields.time_locked_profit.fields.unlocked_balance)
+  const strategies = savingVaultcObj.fields.strategies.fields.contents
+  for(const strategy of strategies){
+    const botrrowedAmount = Number(strategy.fields.value.fields.borrowed)
+    savingVaultTVL += botrrowedAmount
```
g1nt0ki, line 103: hi @Jarekkkkk thanks for the update, it is mostly good except for this part, we cant count borrowed as tvl, it is leverage. 
image

In this example, we can count only the deposited 10 SUI as tvl not the borrowed 10 SUI

g1nt0ki (changes requested):
.

projects/strater/index.js
Jarekkkkk: > hi @Jarekkkkk thanks for the update, it is mostly good except for this part, we cant count borrowed as tvl, it is leverage. image
> 
> In this example, we can count only the deposited 10 SUI as tvl not the borrowed 10 SUI

Hi @g1nt0ki, thx for the reply

actually the vault you're referring to is different from the one we add.

[saving_vault_page](https://www.strater.xyz/vault/auto-compound?type=ST-SBUCK)

above is the vault page on our site, kindly take a look. the TVL only accounts for the deposited BUCK, there aren't any assets being borrwed under the hood. 

let me know if there's any questions !

thank !

g1nt0ki: ah, all good then, thanks for the update.

PR #11017, opened 2024-07-17
Update LineHub Perps store address

0xrmvdao:
We've updated Store address for LineHub Perps. Please help to merge.


llamatester:
The adapter at projects/linehub-perps exports TVL: 
        
 
 
    linea                     735
    
    total                    735.0088554180332

g1nt0ki (changes requested):
retain old storage address as well, else refill will give wrong results

PR #11021, opened 2024-07-17
Update Cygnus Finance with adding TON TVL

greatpie:
## Update TVL

Cygnus Finance is now available on TON, so we've added TVL stats for TON. 
See more at [Cygnus TON](https://ton.cygnus.finance/mint)


llamatester:
The adapter at projects/cygnus-finance exports TVL: 
        
 
 
    base                      31.32 M
    ton                       252.51 k
    
    total                    31.57 M

PR #11124, opened 2024-07-27
🐛 slipstream positions got ignored

j-vp:
The if-branch never got entered since the `positionsIds` is defined in the function arguments -> on a 'normal' `sumTokens2(resolveSlipstream=True)`, they never get resolved.

-> move into the if branch when length is 0


llamatester:
Error while running adapter at projects/arcadia-finance-v2: 
        
 
  
    
    
    [object Object]

llamatester:
Error while running adapter at projects/arcadia-finance-v2: 
        
 
  
    
    
    [object Object]

PR #11134, opened 2024-07-29
update thala lsd logic

LawsonGraham:
update logic to count the whole of the delegation pool udner the hood of thala apt and sthapt


llamatester:
The adapter at projects/thala-lsd exports TVL: 
        
 
 
    aptos                     142.16 M
    
    total                    142.16 M

PR #11144, opened 2024-07-30
Fix: Bean update

0xpeluche:
Add a handler to prevent errors when the `underlying` prop is undefined which breaks the subsequent multicall


llamatester:
The adapter at projects/bean exports TVL: 
        
 
 
    ethereum-staking          25.14 M
    staking                   25.14 M
    ethereum-pool2            265.58 k
    pool2                     265.58 k
    ethereum                  0
    
    total                    0

PR #11150, opened 2024-07-30
fix: Adapter Proxy [Outdated]

0xpeluche:
> Small fix using only on-chain methods to solve the adapter, even if the BTCpx price could be unknown


llamatester:
The adapter at projects/prxy exports TVL: 
        
 
 
    polygon                   17.63 M
    polygon-staking           64.80 k
    staking                   64.80 k
    ethereum                  22.52 k
    
    total                    17.66 M

PR #11190, opened 2024-08-02
update addresses for affine defi

kiwi0401:
We are just wanting to update contract addresses for our restaking product as they have changed


llamatester:
The adapter at projects/affine-defi-liquid exports TVL: 
        
 
 
    ethereum                  770.38 k
    
    total                    770.38 k

PR #11235, opened 2024-08-08
fix: Stader

0xpeluche:
> Since the latest fix on Stader, they have now completely removed the data from their API that was being returned with errors, keeping only the chains on which they are still active

> rm terra2, rm near

https://universe.staderlabs.com/common/tvl

### DIFF
--- a/projects/stader/index.js
+++ b/projects/stader/index.js
@@ -7,20 +7,12 @@ async function getData() {
   return _res;
 }
 
-async function hbarTvl(timestamp) {
+async function hbarTvl() {
   const res = await get("https://universe.staderlabs.com/common/tvl");
   return {
     "hedera-hashgraph": res.hedera.native,
   };
 }
-
-async function fantomTvl() {
-  const res = await getData();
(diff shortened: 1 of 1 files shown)


llamatester:
The adapter at projects/stader exports TVL: 
        
 
 
    ethereum                  358.64 M
    hedera                    24.92 M
    bsc                       12.74 M
    
    total                    396.30 M

g1nt0ki (changes requested):
the way our server works atm is, if some chain is no longer present in tvl adapter, we will not show the chain in the UI

So, if we merge this PR, all past fantom tvl will vanish, this is the reason, you will find this code in lot of places `() => ({})` instead of removing a chain

llamatester:
The adapter at projects/stader exports TVL: 
        
 
 
    ethereum                  360.42 M
    hedera                    24.99 M
    bsc                       12.81 M
    fantom                    0
    terra2                    0
    near                      0
    
    total                    398.23 M

PR #11267, opened 2024-08-12
fix Swell rswETH tvl query to support withdrawals

darvinrio:
Withdrawals went live for Swell Liquid Restaking. 
The current methodology doesn't reflect the actual ETH number. 

Hence updating methodology to show the post Withdrawn amount


llamatester:
The adapter at projects/swell-restaking exports TVL: 
        
 
 
    ethereum                  301.15 M
    
    total                    301.15 M

PR #11505, opened 2024-09-03
Update TVL Calculation to Exclude Unstaked Amounts

0xyongyong:
### Summary of Changes:
- Previously, the `getTotalStakingAmount` function was used to calculate the TVL (Total Value Locked).
- However, there was a bug where the unstaked amounts were also included in the TVL calculation.
- To fix this issue, the function has been changed to `getTotalClaimable` for more accurate TVL calculation.

### Reason for Changes:
- The `getTotalStakingAmount` function was inaccurately reflecting the actual locked value (TVL) by including amounts that had already been unstaked.
- This led to an overestimation of the TVL.
- By switching to the `getTotalClaimable` function, we can ensure that only the amounts still locked and claimable are included, resulting in a more accurate TVL calculation.


llamatester:
The adapter at projects/LairFinance exports TVL: 
        
 
 
    klaytn                    27.53 M
    
    total                    27.53 M

PR #11523, opened 2024-09-04
add underlying pool assets to TVL of Mars Protocol

linkielink:
This PR adds the underlying assets of the pool shares, which are deposited via the farm feature on Mars Protocol.
Since the share tokens have no price in the DefiLlama token registry, they must be split into the underlying tokens to be added to the TVL. 
Share tokens are not borrowable, so no deduction is taken.


llamatester:
        
 
 
    osmosis                   12.70 M
    neutron                   2.39 M
    terra                     0
    
    total                    15.09 M

PR #11660, opened 2024-09-17
fix: bril-finance (outdated)

0xpeluche:
> Added permitFailure on the multicall, with error handling to prevent null response


llamatester:
The adapter at projects/bril-finance exports TVL: 
        
 
 
    bsc                       728.05 k
    
    total                    728.05 k

PR #11663, opened 2024-09-17
fix: DeltaPrime (outdated)

0xpeluche:
> Added a permitFailure on multicalls + filter out null responses  (the ABIs were probably removed or paused from the contracts to contain the hack, which results in fails on all multicalls on Arbitrum)


llamatester:
The adapter at projects/deltaprime exports TVL: 
        
 
 
    avax                      22.19 M
    arbitrum                  8
    
    total                    22.19 M

PR #11664, opened 2024-09-17
fix: ynLSDe TVL

johnnyonline:
fix YieldNest's ynLSDe TVL.

### DIFF
--- a/projects/yieldnest/index.js
+++ b/projects/yieldnest/index.js
@@ -2,7 +2,6 @@
 const ADDRESSES = require('../helper/coreAssets.json')
 
 const YN_ETH = '0x09db87a538bd693e9d08544577d5ccfaa6373a48'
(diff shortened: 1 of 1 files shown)


llamatester:
The adapter at projects/yieldnest exports TVL: 
        
 
 
    ethereum                  11.62 M
    
    total                    11.62 M

danoctavian:
Clarification:

The previous code was just recording balances of the assets deposited into the ynEigen.sol Vault contract.

https://github.com/yieldnest/yieldnest-protocol/blob/main/src/ynEIGEN/ynEigen.sol#L225

However, yieldNest restakes the assets into Eigenlayer and those balances should be counted as well towards TVL.

https://github.com/yieldnest/yieldnest-protocol/blob/main/src/ynEIGEN/AssetRegistry.sol#L271

Recently all our assets were restaked so the TVL was not displayed properly anymore (almost 2m worth of ETH less).

PR #11722, opened 2024-09-24
upgrade awaken indexerUpgrade Awaken's Indexer to v2

harold-awaken:
We need to upgrade Awaken's indexer to version v2, as the old v1 version will soon be deprecated.


llamatester:
The adapter at projects/awaken exports TVL: 
        
 
 
    aelf                      2.59 M
    
    total                    2.59 M

PR #11746, opened 2024-09-25
fix:💥breaking change on the API - update symbol

rcheiko:
We are going to support multiple markets for a single asset, so the symbol for an asset might be incorrect for some case

I've also deleted those variables since we aren't using them, and we are going to delete those fields from the assets query in the next API release
```
qTokenName
qTokenCurrencySymbol
```

### DIFF
--- a/projects/liqwid/index.js
+++ b/projects/liqwid/index.js
@@ -40,11 +40,9 @@ const query = `query($input: MarketsInput)  {
         page
         pagesCount
         results {
+          id
           asset {
             id
-            symbol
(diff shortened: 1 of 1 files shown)


llamatester:
The adapter at projects/liqwid exports TVL: 
        
 
 
    cardano                   26.27 M
    cardano-borrowed          7.54 M
    borrowed                  7.54 M
    cardano-staking           3.82 M
    staking                   3.82 M
    
    total                    26.27 M

PR #11786, opened 2024-09-30
Fix: Quicksilver (outdated)

0xpeluche:
Filter out undefined balances


llamatester:
The adapter at projects/quicksilver exports TVL: 
        
 
 
    quicksilver               473.46 k
    
    total                    473.46 k

PR #11835, opened 2024-10-06
Added base pools for maha.xyz

senamakel:
Currently there's about 170k$ of liquidity in the Aerodrome pools on base but for some reason the MAHA, USDC and USDz TVL does not show up.

Probably because the LP token's price isn't live yet. However we have made the PR adding these pools in.

image

### DIFF
--- a/projects/helper/tokenMapping.js
+++ b/projects/helper/tokenMapping.js
@@ -50,6 +50,7 @@ const fixBalancesTokens = {
   },
   base: {
     [ADDRESSES.base.rETH]: { coingeckoId: 'rocket-pool-eth', decimals: 18 },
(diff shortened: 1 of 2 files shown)


llamatester:
The adapter at projects/mahaxyz exports TVL: 
        
 
 
    ethereum                  178.85 k
    ethereum-pool2            170.67 k
    pool2                     170.67 k
    base-pool2                0
    
    total                    178.85 k

Rezacb0413 (approved):

g1nt0ki:
hi @senamakel 
I made a minor fix, but it is still lower because zai token price is not pulled from the correct pool in coingecko, can you get in touch with their team? 
image

PR #12072, opened 2024-10-28
Fix: Kogefarm logic

0xpeluche:
The adapter hasn't been updated for nearly 20 days: It appears that the logic using the balance method from the contracts' ABI on Moonriver was no longer returning any results. The logic has been changed to use the balanceOf method from the strategies associated with the pools. The balances are exactly the same as before, but this workaround bypasses the issue with the non-functional ABI method

### DIFF
--- a/projects/kogefarm/abi.json
+++ b/projects/kogefarm/abi.json
@@ -1,4 +1,0 @@
-{
-  "balance": "uint256:balance",
-  "token": "address:token"
-}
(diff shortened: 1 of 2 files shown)


llamatester:
The adapter at projects/kogefarm exports TVL: 
        
 
 
    polygon                   201.77 k
    moonriver                 23.94 k
    polygon-staking           3.78 k
    staking                   3.78 k
    polygon-pool2             2.05 k
    pool2                     2.05 k
    fantom                    900
    kava                      20
    
    total                    226.63 k

PR #12073, opened 2024-10-28
Fix: Oh-finance

0xpeluche:
The TVL of the Oh-finance adapter hasn't been updated for about 20 days. One of the addresses was returning errors during on-chain calls. A small refactor was done to use multicalls instead of single calls, with a permitFailure option in case of failure, and added error handling to eliminate null responses


llamatester:
The adapter at projects/oh-finance exports TVL: 
        
 
 
    avax                      2.58 k
    ethereum                  0
    metis                     0
    moonriver                 0
    
    total                    2.58 k

PR #12192, opened 2024-11-07
echelon support FA TVL tracking

LawsonGraham:
echelon previously hardcoded to expect only coin types... this change supports our FA too which tracks truAPT now


llamatester:
The adapter at projects/echelon exports TVL: 
        
 
 
    aptos-borrowed            128.49 M
    borrowed                  128.49 M
    aptos                     74.44 M
    
    total                    74.44 M

PR #12213, opened 2024-11-08
Fix: Velvet v3

0xpeluche:
There was an address on Base that is not a token at all, but an EOA: `0x96af5739ca66ca55ab71ac9f308720d5044995ee` which prevented the rest of the code from executing.
Small refactor and addition of this address to the blacklist


llamatester:
The adapter at projects/velvet-capital-v3 exports TVL: 
        
 
 
    base                      850.44 k
    bsc                       79.00 k
    
    total                    929.44 k

PR #12347, opened 2024-11-19
Fix: Estate-protocol (Outdated)

0xpeluche:
There was a `propertyAddress` among the items that did not match the EVM format and was non-null, which bypassed the filter. Added a more robust regex to ensure that if an address is not EVM-compliant, it will not pass to the subsequent multicall, preventing the code from breaking


llamatester:
The adapter at projects/estate-protocol exports TVL: 
        
 
 
    arbitrum                  1.75 M
    
    total                    1.75 M

PR #12423, opened 2024-11-25
Fix: Small fix Cian yield

0xpeluche:
Adapter has not been updated for a week due to an issue with the `totalAssets` method call on the contract `0xd4Cc9b31e9eF33E392FF2f81AD52BE8523e0993b`. From Etherscan, we can see that the contract returns 0 on this method, which is likely blocking subsequent calls. A `permitFailure` has been added to prevent the code from being blocked

### DIFF
--- a/projects/cian-yl/index.js
+++ b/projects/cian-yl/index.js
@@ -18,8 +18,6 @@ module.exports = {
 
 Object.keys(config).forEach((chain) => {
     module.exports[chain] = {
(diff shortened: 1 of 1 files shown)


llamatester:
The adapter at projects/cian-yl exports TVL: 
        
 
 
    ethereum                  170.15 M
    arbitrum                  9.63 k
    
    total                    170.16 M

PR #12571, opened 2024-12-05
[Bucket] add new collateral - sbETH 

Jarekkkkk:
add new collateral native ETH ( sbETH )


llamatester:
The adapter at projects/bucket-protocol exports TVL: 
        
 
 
    sui                       45.48 M
    
    total                    45.48 M

PR #12640, opened 2024-12-10
Fix: SailingProtocol

0xpeluche:
Removing the `ETHE` asset that no longer exists in the protocol and therefore returns an error in the API call, which can’t retrieve its price


llamatester:
The adapter at projects/sailingprotocol exports TVL: 
        
 
 
    kava                      283.02 k
    
    total                    283.02 k

PR #12659, opened 2024-12-10
add folks galgo hallmarks

palace22:
Added Folks gAlgo hallmarks for Algorand governance recommitments

### DIFF
--- a/projects/folks-galgo/index.js
+++ b/projects/folks-galgo/index.js
@@ -15,3 +15,15 @@ module.exports = {
     },
   },
 };
+
+module.exports.hallmarks = [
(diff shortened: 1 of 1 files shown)


g1nt0ki (changes requested):
Hi @palace22 

we try to keep the hallmarks to a minimum, else users would just disable feature, how many of these have an impact on the chart?

palace22:
> Hi @palace22
> 
> we try to keep the hallmarks to a minimum, else users would just disable feature, how many of these have an impact on the chart?

Ehy @g1nt0ki, all of them, each will explain the tvl drop, as you can see https://defillama.com/protocol/folks-finance

PR #12695, opened 2024-12-13
add MARS perps counter party vault

linkielink:
Added the Perps Counter Party Vault, a vault users fund with USDC with a lockup period of 10 days.
These funds should be counted towards the protocol's total TVL.

### DIFF
--- a/projects/mars/index.js
+++ b/projects/mars/index.js
@@ -10,6 +10,7 @@ const contractAddresses = {
   neutron: {
     params: 'neutron1x4rgd7ry23v2n49y7xdzje0743c5tgrnqrqsvwyya2h6m48tz4jqqex06x',
     redBank: 'neutron1n97wnm7q6d2hrcna3rqlnyqw2we6k0l8uqvmyqq6gsml92epdu7quugyph',
+    perps: 'neutron1g3catxyv0fk8zzsra2mjc0v4s69a7xygdjt85t54l7ym3gv0un4q2xhaf6'
   },
 };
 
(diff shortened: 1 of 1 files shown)


llamabutler:
        
 
 
    osmosis                   24.21 M
    neutron                   6.57 M
    terra                     0
    
    total                    30.77 M

linkielink:
The Counter Party Vault is currently funded with 500,000 USDC. So, the TVL should be 500k higher than the current implementation.

Screenshot 2024-12-13 at 10 46 20
Which seems to be the case

g1nt0ki (commented):
this is a perp product? if so, better to create a new listing for it. Also, the website is down?

g1nt0ki:
* https://app.marsprotocol.io/

g1nt0ki:
I have created a new listing.

linkielink:
The App does block the usage of VPNs (maybe that's why you can't access it?) 
Perps is launching today in about 2.5h.
The perps contracts are deployed however and the Perps platform will be part of the app.
So I would disagree creating a new listing as it is part of the Mars Protocol's feature set.

g1nt0ki:
hi @linkielink booooo for blocking VPNs \s

It will still be shown as part of part, we will create a sublisting, like what we have for jupiter perp: https://defillama.com/protocol/jupiter-perpetual-exchange

linkielink:
> hi @linkielink booooo for blocking VPNs \s
> 
> It will still be shown as part of part, we will create a sublisting, like what we have for jupiter perp: https://defillama.com/protocol/jupiter-perpetual-exchange

Ah ok, that makes sense then.
I'll add the Cumulative 24h Trading Volume endpoint to the adapter then as well.

PR #12724, opened 2024-12-16
[WIP - Pending crosscheck] - Fix: HyperliquidSpot

0xpeluche:
**Methodology:**

- Through various APIs, we retrieve the assets and their contexts, allowing us, via mapping, to consolidate both the tokens and their price relative to USDC.
- In a second step, we retrieve all the tokens still present on the HIP-2 order-spot contract.

From there, we calculate the sum of all tokens, for example:
https://app.hyperliquid.xyz/explorer/token/0xba3aaf468f793d9b42fd3328e24f1de9.

image

The `total` variable allows us to determine what remains on the contract, meaning the difference:
Hyperliquidity Genesis Balance - Hyperliquidity Balance Sold. Here, we multiply the remaining assets by their price

We process this for each object, which ultimately provides the TVL present on the contract, representing what is still available on the spot DEX order book.

### DIFF
--- a/projects/hyperliquid-spot/index.js
+++ b/projects/hyperliquid-spot/index.js
@@ -1,40 +1,39 @@
-const ADDRESSES = require('../helper/coreAssets.json')
 const axios = require('axios')
 
-const USDC = ADDRESSES.ethereum.USDC
 const API_URL = 'https://api.hyperliquid.xyz/info'
 
 const assetsInfos = async () => {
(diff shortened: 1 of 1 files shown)


llamabutler:
The adapter at projects/hyperliquid-spot exports TVL: 
        
 
 
    hyperliquid               179.02 M
    
    total                    179.02 M

PR #12903, opened 2024-12-28
DackieSwap update Dackie address and staking contract

DackieSwap:


llamabutler:
The adapter at projects/dackieswap-v2 exports TVL: 
        
 
 
    base                      207.01 k
    wc                        59
    linea                     58
    inevm                     58
    arbitrum                  56
    optimism                  42
    blast                     39
    xlayer                    38
    mode                      18
    base-staking              0
    staking                   0
    
    total                    207.38 k

waynebruce0x:
is it possible to keep the old factory as well as adding the new factory here? then we can more easily backfill any data as required

DackieSwap:
@waynebruce0x 
DackieSwap just migrate token to new contract, so we also has new staking contracts, cannot use the old contract, it's ok if data calculated from now, no need backfill.

Refer our tweet here: https://x.com/DackieSwap/status/1869236545011462148

PR #12905, opened 2024-12-28
update Stability: add base, real, sonic

a17:
Also our main porject category is "Yield aggregator" now, not "Liquidity manager".


llamabutler:
The adapter at projects/stability exports TVL: 
        
 
 
    sonic                     39.59 k
    polygon                   571
    base                      387
    real                      22
    
    total                    40.57 k

PR #12963, opened 2025-01-04
Add new denoms to Quicksilver

joe-bowman:
- Add new denominations to quicksilver adapter
- Handle exponents properly, instead of hardcoding


llamabutler:
The adapter at projects/quicksilver exports TVL: 
        
 
 
    quicksilver               639.90 k
    
    total                    639.90 k

PR #12979, opened 2025-01-06
Fix: Graphene

0xpeluche:
Fix for `Graphene by Velocimeter`
The recent addition of Iota prevented the code from executing properly. Modified on-chain data calls to use on-chain methods instead of logs wherever possible, as it is less resource-intensive for updates


llamabutler:
The adapter at projects/graphene exports TVL: 
        
 
 
    iotaevm                   1.12 M
    base                      49.53 k
    fantom                    5.38 k
    mantle                    243
    
    total                    1.17 M

PR #12983, opened 2025-01-06
Fix: Idex-v4

0xpeluche:
Fix for the Idex-v4 protocol, which hasn’t been updated for 3 weeks. Using the xchain’s REST API (IDEX) to fetch the data directly


g1nt0ki (changes requested):
the issue was caused by incorrect rpc configuration. fixed it in the sdk: https://github.com/DefiLlama/defillama-sdk/commit/af6673caa8696b856a680acef422514eafd66395

PR #13004, opened 2025-01-07
Update gAlgo hallmarks with Gov 14

palace22:
Added Algo Gov 14 hallmark ts.


llamabutler:
The adapter at projects/folks-galgo exports TVL: 
        
 
 
    algorand                  6.73 M
    
    total                    6.73 M

PR #13021, opened 2025-01-08
Fix: Pixel-Swap (FIX)

0xpeluche:
Small fix on `pixel-swap`, which hasn’t been updated for a few days 
> There was a balance returning null, causing an error that prevented the rest from executing

null


llamabutler:
The adapter at projects/pixel-swap exports TVL: 
        
 
 
    ton                       38.62 k
    
    total                    38.62 k

llamabutler:
The adapter at projects/pixel-swap exports TVL: 
        
 
 
    ton                       38.34 k
    
    total                    38.34 k

Issue #13026, opened 2025-01-08
Sonic DEX on ICP: TVL issue


realdealshaman:
User Reports Bot
Protocol: Sonic
Data: TVL
What's wrong: TVL value is wrong .
Correct data: https://v3.data.sonic.ooo/
https://defillama.com/protocol/sonic

0xpeluche:
The adapter is functional and reflects the TVL of Sonic v1, but there is a v3. I will look into integrating it

**V1:**
https://data.sonic.ooo/

![image](https://github.com/user-attachments/assets/43c1d82c-56d3-4fc0-99d8-a33f83dbc63f)

PR #13045, opened 2025-01-09
Fix&Update: Sonicooo (ICP)

0xpeluche:
#13026

Fix for v1 which returned nulls for the most recent responses, preventing the display of the response. Added v3


llamabutler:
The adapter at projects/sonicooo exports TVL: 
        
 
 
    icp                       1.99 M
    
    total                    1.99 M

PR #13050, opened 2025-01-10
add new vault to TVL calculation

0xKMG:


llamabutler:
        
 
 
    solana                    13.80 M
    
    total                    13.80 M

PR #13061, opened 2025-01-12
update ratex tvl

ratex-tony:
use underlying tokens to calculate the tvl of unsupported tokens


llamabutler:
The adapter at projects/rate-x exports TVL: 
        
 
 
    solana                    25.96 M
    
    total                    25.96 M

waynebruce0x:
we're adding price support for these assets so there will be no need to unwrap them

PR #13091, opened 2025-01-14
Fix: Clober liquidity vault update

0xpeluche:
Fix of the adapter that had been non-functional since the Dec 2024 hack. Added a hallmark to specify the event


llamabutler:
        
 
 
    base                      0
    
    total                    0.38687480603554897

PR #13135, opened 2025-01-17
Add Launch of LeverageX and Remove old Chain

rogi-sh:
We decided to remove the previous chain because if you look at TVL it looks like the project has died. Since we only operate now on Base, it is useful to remove the old chain so the correct picture of the project on Base is reflected.


llamabutler:
The adapter at projects/javsphere exports TVL: 
        
 
 
    base-vesting              5.61 M
    vesting                   5.61 M
    base-staking              2.16 M
    staking                   2.16 M
    base                      396.89 k
    
    total                    396.89 k

g1nt0ki (changes requested):
https://github.com/DefiLlama/DefiLlama-Adapters/discussions/13020

PR #13138, opened 2025-01-17
Fix: Arcadia-v2

0xpeluche:
Arcadia-v2 has not been updated for a week due to the introduction of accounts with a different structure that breaks the `generateAssetData` multicall


llamabutler:
        
 
 
    base                      9.06 M
    
    total                    9.06 M

PR #13156, opened 2025-01-18
add inElixir to vaults

ungaro:
add inElixir to vaults (can't share the page as it's private for institutions)


llamabutler:
The adapter at projects/nest-staking exports TVL: 
        
 
 
    ethereum                  80.65 M
    
    total                    80.65 M

PR #13157, opened 2025-01-18
feat: shapeshift multichain update

firebomb1:
Adds the missing data from the various chains where the FOX token, ShapeShift DAO's native token, lives (Ethereum, Aribtrum, Gnosis and Polygon) and where it has locked value (even if none qualifies as our protocol TVL per DefiLlama's methodology):
 
- A new ETH/FOX LP staking staking farm on Ethereum (`0xe7e16e2b05440c2e484c5c41ac3e5a4d15da2744`). This is an evergreen contract so we shouldn't need to add new ones on Ethereum after that (hopefully).
- RFOX. This is a "FOX on Arbitrum" (`0xf929de51d91c77e42f5090069e0ad7a09e513c73`) staking contract deployed on Arbitrum (Proxy: `0xac2a4fd70bcd8bab0662960455c363735f0e2b56`, Implementation:`0x4f9c6a6cc987de98c8109e121516008906a899c9`). It is single asset staking, 28 days lock minimum, with yield distributed in RUNE on Thorchain. It serves as a way for our DAO to do both revenue sharing with stakers and FOX burning monthly.

Few things of note:
- I've moved the ETH/FOX LP token staking in `ethereum` to `staking` because they are effectively a staking farm and not just a liquidity pool... and since we should not double count them, under `pool2` we have another pool that is not used for  farming/staking.
  - I would like to know the benefits of using `staking` to list these? Aren't we missing out on counting tokens that might be in the Uniswap v2 pool but aren't staked in any of the staking farms? If so, shouldn't we instead just remove all the staking/farms and only list the two pools under `pool2` (Uniswap v2 and v3)?
- I've this PR it and the values appear to be correct for all chains but I would appreciate a sanity check, especially around the RFOX proxy/contract, as it is a custom contract/implementation. It seems to work with `stakings()`, but I do not know all of the details of this functions and maybe the particularities of the RFOX contract.

If that might be useful for the tests, we maintain a dashboard with the LP on these various chains here: https://flipsidecrypto.xyz/Rayyyk/shapeshift-lp-dashboard-ciW_N4 . It isn't live data, it has to be refreshed, and doesn't list USD values for the FOX share of the pools unfortunately, but it could still be of use. For RFOX we maintain a dashboard here with the total FOX locked: https://dune.com/shapeshift/rfox

### DIFF
--- a/projects/shapeshift/index.js
+++ b/projects/shapeshift/index.js
@@ -1,38 +1,92 @@
-const { staking } = require("../helper/staking");
-const { pool2 } = require("../helper/pool2");
+const { stakings } = require("../helper/staking");
+const { pool2s } = require("../helper/pool2");
 
-// Contracts
-const stakingUNIv2Contracts = [
-  "0xDd80E21669A664Bce83E3AD9a0d74f8Dad5D9E72", // v1
-  "0xc54b9f82c1c54e9d4d274d633c7523f2299c42a0", // v2
-  "0x212ebf9fd3c10f371557b08e993eaab385c3932b", // v3
-  "0x24FD7FB95dc742e23Dc3829d3e656FEeb5f67fa0", // v4
-  "0xC14eaA8284feFF79EDc118E06caDBf3813a7e555", // v5
-  "0xEbB1761Ad43034Fd7FaA64d84e5BbD8cB5c40b68", // v6
-  "0x5939783dbf3e9f453a69bc9ddc1e492efac1fbcb", // v7
-  "0x662da6c777a258382f08b979d9489c3fbbbd8ac3", // v8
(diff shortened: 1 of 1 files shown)


llamabutler:
The adapter at projects/shapeshift exports TVL: 
        
 
 
    staking                   5.60 M
    ethereum-staking          2.94 M
    arbitrum-staking          2.66 M
    pool2                     245.97 k
    arbitrum-pool2            96.23 k
    polygon-pool2             76.56 k
    xdai-pool2                73.17 k
    ethereum-pool2            0
    ethereum                  0
    arbitrum                  0
    polygon                   0
    xdai                      0
    
    total                    0

PR #13205, opened 2025-01-22
Fix: Splash-protocol

0xpeluche:


llamabutler:
The adapter at projects/splash.js exports TVL: 
        
 
 
    cardano                   45.86 M
    
    total                    45.86 M

PR #13286, opened 2025-01-28
Mellow update vaults

0xsign:
Move vaults to ERC4626 compatible


llamabutler:
        
 
 
    ethereum                  426.24 M
    
    total                    426.24 M

PR #13326, opened 2025-01-31
Add CreateMyToken Adapter

akshatmittal:
Adding adapter for [CreateMyToken](https://www.createmytoken.com) specifically noting that this only tracks TVL for Fair Launch tokens, and only for when the token agrees to use public factory. Additionally the adapter only counts paired native token and does not include the created token itself (otherwise the TVL would technically be inflated)

---


llamabutler:
The adapter at projects/createmytoken exports TVL: 
        
 
 
    base                      16.39 k
    ethereum                  11.67 k
    bsc                       1.91 k
    polygon                   222
    arbitrum                  193
    avax                      35
    optimism                  8
    blast                     0
    
    total                    30.43 k

PR #13361, opened 2025-02-04
Add Torch Finance V2 DEX

WaterSo0910:
We are already listed on DeFiLlama. Now that we have launched our new DEX V2, we need to update the TVL calculation accordingly.

[DeFiLlama - Torch Stable Swap](https://defillama.com/protocol/torch-stable-swap)

image


llamabutler:
The adapter at projects/torch exports TVL: 
        
 
 
    ton                       629.95 k
    
    total                    629.95 k

PR #13367, opened 2025-02-04
Mellow update vaults

0xsign:
- add rsuniBTC
- add rsPOND
- set migrated vaults in both mellow-lrt and simple-lrt groups to prevent "hole" in TVL. `permitFailure` should handle the smooth transition between migrated vaults


llamabutler:
        
 
 
    ethereum                  41.29 M
    
    total                    41.29 M

PR #13428, opened 2025-02-09
Remove vaults page and clean up reporting

vvalecha519:
Questions:
1) ebtc lrt tvl is calculated here: updateEbtcTvl(api) and there are many vaults that should double count ebtc (ie line 94) therefore they should not increase total tvl but they should be part of vault tvl. How would I achieve this?

2) Before our tvl oracle was reporting all vaults non-eeth and non-ebtc in liquid whereas now it will report non-eeth in liquid and non-ebtc in liquid separately and it reports the debt in the liquid vaults. We do not want the migration to the new version to mess up our graph when we using the other method. Should I have a condition of using the new method if block > X and if block < X use old method? 

3) Can you confirm the total tvl before merging and confirm the graph is not messed up in terms of eth tvl


llamabutler:
        
 
 
    ethereum                  2.19 B
    
    total                    2.19 B

0xngmi:
> should not increase total tvl but they should be part of liquid tvl. How would I achieve this?

we have a system where certain tagged assets can be counted as doublecounted on a protocol and excluded from the parent tvl calculation, you should use that

> Should I have a condition of using the new method if block > X and if block < X use old method?

it would be better to just backfill the data, otherwise there will be jumps in the chart that arent correct

PR #13436, opened 2025-02-10
feat: Parasail use helper function to get TVL

hcheng826:
change to use a helper function to account TVL since we need to include some other components to the TVL calculations, such as node collaterals. Also to follow the same pattern as pFIL case


hcheng826:
@g1nt0ki thanks for merging the PR! It's been ~36 hrs since the PR was merged and i think the change is not yet reflected on the defillama web page. i noticed that for this PR there was no calculation preview generated by the Github Actions pipeline like it usually did. could it be somehow missing and also cause the code change not deployed? Thanks!

hcheng826:
issue fixed in https://github.com/DefiLlama/DefiLlama-Adapters/pull/13507

PR #13437, opened 2025-02-10
Fix: Beam-Dex

0xpeluche:


llamabutler:
The adapter at projects/beam-dex exports TVL: 
        
 
 
    zeta                      491.75 k
    
    total                    491.75 k

PR #13507, opened 2025-02-14
Fix Parasail helper contract address

hcheng826:
original contract address was wrong. update to the correct one


llamabutler:
The adapter at projects/repl exports TVL: 
        
 
 
    filecoin                  47.30 M
    fluence                   916.38 k
    arbitrum                  51.75 k
    swan                      49.90 k
    
    total                    48.32 M

PR #13545, opened 2025-02-17
Just update O.LAB adapter

ivan-alphaorbeta:
Just update adapter. Using USDC locked in the platform instead of the conditional tokens contract.


g1nt0ki (changes requested):
https://github.com/DefiLlama/DefiLlama-Adapters/discussions/432

PR #13640, opened 2025-02-23
Update index.js

bit-reserve:
feat add sonic chain support


llamabutler:
The adapter at projects/bit-reserve exports TVL: 
        
 
 
    core                      150.03 k
    sonic                     42
    
    total                    150.07 k

g1nt0ki (changes requested):
sorry, please compute tvl on chain and what is counted as tvl on sonic, it is not very clear.

PR #13676, opened 2025-02-25
feat: added clisBNB CDP Vault to Loop

0xd4n1el:
Added a TVL tracker for CDP Vault using Pendle clisBNB token. The tests shows that the token already has a price feed.


llamabutler:
The adapter at projects/loopfi exports TVL: 
        
 
 
    ethereum                  2.07 M
    bsc                       1.41 M
    scroll                    96.19 k
    
    total                    3.57 M

PR #13681, opened 2025-02-26
Update: Neemo (Calculate TVL before CCIP)

0xpeluche:
We strive to remain as fair as possible between adapters to prevent any unfair or unjust competition among DeFi protocols. We always prefer to count TVL on the chain where the collaterals are actually held.

I have rewritten the adapter to calculate Neemo Finance's TVL based on the deposit chains before the use of CCIP


llamabutler:
The adapter at projects/neemo-finance exports TVL: 
        
 
 
    astar                     2.79 M
    ethereum                  92.19 k
    
    total                    2.88 M

PR #13698, opened 2025-02-28
Create index.js - Adding Support for Onfa Chain to DeFiLlama

onfadev:
Add a new network Onfa chain to DeFiLlama


projects/onfa/index.js
```diff
@@ -0,0 +1,8 @@
+const { sumTokensExport } = require('../helper/unwrapLPs')
+const ADDRESSES = require('../helper/coreAssets.json')
+
+module.exports = {
+  onfa: {
```
waynebruce0x, line 5: please use key 'onfachain' to match chainlist and our sdk

onfadev: Hi, I have been updated

PR #13737, opened 2025-03-05
Update Binaryx Platform adapter to include Off-plan assets

dmitriy-green:
A new asset type has been added on Binaryx Platform - Off-plan (a property which is under construction). It should be included in the TVL.

image


llamabutler:
The adapter at projects/binaryx-platform exports TVL: 
        
 
 
    polygon                   3.35 M
    
    total                    3.35 M

PR #13748, opened 2025-03-05
update Cadabra Finance Adapter: count Cadabra-routed full-range TVL

chopachom:
Thank you @g1nt0ki for merging our previous PR (#13547)!

In our previous PR you've excluded the TVL of the Uniswap V3 pool with our token.
While we agree that our previous approach of counting the whole TVL of the UniV3 pool was a bit naive, there is important distinction: most of this pool's TVL is channeled via our wrapper contracts. Users deposit their funds via our full-range wrapper contract, which then channels this liquidity into Uniswap V3. In this context, we believe it is appropriate to consider this liquidity as part of our protocol's TVL, since it is not merely passively held on Uniswap V3 and we plan to seamlessly transition this liquidity to Uniswap V4 in the future. So effectively we use Uniswap as an underlying AMM layer. Given that we have marked it as doublecounted: true, we think this approach makes sense.

And there's another key point, this pool also has full-range liquidity that belongs to the protocol, it's not a treasury or any team-held allocation. This liquidity was formed through token purchases from our V1 version and, in essence, represents liquidity indirectly contributed by users to the protocol. This liquidity will remain in the pool, so we believe it's should be counted as TVL.

So this PR adjusts the TVL calculation to account for these two liquidity sources.

### DIFF
--- a/projects/cadabra/index.js
+++ b/projects/cadabra/index.js
@@ -2,9 +2,43 @@ const { getLogs2 } = require("../helper/cache/getLogs");
 
 const BSC_VAULT = '0x6f24f5EB8f50DE36C1D1FEcD31c423e1fc8ba4ae';
 const BSC_FROM_BLOCK = 45524968;
+const BSC_CADABRA_FULL_RANGE_PAIR_LENS = '0x2A479CFf64574Ed192f42509a205e3eb94Dc258C';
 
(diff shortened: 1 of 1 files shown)


g1nt0ki (changes requested):
thanks, we consider this as pool2 as one of the tokens is your gov token, more info: https://docs.llama.fi/list-your-project/what-to-include-as-tvl

I have moved it to pool2

PR #13752, opened 2025-03-05
Adding TVL for Milk Vaults

yy-analytics:
**NOTE**

---

> - If you would like to add a `volume` adapter please submit the PR [here](https://github.com/DefiLlama/adapters).

3. Please fill the form below  **only if the PR is for listing a new protocol** else it can be ignored/replaced with reason/details about the PR
4. **For updating listing info** It is a different repo, you can find your listing in this file: https://github.com/DefiLlama/defillama-server/blob/master/defi/src/protocols/data2.ts, you can  edit it there and put up a PR

---

This PR adds the new Yak Milk Vaults to Yield Yak's TVL values. The vaults themselves are for rsAVAX (restaked sAVAX), rBTC.b, rggAVAX and aiAVAX (for now - more to come which will be added to the subgraph). Of these tokens, only rsAVAX has price data so instead we're using the "base asset" (i.e. sAVAX for rsAVAX, BTC.b for rBTC.b etc., which is the token that is deposited into / withdrawn from the vaults) and calculating the amount of that asset using the shares (totalSupply) of the vault in combination with the exchangeRate for the vault's accountant contract. If any further explanation is needed please let me know. All of the new information for Milk Vaults here comes from our subgraph at the moment but if you prefer to make contract calls I can edit accordingly.

### DIFF
--- a/projects/yieldyak/index.js
+++ b/projects/yieldyak/index.js
@@ -2,8 +2,21 @@ const { staking } = require('../helper/staking');
 const { cachedGraphQuery, getConfig } = require('../helper/cache')
 const sdk = require('@defillama/sdk')
 
-const graphUrl = sdk.graph.modifyEndpoint('7oSYYdK5RKmqggdzFyfUnojP6puDAj31C4ezDGrgVfk9')
-const graphQuery = `{ farms(first: 1000) { id }}`;
+const graphUrl = sdk.graph.modifyEndpoint('https://gateway.thegraph.com/api/[api-key]/subgraphs/id/YC6LMdHjtWZ2cCq7YKFmDcDt2eCcBvSNSiWi56bnqoW')
+const graphQuery = `
+{
+  yakMilkVaults(first: 1000) {
+    totalSupply
+    decimals
+    accountant {
+      base {
+        id
+        symbol
+      }
+      exchangeRate
+    }
+  }
+}`;
 
 async function avaxTvl(api) {
   const farms = await getConfig('yieldyak/avax', 'https://staging-api.yieldyak.com/43114/farms')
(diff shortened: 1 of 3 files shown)


projects/yieldyak/index.js
```diff
@@ -15,14 +28,22 @@ async function avaxTvl(api) {
...
   const excludedSymbol = 'YAK';
 
   tokens.forEach((token, i) => {
     if (!token || !vals[i] || token === excludedAddress || symbols[i] === excludedSymbol) return;
     api.add(token, vals[i]);
   });
+
+  // For getting the vaults' TVL values we use the underlying token rather than the vault token symbol itself
+  // as price data is more readily available for these.
+  const { yakMilkVaults } = await cachedGraphQuery('yieldyak/milk-vaults-avalanche', graphUrl, graphQuery)
```
waynebruce0x, line 41: can we count token balances here rather than totalsupply

yy-analytics: I don't think it will be possible to count token balances for these contracts. They're based on the "Boring Vault" architecture from Veda. I'm pretty sure it needs to be totalSupply because that is the total number of shares for the vault and you get the number of underlying tokens by multiplying this by the exchange rate for the vault. e.g. a user entering our rsAVAX vault deposits/withdraws sAVAX to/from the vault. Shares (rsAVAX tokens) are minted according to the exchange rate on the accountant contract at that time. This exchange rate increases in value over time as rsAVAX accrues value against sAVAX. These "Boring Vaults" were designed to be pretty flexible but still have some controls (e.g. the exchange rate cannot change too often and cannot change by too much in one go).

The alternative would have been using the share tokens themselves as the token, which was what I tried during testing. But at the moment, only rsAVAX is recognised and has price data on Defillama it seems (I think because it's the only one with decent (incentivised) liquidity on a Dex), so the others, which still have decent levels of TVL, would not be counted if we just used the vault/share tokens themselves.

PR #13764, opened 2025-03-06
Feat: Hedgey

0xpeluche:
##### Website Link:
https://app.hedgey.finance/

##### Current TVL:
https://intel.arkm.com/explorer/entity/hedgey-finance


0xpeluche:
The adapter is really heavy, I had to batch in parallel multicalls to avoid a timeout reached because of Base + Polygon

image

PR #13765, opened 2025-03-06
Add aave v3 liquidations adapter

Cryptofowl:
HomicidalChicken

##### Twitter Link:
https://x.com/HomicidalChickn

##### Chain:
Ethereum, Polygon, Arbitrum, Base

##### Data sources: 
This adapter uses the Messari AAVE v3 subgraphs for Ethereum, Polygon, Arbitrum, and Base. 

##### methodology:
The adapter weights each collateral in a user's portfolio according to its composition and liquidation thresholds, and accordingly calculates the percentage drop in price required for that single collateral to result in a liquidation. 

The difference between current LTV and the weighted liquidation threshold is used to calculate the liquidation price, which is multiplied by the collateral weight percentage. 

https://github.com/Cryptofowl

##### Testing:
liquidations/test.ts was run successfully, but a change was required in the tsconfig.json to compile the typescript files. I have not included this change in the PR, assuming that the compiler is configured correctly for your environment. The change is as follows:
```
  "compilerOptions": {
    "module": "NodeNext",
    "moduleResolution": "NodeNext",
    "noImplicitAny": false,
    "strictNullChecks": false,
    "typeRoots": ["./typings/", "./node_modules/@types/"]
  }
  ```

### CODE REVIEW

File: liquidations/aave-v3/index.ts [hunk 1]
```diff
@@ -0,0 +1,278 @@
...
+  [Chains.ethereum]: {
+    name: "aave",
+    chain: Chains.ethereum,
+    usdcAddress: "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48",
+    subgraphUrl: sdk.graph.modifyEndpoint('JCNWRypm7FYwV8fx5HhzZPSFaMxgkPuw4TnR3Gpi81zk'), // Messari AAVE v3
+    explorerBaseUrl: "https://etherscan.io/address/",
+  },
+  [Chains.arbitrum]: {
+    name: "aave",
+    chain: Chains.ethereum,
```
waynebruce0x, line 86: is this chain value correct?

Cryptofowl: Good catch, thank you. Amended

PR #13795, opened 2025-03-08
feat: aave sonic into rings backing

0xtekgrinder:
Add aave sonic into rings backing.

Still needs to add defillama pricing of aave tokens


llamabutler:
The adapter at projects/rings exports TVL: 
        
 
 
    ethereum                  64.18 M
    sonic                     23.11 M
    
    total                    87.29 M

PR #13814, opened 2025-03-10
Puppyfun TVL adapter

giantHound:
Puppy.fun is an AI-powered memecoin launchpad of the BabyDoge ecosystem, offering seamless token creation, multi-DEX support, and secure liquidity locking. With built-in AI agents for engagement and Gems rewards for trading and holding, it gamifies the meme token experience like never before. 
Whether you're launching, trading, or engaging with the community, Puppy.fun makes memecoins smarter, safer, and more fun.

### DIFF
--- a/projects/puppyfun/index.js
+++ b/projects/puppyfun/index.js
@@ -1,32 +1,53 @@
-const axios = require('axios')
-
 const BNB_ADDRESS = '0x0000000000000000000000000000000000000000'
+const FACTORIES = ['0xbFC5229ab471c54B58481fA232CFd4a18371C51C']
 
-const apiBaseURL = 'https://bd-fun-defilama-ts-backend-main.puppy.fun/lama-api'
-const tvlMethod = '/tvl'
+const factoryAbi = {
+    allPairsLength: "function allPairsLength() view returns (uint256)",
+    allPairs: "function allPairs(uint256) view returns (address)"
+};
 
-const maxRetriesCount = 3
-
-module.exports = {
-  methodology: "TVL is counted as the total amount locked in PuppyFun smart contracts. In our particular case it is the total BNB amount in PuppyFun virtual DEX (all tokens which are not yet migrated). Calculated similarly to UniswapV2 based DEXes.",
-  bsc: {
-    tvl,
-  },
-}
+const pairAbi = {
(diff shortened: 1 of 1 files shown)


projects/puppyfun/index.js
```diff
@@ -0,0 +1,32 @@
...
+
+async function tvl(api) {
+  const queryURL = apiBaseURL + tvlMethod
+
+  for (let tryCount = 0; tryCount < maxRetriesCount; ++tryCount) {
+    try {
+      console.log(`Query TVL ...`)
+      const response = await axios.get(queryURL)
+      console.log(`Got response: tvlDirty: ${response.data.tvlDirty / 10 ** 18} BNB | tvlClean ${response.data.tvlClean / 10 ** 18} BNB`)
+      api.add(BNB_ADDRESS, response.data.tvlClean)
```
waynebruce0x, line 25: please use on chain methods to count contract token balances here

projects/puppyfun/index.js
```diff
@@ -0,0 +1,32 @@
...
+      break
+    } catch(err) {
+      console.log(`Error getting TVL`, err)
```
waynebruce0x, line 28: please remove any try/catch blocks so our monitoring systems work

giantHound: okay, no problem

projects/puppyfun/index.js
giantHound: we use on-chain log parsing in our collector service, there are some technical reasons why we do it this way (we have a database where the older versions of contracts are hosted together with the latest one). Can we leave API requests here for now as it is done in the Dimensions repo? or is it obligatory to perform on-chain calls only here?

waynebruce0x: you can use proprietary APIs to fetch config info but token balances must be read on chain

giantHound:
Hello! I have recently added a commit with On-chain TVL collection (using the in built api.call function). Also the try / catch mechanic is removed. Could you please confirm that the functionality is correct now?

llamabutler:
The adapter at projects/puppyfun exports TVL: 
        
 
 
    bsc                       31.27 k
    
    total                    31.27 k

PR #13837, opened 2025-03-12
Stout

waynebruce0x:


llamabutler:
The adapter at projects/stout exports TVL: 
        
 
 
    sonic-borrowed            2.58 M
    borrowed                  2.58 M
    sonic                     690.05 k
    sonic-staking             109.67 k
    staking                   109.67 k
    sonic-pool2               0
    pool2                     0
    
    total                    690.05 k

stout-protocol:
Hi @waynebruce0x! Could you kindly explain the changes you made in this commit?  

[[DefiLlama PR #13837 – Commit f5f13dd](https://github.com/DefiLlama/DefiLlama-Adapters/pull/13837/commits/f5f13dd69083233845795d03a5548c0d035ea464)](https://github.com/DefiLlama/DefiLlama-Adapters/pull/13837/commits/f5f13dd69083233845795d03a5548c0d035ea464)  

Thanks!

waynebruce0x:
https://github.com/DefiLlama/DefiLlama-Adapters/pull/13789#issuecomment-2714820730 
STTX, veSTTX and DUSX are all backed by the same USDC. Since the adapter already counts USDC as TVL it would be double, triple, quadruple counting to count STTX, veSTTX and DUSX too. This is why we dont count borrowed funds in core TVL for lending protocols.

stout-protocol:
There's a misconception about how the protocol works.  

Regarding **veSTTX and STTX**, they are **not** backed by USDC. Instead, they have a **floor price in DUSX**, which is entirely different. The value of STTX (and, by extension, veSTTX) is determined by the market and how it values the **Stout Protocol**. While STTX can trade above the floor price, if it ever falls below, it would present a free arbitrage opportunity, quickly driving the price back up.  

Regarding **DUSX**, it is a **decentralized stablecoin backed by overcollateralized loans**. The protocol includes a **PSM**, similar to DAI, but that doesn’t define its core backing.  

Both the **veSTTX in the vault contract** and the **DUSX in the floor contract** are real assets and are **not** being double-counted.

stout-protocol:
[[Link to comment](https://github.com/DefiLlama/DefiLlama-Adapters/pull/13789#issuecomment-2714820730)](https://github.com/DefiLlama/DefiLlama-Adapters/pull/13789#issuecomment-2714820730)

A good comparison would be **STTX (and veSTTX) to MKR**, **DUSX to DAI**, and **stDUSX to sDAI**, and all three are counted toward Aave's TVL.

waynebruce0x:
Yeah but MKR and DAI are from a seperate protocol, Maker. We dont count these assets towards Makers TVL, and we dont count AAVE towards Aave TVL

stout-protocol:
I see... in that case, shouldn't they be counted when the "Gov Tokens (Include governance tokens)" option is enabled?

waynebruce0x:
From my understanding:

- STTX is effectively a receipt token for depositing USDC to the PSM. Counting both STTX and USDC would be double count.
- veSTTX is a receipt token for depositing STTX to the vesting contract. Counting veSTTX and STTX would be double count. 
- DUSX is a CDP collateralised by the USDC in the PSM. Counting both DUSX and USDC would be double count.

stout-protocol:
With all due respect, @waynebruce0x, that understanding is not correct. It seems there's some confusion due to our TGE.  

**STTX is NOT backed by USDC.** USDC is simply used as a means to mint STTX — just like how people used to send ETH to an ICO contract. The ICO token wasn’t backed by ETH, and similarly, STTX is not backed by USDC.  

What may be causing confusion is that **STTX has a floor price**, which gives holders confidence in a **low maximum drawdown**. Additionally, since the market is down, **STTX is currently trading close to its floor price**, which might be reinforcing this misunderstanding.  

- **veSTTX** is a receipt token for depositing STTX, which is why we only counted **veSTTX in the vault**, while **STTX was counted in the staking formula**.  
- **DUSX is not collateralized solely by USDC in the PSM**, just as **DAI isn’t either**. Right now, only **veSTTX is being used as collateral to borrow DUSX**, but additional assets will be introduced this week.  

I'd be happy to jump on a call with you to clear up any misunderstandings!

waynebruce0x:
In that case we can add back the balance of STTX in the vesting contract - does that sound right?

stout-protocol:
Yes, that sounds right. And the **veSTTX in the Vault contract** should be counted if the **"Gov Tokens"** option is enabled, correct?  

Now, about the **DUSX in the Floor contract** — shouldn't this also be counted as TVL?

waynebruce0x:
veSTTX would be doublecounting the STTX, and DUSX would be counting the CDP collateral

PR #13855, opened 2025-03-13
Fix bug in helper dir coreTokens function

molefund:
This is a fix for bug :

coreTokens has been changed in recent couple of days, it contains bugs that a JSON not contain function "includes". We have to fix it right


llamabutler:
The adapter at projects/mole exports TVL: 
        
 
 
    sui                       10.69 M
    aptos                     1.61 k
    
    total                    10.69 M

molefund:
image

We have to fix this bug introduced by new coming commit : [13812](https://github.com/DefiLlama/DefiLlama-Adapters/pull/13812)

g1nt0ki:
hi @molefund sorry for introducing the bug, fixed it now, still keeping it as string array instead of as a single json string

PR #13868, opened 2025-03-14
remove invalidated collateral tokens from tvl calculation

ewansheldon:
removes tokens from tvl calculation, because they are no longer used in our project


llamabutler:
The adapter at projects/standard-smart-vaults exports TVL: 
        
 
 
    arbitrum                  783.99 k
    
    total                    783.99 k

waynebruce0x:
closing because we want to keep this for historical data

PR #13880, opened 2025-03-17
Mellow add Lido v3 vaults

0xsign:
add another batch of Lido V3 vaults


llamabutler:
        
 
 
    ethereum                  254.52 M
    
    total                    254.52 M

PR #13933, opened 2025-03-19
Fix Nolus TVL

metodi96:
@g1nt0ki it seems like this commit: https://github.com/DefiLlama/DefiLlama-Adapters/commit/fdf5ff305d5a24ddba7dd549c4b2f6d83dba2060 has broken the TVL calculation of the leverage positions on Nolus. It is working fine with the original code so I pushed it back. You had removed this transformBalances function entirely so the tvl(protocols) function was not returning a thing, hence the 0 TVL for the positions.

### DIFF
--- a/projects/nolus/index.js
+++ b/projects/nolus/index.js
@@ -1,4 +1,5 @@
 const sdk = require('@defillama/sdk')
+const { transformBalances } = require('../helper/portedTokens')
 const { queryContract, queryManyContracts, queryContracts } = require('../helper/chain/cosmos')
 const { sleep } = require('../helper/utils')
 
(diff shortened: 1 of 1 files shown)


llamabutler:
        
 
 
    osmosis                   1.39 M
    nolus                     318.37 k
    neutron                   16.23 k
    
    total                    1.73 M

metodi96:
Hmm, when I run this locally, I get TVL of 2.44M. I did a fresh `npm i`. 
Any idea why there's such a mismatch? The 2.44M is closer to the truth (it's close to the value we calculate with our own indexer - 2.5M, where the difference is that allBTC and allSOL positions' TVL isn't returned.

PR #13935, opened 2025-03-19
Fix: Deflex

0xpeluche:
use .toString() to prevent `Error in algorand: Error: token must be a string`


llamabutler:
The adapter at projects/deflex exports TVL: 
        
 
 
    algorand                  38
    
    total                    38.455632

PR #13973, opened 2025-03-21
added new chains in tvl and giveaway factory contracts in tvl for cha…

MaryamNaveedFunavry:
This PR introduces giveaway factory contracts and adds support for some new chains(bera, sonic and blast) in the TVL (Total Value Locked) calculation of our project(chainGPT) on Defillama.

### DIFF
--- a/projects/chaingpt/index.js
+++ b/projects/chaingpt/index.js
@@ -44,11 +44,65 @@ const config = {
   ],
   era: [
     { factory: '0xf25F7c9522cdCD839697F1644CFCA1312306885C', fromBlock: 37458600 },
+  ],
+  blast: [
+    { factory: '0xBB9667bcbd04bD4F2C1f2FCd14995Bd3d00c5655', fromBlock: 8619958 },
+  ],
+  sonic: [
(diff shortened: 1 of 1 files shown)


waynebruce0x:
Where can the giveawap pools be seen on your UI?

ceoguy:
> Where can the giveawap pools be seen on your UI?

They can be seen here: 
https://pad.chaingpt.org/giveaway

projects/chaingpt/index.js
```diff
@@ -44,11 +44,65 @@ const config = {
...
+  sonic: [
+    { factory: '0x9840652DC04fb9db2C43853633f0F62BE6f00f98', fromBlock: 11999667 },
+  ],
+  berachain: [
+    { factory: '0xBB9667bcbd04bD4F2C1f2FCd14995Bd3d00c5655', fromBlock: 787656 },
+  ],
+
+}
+
+const giveawayConfigs = {
```
waynebruce0x, line 60: TVL coming from giveaways should be exported under key 'vesting' rather that 'tvl' because the tokens are non-circulating

g1nt0ki (changes requested):
sorry, no. these are not tokens locked by users, we cant count it as tvl

PR #14003, opened 2025-03-24
Add StableJack YT/PT-scUSD Sonic adapter

MuhammedAkinci:
Note for DeFi Llama Team: 
This initial PR adds support for **YT-scUSD** and **PT-scUSD** tokens on the **Sonic** blockchain as part of our testing process. Once this adapter implementation is approved and merged by the DeFi Llama team, we will follow up with additional incremental PRs to support all remaining tokens listed in the table below:

| Chain     | Tokens to be added                 |
|-----------|------------------------------------|
| Sonic     | YT/PT-stS, YT/PT-wOS               |
| Avalanche | YT/PT-rsAVAX, YT/PT-savUSD         |

We will also ensure compliance with all points listed under **"Important Notes"** above in each incremental PR.


projects/stablejack-scusd-sonic/index.js
```diff
@@ -0,0 +1,24 @@
+const YT_scUSD = "0x38C6Dd3844Cc9B6276468000f7b06e8D53e5D1C0";
+const PT_scUSD = "0xeAe9Cbe1277fb3DD40576db8A7B0c6bDFEB58AcA";
```
waynebruce0x, line 2: these addresses show no contracts on sonicscan

llamabutler:
The adapter at projects/stablejack-scusd-sonic exports TVL: 
        
 
 
    sonic                     0
    
    total                    0

projects/stablejack-scusd-sonic/index.js
MuhammedAkinci: our YT-scUSD is a yield bearing asset which accumulates aTKNs. 

aTKN is a derivative of a scUSD. 

This is similar to our v1 logic, where we had aUSD (first derivative of sAVAX) and xAVAX (second derivative). The only difference now is that aTKN is directly locked to an ERC4626 contract (YT) and YT is now rebasing

waynebruce0x: I see - how is the value of aTKN related to the value of scUSD? We need a way of pricing aTKN using on chain function calls so that we can value the YT and PT

MuhammedAkinci: aTKN is 1$, always

waynebruce0x: Which protocol mints this?

MuhammedAkinci: This is one of our assets at Stable Jack. Users bring their scUSD tokens and deposit them to us. In exchange, they receive PT/YT tokens

waynebruce0x: In which case please count the scUSD balance rather than PT/YT

llamabutler:
The adapter at projects/stablejack-scusd exports TVL: 
        
 
 
    sonic                     538.50 k
    
    total                    538.51 k

PR #14056, opened 2025-03-27
Fix: Ttswap

0xpeluche:
> Fix, even if there are no liquidities


llamabutler:
The adapter at projects/ttswap exports TVL: 
        
 
 
    thundercore               40.77 k
    
    total                    40.77 k

PR #14082, opened 2025-03-28
LHYPE underlying token endpoint update

DashanMcCain:
### Changes
- Added LHYPE as a token held by the LHYPE vault (we've implemented a buyback mechanism on dexs to combat de-pegs)
- Modified the endpoint to fetch underlying strategies


DashanMcCain:
One note is that it seems to not be counting our collateral in HypurrFi still. I mentioned this in my initial PR here:
https://github.com/DefiLlama/DefiLlama-Adapters/pull/13735#issuecomment-2701672715

And was told it was fixed here
https://github.com/DefiLlama/DefiLlama-Adapters/pull/13735#issuecomment-2703542590


Which did add support for the debt token, but the collateral token `hywstHYPE` is not being counted towards the TVL

waynebruce0x:
Thanks for flagging this, pricing should be back in the next hour

PR #14118, opened 2025-03-31
feat: add treasuries for FolksFinance

moshenskyDV:
Add treasuries for FolksFinance.

One issue there to ask.
WETH (`887406851`) from Algorand isn't reflected in test results. So the total treasury amount is $36k lower than expected. Could you give a hint, why this token isn't visible?


g1nt0ki (approved):
looks good, thanks

moshenskyDV:
@g1nt0ki, what about WETH (`887406851`)? (Any idea, why it isn't counted?)

image

PR #14143, opened 2025-04-01
LHYPE (HyperEVM) Hyperlend Valuation Request

DashanMcCain:
Hi, thanks for fixing the previous issue with the collateral in HypurrFi. Now we deposited into Hyperlend and our collateral and debt tokens are not being tracked. For reference, here are the respective token addresses `0x0Ab8AAE3335Ed4B373A33D9023b6A6585b149D33` and `0x747d0d4Ba0a2083651513cd008deb95075683e82`. Where the collateral token (`0x0Ab...`) is valued the same as the collateral in HypurrFi and debt is just in `WHYPE`.  This is an aave v3 fork so the valuation for these assets are the same as I outlined here https://github.com/DefiLlama/DefiLlama-Adapters/pull/13735#issuecomment-2701672715. 

For future reference, should I open up an issue or another PR next time?

### DIFF
--- a/projects/looped-hype/index.js
+++ b/projects/looped-hype/index.js
@@ -12,11 +12,11 @@ const LHYPE_VAULT_ADDRESS = ['0x5748ae796AE46A4F1348a1693de4b50560485562'];
 const tvl = async () => {
   const strategies = await getConfig(
     'lhype-tokens',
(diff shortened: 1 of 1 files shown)


llamabutler:
        
 
 
    hyperliquid               8.51 M
    
    total                    8.51 M

waynebruce0x:
we've added support for these assets, does this second result look better? For missing assets its best to message our discord if possible, or add me on telegram

waynebruce0x:
waynebruce0x

PR #14178, opened 2025-04-03
add corecoins mtt network

zhaolee:


waynebruce0x:
This PR makes no change to anything

zhaolee:
> This PR makes no change to anything

Hello, we want to add USDT to the mtt network because our YaakoSwap trading pair includes this token, which should affect the calculation of tvl

PR #14267, opened 2025-04-10
feat: add lista lending adaptor

tyler-tsai:
Add a new lending protocol into listaDao
https://defillama.com/protocol/lista-dao#information

Lista Lending is a fully decentralized and permissionless P2P lending protocol crafted for BNB Chain, breaking free from the constraints of traditional large-pool lending to cultivate a more inclusive and resilient ecosystem.
Lista Lending’s core revolves around a vault-based system, pooling liquidity and dynamically allocating it across different lending & collateral pairs, which we call markets, based on supply and demand.

image

PR #14319, opened 2025-04-15
Feat/lista lending

tyler-tsai:
supplement for pr 14267
https://github.com/DefiLlama/DefiLlama-Adapters/pull/14267

### DIFF
--- a/projects/lista-lending/index.js
+++ b/projects/lista-lending/index.js
@@ -1,0 +1,113 @@
+const axios = require('axios')
+const { getLogs2 } = require('../helper/cache/getLogs')
+
+const config = {
+  bsc: {
+    vault: '0x8F73b65B4caAf64FBA2aF91cC5D4a2A1318E5D8C',
+    fromBlock: 48000000,
+  }
+}
+const abi = {
+    totalAssets: "uint256:totalAssets",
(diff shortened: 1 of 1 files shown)


tyler-tsai:
@waynebruce0x The 451 error occurs because the GitHub runner uses a US-based IP address, and our API doesn't allow access from the US.
It works fine on your server since it's using a different IP.

waynebruce0x:
are collateral and deposits seperate?

waynebruce0x:
your UI seems to indicate that TVL would be more like 130M$

tyler-tsai:
> your UI seems to indicate that TVL would be more like 130M$

Yeah, they are seperate, total deposits is not equal TVL.

The formula is total deposits + total collateral - total borrowed.

I also add total borrow as option.

tyler-tsai:
Morpho’s TVL (Total Value Locked) can be estimated as Total Vault + Total Collateral - Total Borrowed.
•  Total Vault: Assets deposited in Morpho Vaults for lending.
•  Total Collateral: Assets locked as collateral in markets to support borrowing.
•  - Total Borrowed: Assets borrowed from markets, subtracted to avoid double-counting.
This formula captures the net value of assets locked in the protocol, reflecting deposits and collateral while excluding borrowed funds.

tyler-tsai:
@waynebruce0x We have updated a version fetching TVL and borrowed info on chain.

PR #14325, opened 2025-04-15
add Hyperbeat TVL

HugoHL1:
---

### DIFF
--- a/projects/hyperbeat/index.js
+++ b/projects/hyperbeat/index.js
@@ -1,0 +1,52 @@
+const VAULTS = {
+  hyperliquid: {
+    hbHYPE: {
+      vault: '0x96C6cBB6251Ee1c257b2162ca0f39AA5Fa44B1FB', // hbHYPE
+      underlying: '0x5555555555555555555555555555555555555555', // WHYPE
+    }
+  },
+  ethereum: {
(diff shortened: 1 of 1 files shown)


projects/hyperbeat/index.js
```diff
@@ -0,0 +1,52 @@
...
+    }
+  }
+};
+
+async function calculateVaultTvl(api, vaultConfigs) {
+  const vaultAddresses = vaultConfigs.map(c => c.vault);
+  const underlyingTokens = vaultConfigs.map(c => c.underlying);
+
+  const totalSupplies = await api.multiCall({
+    abi: 'erc20:totalSupply',
```
waynebruce0x, line 29: can we use erc20:balanceOf here instead please?

HugoHL1: yes sure

HugoHL1 (commented):
perfect, thanks

projects/hyperbeat/index.js
waynebruce0x: should I make this change? which tokens and holders should we use?

HugoHL1:
Hey @waynebruce0x thanks for the tips, I push an update, I use erc20:balanceOf now :)

PR #14328, opened 2025-04-15
Fix: Vectis

0xpeluche:
Added a filter to prevent `null` responses from `const accounts = await getMultipleAccounts(vaultUserAddresses)` from entering the `deserializeUserPositions` logic, which would otherwise throw an error


llamabutler:
        
 
 
    solana                    14.06 M
    
    total                    14.06 M

PR #14335, opened 2025-04-15
Address update for the new token used in the protocol

ironrai:
We launched the new version of the token and users are already migrating to the new address.


llamabutler:
The adapter at projects/primestaking-xyz exports TVL: 
        
 
 
    xdc                       1.01 M
    
    total                    1.01 M

PR #14347, opened 2025-04-16
feat: update sandglass tokens and owners list

metalgear-dev:
I updated the tokens and owners list for sandglass. Please check.

### DIFF
--- a/projects/sandglass/index.js
+++ b/projects/sandglass/index.js
@@ -1,27 +1,27 @@
 const { sumTokens2, } = require('../helper/solana')
 
-const tokenAccounts = {
+const tokensAndOwners = {
   eclipse: [
-    "3983hQGtV5FqqQrsVm6Mhw1Tn4BLuYYSaaSyQLRg4V5A"
+    ["A8uPGauLyDTw9dMjBpb9Vrgq7frbWX46XqX71paW4pri", "GGgv3ixpQjpuHL7GeLedEHHVSq6BMKmRk5RckTwvmgka"]
   ],
   solana: [
-    "CSHiqfGfP1ch2YmWoyjB2tNbdnuUw3o1WdgDaicgRdmM",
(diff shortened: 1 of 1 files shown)


projects/sandglass/index.js
```diff
@@ -1,27 +1,27 @@
...
-    "icYHCTuFtkcRPpY4mxXXLJbZpsLMo9qL4xdySJafJCr",
-    "pvVG7N97GBeHrY29QbUU8hVtbwdVJUTsU2zAFPredeV",
-    "Fd788Rcsaj7Rda3rg8q63cSeTs3mesd7idqV64QugFGY",
-    "AxmFbfzAHfzR7LSuBkKDUyEQnGm68HqdxGRkQJ2e4y63",
-    "4ySzN7U4CNhp9WkpVexh1G1zPs5gmnejFVURP8UF5WQA",
-    "EbWHbVL3SotgqGoj1wPEM7VSkqKSY2pUHtSNMe83HF2c",
-    "RrpDctr3qkcJjfh6eM9PA5jhH1ZndDhtKBEvQ5jwM2p",
-    "DazRVQErAofKnSS9L4vuR5MFLkAMHCF9BZiKTEQcKduC",
-    "66i7W1q7Z4sQRuCbNPhGpmBkeiDj3zunEF3Z1GxLqc21"
+    ["EAsRTTRK2wjte4DecPYUjieTkakqFEYe9WD8z2mqvwS8", "GeXibGhHiCPvtJKCGEKWHnHvGnk1tT3VfzYz8s5G3v4y"],
```
waynebruce0x, line 8: please count the underlying Kamino cUSDC instead. Follow same logic for other markets.

metalgear-dev: The first one is YT mint, and the second one is the owner. It has two token accounts. So we updated like that.
Is it not allowed to set the list of mints and owners?

waynebruce0x: I understand now, sorry my mistake

PR #14351, opened 2025-04-16
Fix: Bitflow

0xpeluche:
> Bitflow fix: added a sleep to avoid rate limiting


llamabutler:
The adapter at projects/bitflow exports TVL: 
        
 
 
    stacks                    10.33 M
    
    total                    10.33 M

PR #14428, opened 2025-04-23
Remove self referential Usual TVL

MickdeGraaf:
Fixes self referential TVL being double counted.

USL (Usual stability loans) are deposits on Euler of unbacked USD0 which can be borrowed using USD0++ which leads to doublecounting of TVL


llamabutler:
The adapter at projects/usual exports TVL: 
        
 
 
    ethereum                  462.95 M
    
    total                    462.95 M

PR #14433, opened 2025-04-24
Veda: Fail gracefully on query revert

philipjames44:
Veda would like to backfill historical data and had an issue backfilled any earlier than Jan 20 2025. After some testing it became apparent certain contract queries revert for various reasons historically. This change allows us to fail gracefully such that the veda protocol TVL can be backfilled correctly.

### DIFF
--- a/projects/veda/helper_methods.js
+++ b/projects/veda/helper_methods.js
@@ -1,108 +1,167 @@
 async function sumLegacyTvl({ vaults, api, ownersToDedupe = [] }) {
-  const [assets, bals] = await Promise.all([
-    api.multiCall({
-      abi: "address:asset",
-      calls: vaults,
-    }),
-    api.multiCall({
-      abi: "uint256:totalAssets",
-      calls: vaults,
-    })
-  ]);
+  try {
+    const [assets, bals] = await Promise.all([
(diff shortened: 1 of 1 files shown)


waynebruce0x:
These try catches would mean our monitoring will not flag when Veda has drops in TVL which will make debugging impossible. Also when backfilling we wont know if the data is accurate until its written to production

llamabutler:
The adapter at projects/veda exports TVL: 
        
 
 
    ethereum                  1.63 B
    berachain                 555.68 M
    sonic                     115.89 M
    base                      5.44 M
    bob                       3.05 M
    arbitrum                  2.57 M
    bsc                       2.35 M
    
    total                    2.32 B

philipjames44:
@waynebruce0x you're right thanks for pointing that out. I removed the try catches and just left the better share handling. let me know if this works

waynebruce0x:
allowing failed RPC calls can also result in inaccurate data - is this specifically for backfilling? We could run your code in CLI to backfill and then keep the current implementation for future data?

waynebruce0x:
Hey I've just got around to running this and I'm running into some issues. Is there an easier way I can contact you? Discord or telegram?

PR #14479, opened 2025-04-29
Update wand protocols

nonolooklook:


llamabutler:
The adapter at projects/wand exports TVL: 
        
 
 
    blast                     736.02 k
    sty                       245.02 k
    
    total                    981.04 k

PR #14520, opened 2025-05-02
Add Beracana Protocol TVL on Berachain

prasenjit-26:

### DIFF
--- a/projects/beracana/index.js
+++ b/projects/beracana/index.js
@@ -1,0 +1,30 @@
+const { sumTokens2 } = require('../helper/unwrapLPs');
+const defaultAddresses = require('./berachain-contracts.json')
+
+async function getContracts() {
+    const contracts = await fetch('https://api.beracana.com/beracana/getContracts');
+    const resp = await contracts.json();
+    if (resp.success && resp.data && resp.data.banks && resp.data.vaults){
+      return { banks: resp.data.banks, vaults: resp.data.vaults }
+    } else {
+      return { banks: defaultAddresses.banks, vaults: defaultAddresses.vaults }
+    }
+}
+async function getTvl(api) {
+    const { banks, vaults } = await getContracts();
+    const tokens = banks.map(i => i.token)
+    const bankAddresses = banks.map(i => i.bank)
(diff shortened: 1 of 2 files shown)


llamabutler:
The adapter at projects/beracana exports TVL: 
        
 
 
    berachain                 16.37 k
    
    total                    16.37 k

projects/beracana/index.js
```diff
@@ -0,0 +1,30 @@
...
+      return { banks: resp.data.banks, vaults: resp.data.vaults }
+    } else {
+      return { banks: defaultAddresses.banks, vaults: defaultAddresses.vaults }
+    }
+}
+async function getTvl(api) {
+    const { banks, vaults } = await getContracts();
+    const tokens = banks.map(i => i.token)
+    const bankAddresses = banks.map(i => i.bank)
+    const currentFunds = await api.multiCall({  abi: 'function currentFunds() view returns (uint256)', calls: bankAddresses })
```
waynebruce0x, line 17: can we use erc20:balanceOf rather than currentFunds() and totalFunds() ?

prasenjit-26: Hey we can't use erc20:balanceOf directly because 

currentFunds() returns the amount of funds available to borrow in the bank, while totalFunds() returns the total amount the bank controls: both the available funds plus the amount already lent out as debt.

Since funds are distributed across different vaults and not just held in the contract itself, there's no single balanceOf function that can accurately reflect the true total value locked (TVL). That’s why we need these custom functions—to get the full picture across all assets, not just what’s sitting in the contract.

waynebruce0x: please export currentFunds() as `tvl` and the difference between totalFunds() and currentFunds() as `borrowed`

waynebruce0x: do the tokens sit in the vault contracts? what do the vaults do?

waynebruce0x: okay lets try counting the vault assets instead

prasenjit-26: Hey , sorry for the delay , pushed the changes let me know if this works .
Thanks

PR #14552, opened 2025-05-06
Fix: Beraborrow

0xpeluche:
Fixed the adapter, `address:collVault` method is no longer available across all vaults and consistently throws errors:

image

→ Commented out the call and updated the final blacklist accordingly


llamabutler:
The adapter at projects/beraborrow exports TVL: 
        
 
 
    berachain                 417.42 M
    
    total                    417.42 M

PR #14562, opened 2025-05-07
Smardex: Liquidation adapter for leveraged positions on USDN

estebgonza:
Here is our liquidation adapter for the decentralized perpetual protocol USDN by Smardex.

Few links:
https://smardex.io/usdn/long
https://smardex.io/analytics

USDN has leveraged positions that are liquidable.

As discussed on Discord, this adapter is a version based on our internal API, we will migrate this adapter when we have subgraphs on the protocol.

**API Key**
Would it be possible to provide you with an API key privately on Discord to access our API? I'm currently using the `SMARDEX_USDN_API_KEY` environment variable; let me know if that doesn't work.
This is an API used only for Defillama so we would like to not expose it publicly.

We've already provided an API key to access the subgraphs of the DEX portion of Smardex in the past, so I assume this is possible for our USDN API 

Thank you !

### DIFF
--- a/liquidations/smardex/index.ts
+++ b/liquidations/smardex/index.ts
@@ -1,0 +1,118 @@
+import { ethers } from "ethers";
+import { Liq } from "../utils/types";
+import axios from "axios";
+interface PositionHistory {
+  transactionHash: string;
+}
+
+interface PositionMainOwner {
+  address: string;
+}
+
+interface PositionMain {
+  lastOwner: PositionMainOwner;
+}
+
+interface Position {
+  tick: number;
+  tickVersion: string;
+  index: string;
+  validator: string;
+  recipient: string;
+  amount: string;
+  startPrice: string;
+  liquidationPenalty: number;
+  totalExpo: string;
+  amountReceived: string;
+  amountRemaining: string;
+  profit: string;
+  liquidationPrice: string | null;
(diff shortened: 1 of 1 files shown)


waynebruce0x:
Any chance we can use the same API key as the subgraph access?

asyx21:
> Any chance we can use the same API key as the subgraph access?

The key is not the same at the moment. Would that be more conviniant to use the same ?

asyx21:
> > Any chance we can use the same API key as the subgraph access?
> 
> The key is not the same at the moment. Would that be more conviniant to use the same ?

we'll change to use the same as for subgraphs ;)

estebgonza:
> Any chance we can use the same API key as the subgraph access?

Done, we replaced the env var by `SMARDEX_SUBGRAPH_API_KEY`

liquidations/smardex/index.ts
```diff
@@ -0,0 +1,118 @@
...
+    const response = await axios.get<ApiResponse>(API_URL, {
+      headers: { "x-api-key": API_KEY },
+    });
+    
+    if (response.data && response.data.positions && Array.isArray(response.data.positions)) {
+      return response.data.positions;
+    } else {
+      return [];
+    }
+  } catch (error) {
```
waynebruce0x, line 62: please remove any try/catch

PR #14621, opened 2025-05-13
feat: Add logic for Harvest Autopilot TVL

CryptJS13:
We have recently launched a new product called Autopilot. Their interface is slightly different from our original vaults, so I had to make an alteration to include their TVL in the calculation.

### DIFF
--- a/projects/harvest.js
+++ b/projects/harvest.js
@@ -31,6 +31,14 @@ const tvl = async (api) => {
     if (token) api.add(token, bals2[i])
   })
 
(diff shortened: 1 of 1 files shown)


llamabutler:
The adapter at projects/harvest.js exports TVL: 
        
 
 
    base                      23.52 M
    ethereum                  5.56 M
    arbitrum                  1.20 M
    era                       206.53 k
    polygon                   79.70 k
    
    total                    30.57 M

PR #14683, opened 2025-05-16
Fix: LiquidOps borrow adapter

LorimerJenkins:
- add extra await to not overload AO node
- return the borrow balances correctly


llamabutler:
The adapter at projects/LiquidOps exports TVL: 
        
 
 
    ao                        61.14 k
    ao-borrowed               19.43 k
    borrowed                  19.43 k
    
    total                    61.14 k

PR #14699, opened 2025-05-19
add wsteth & wsteth / arb gamma vault to tvl checks

ewansheldon:
adds wsteth to tvl check, as well as wsteth / arb gamma hypervisor vault


llamabutler:
The adapter at projects/standard-smart-vaults exports TVL: 
        
 
 
    arbitrum                  1.56 M
    
    total                    1.56 M

PR #14712, opened 2025-05-20
Mellow update vaults

0xsign:
add new Lisk vaults


llamabutler:
        
 
 
    ethereum                  336.51 M
    lisk                      372
    
    total                    336.51 M

PR #14713, opened 2025-05-20
Haven1

TSxo:
Adds support for Staking TVL and vesting tokens on Haven1 (`projects/haven1/index.js`).


llamabutler:
The adapter at projects/haven1 exports TVL: 
        
 
 
    haven1-vesting            44.94 M
    vesting                   44.94 M
    haven1                    389.00 k
    haven1-staking            269.57 k
    staking                   269.57 k
    
    total                    388.99 k

projects/haven1/index.js
```diff
@@ -1,6 +1,107 @@
...
+};
+
+const STAKING = {
+    flex: "0x708E6dd0452D2C245e7d461c6c8B70F587ca3167",
+    ve: "0xac488B7E18Cef83aC300E9ffD9324BAa5BB62a13",
+};
+
+const STAKING_CONFIG = [
+    { abi: ABI.totalSupply, target: STAKING.flex },
+    { abi: ABI.supply, target: STAKING.ve },
```
waynebruce0x, line 47: why are we using supply/totalSupply here instead of counting token balances with erc20:balanceOf?

waynebruce0x: please use sumTokens2() or similar, with a list of H1 varients

TSxo: Because one of the variants, esH1, is not transferable to contracts other than these staking contracts, it does not have a "price", so to speak (as mentioned in the file comments) - so the most transparent way to display this is via the `supply` and `totalSupply` calls, and price it as H1. 

I can modify the logic to instead count all H1 currently in escrow, but you would losing accuracy.

waynebruce0x: How does esH1 get there in the first place?

TSxo: Thank you for the follow up! Admins have the ability to whitelist certain addresses. The staking contracts have been added to this whitelist to allow user's to stake and earn against their esH1.

PR #14766, opened 2025-05-23
Fix Sonic-ICP

0xpeluche:
Adapter updated: removed the logic fetching data from v1, which has been inaccessible for months. The data is no longer available via GraphQL and there has been no liquidity since the transition to v3. Minor change on v3, which has been using an ascending order by clause for the past few weeks


llamabutler:
The adapter at projects/sonicooo exports TVL: 
        
 
 
    icp                       4.63 M
    
    total                    4.63 M

PR #14770, opened 2025-05-23
Fix: Zunami

0xpeluche:
Fixed the Zunami protocol: the `tokenHoldings` method has been returning all zeros for some time (likely due to a contract upgrade). Now using the strategies methods instead to retrieve and sum the strategy balances


llamabutler:
The adapter at projects/zunami exports TVL: 
        
 
 
    ethereum                  569.78 k
    ethereum-staking          8.01 k
    staking                   8.01 k
    
    total                    569.78 k

PR #14776, opened 2025-05-25
fix: WBTC supply on ethereum chain

noateden:


llamabutler:
The adapter at projects/wbtc.js exports TVL: 
        
 
 
    bitcoin                   13.85 B
    
    total                    13.85 B

PR #14780, opened 2025-05-26
add polygon tvl

tanpx12:
Vault deployed on Polygon, so we add Polygon to the TVL calculation


llamabutler:
The adapter at projects/untangled-yield exports TVL: 
        
 
 
    celo                      240.86 k
    polygon                   5.31 k
    
    total                    246.16 k

tanpx12:
Hi, I would like to know why the test_change fail here.

PR #14835, opened 2025-05-28
feat: fix cache misses for Panoptic and add Unichain v1.1

dyedm1:
This PR resolves discrepancies in the displayed Panoptic TVL wherein TVL would only spike periodically to the correct value when the cache was dropped and all the data from subgraph queries was successfully retrieved:

image

The cause of this issue seems to be that the same query key is reused for different chains and protocol versions, despite the fact the queries come from multiple different subgraphs. To resolve this and discriminate those subgraph queries, the protocol version and chain name have been added to the query cache key.

Also, support for Panoptic v1.1 on Unichain has been added.

### DIFF
--- a/projects/panoptic-v1_1/index.js
+++ b/projects/panoptic-v1_1/index.js
@@ -38,12 +38,12 @@ const config = {
     safeBlockLimit: 50,
     PoolManager: '0x498581ff718922c3f8e6a244956af099b2652b2b'
   },
-//   unichain: {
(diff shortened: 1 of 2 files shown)


llamabutler:
The adapter at projects/panoptic-v1_1 exports TVL: 
        
 
 
    base                      246.43 k
    ethereum                  0
    unichain                  0
    
    total                    246.43 k

llamabutler:
The adapter at projects/panoptic exports TVL: 
        
 
 
    ethereum                  2.15 M
    base                      112.70 k
    unichain                  37.47 k
    
    total                    2.30 M

PR #14879, opened 2025-06-02
Fix Moar TVL calculation

lumos42:
Moar recently started supporting leveraged CLMM positions. The TVL on defillama started showing incorrect numbers after this because the existing calculation didn't account for the value of CLMM position. This change fixes that issue.


ghost:
The adapter at projects/moar exports TVL: 
        
 
 
    aptos                     1.47 M
    aptos-borrowed            435.34 k
    borrowed                  435.34 k
    

PR #14893, opened 2025-06-02
feat(mimswap): add support for nibiru

0xmDreamy:
The pool for Nibiru has been hardcoded for now as there's currently only one pool, and because getLogs doesn't work on Nibiru. I suspect it attempts to fetch block 1, but it doesn't exist.

### DIFF
--- a/projects/mimswap/index.js
+++ b/projects/mimswap/index.js
@@ -2,17 +2,27 @@ const { getLogs2 } = require('../helper/cache/getLogs')
 
 const config = {
   arbitrum: { factory: '0x8D0Cd3eEf1794F59F2B3a664Ef07fCAD401FEc73', fromBlock: 205217727 },
-  ethereum: { factory: '0xDF46F6b1a5F794F21eaD4008C7De4E02Dc045297', fromBlock: 20537337 },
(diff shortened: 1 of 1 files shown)


ghost:
The adapter at projects/mimswap exports TVL: 
        
 
 
    arbitrum                  15.74 M
    kava                      2.73 M
    blast                     213.30 k
    nibiru                    129.79 k
    ethereum                  100.54 k
    
    total                    18.92 M

Unique-Divine:
@noateden, I looked into this more closely and found that the error in `eth_getLogs` is related to node pruning settings. When I manually override Nibiru's RPC to be an archive node endpoint, all of the errors go away. Can you advise on what needs to be updated to have DeFiLlama reference an archive node for requests that look this far back into the history? 

I tried adding `NIBIRU_ARCHIVAL_RPC` to the `env.js` file, but the `api` object doesn't use the archive node even when the `api.provider.archivalRPCs` field is populated. I found this out with an ugly hack
```js
if (api.chain === "nibiru") {
  api.provider.rpcs = api.provider.archivalRPCs
}
```

noateden:
hi @0xmDreamy @Unique-Divine, thanks you for checking with issue of nibiru rpc. We have already fix it and make update in this PR: https://github.com/DefiLlama/DefiLlama-Adapters/pull/14940

PR #14952, opened 2025-06-06
Feat: add new WBTC vault 

nguyenquangthang1997:
- add new WBTC vault at https://app.b14g.xyz/vaults/wbtc


ghost:
The adapter at projects/b14g exports TVL: 
        
 
 
    bitcoin                   183.44 M
    core                      18.77 M
    
    total                    202.21 M

projects/b14g/index.js
```diff
@@ -1,31 +1,39 @@
...
+            const receiverRewards = logs.map(i => i.rewardReceiver)
+            const coreStakeAmounts = await api.multiCall({abi: 'uint256:totalCoreStake', calls: receiverRewards})
+            const totalCoreStakedVault = await api.call({
+                target: "0xee21ab613d30330823D35Cf91A84cE964808B83F",
+                abi: 'uint256:totalStaked',
+            });
+
 
+            const totalCoreStakedWBTCVault = await api.call({
+                target: "0x2e3ea6cf100632a4a4b34f26681a6f50347775c9",
```
noateden, line 23: this is cWBTC token on core, but why are you adding WBTC `0x5832f53d147b3d6cd4578b9cbd62425c7ea9d0bd` on core to tvl?

mas0610: Ah, in our wbtc vault, users deposit wbtc on this, and vault will supply on colend (to mint cWBTC with rate 1:1). Here is an deposit example: https://scan.coredao.org/tx/0x2d6b2e72fb93c862df08941623696246e6ad12012a9967264f69bf0ce3ec1449

noateden: I don't think it would work here, the WBTC:cWBTC ratio is not 1:1
you should consider to calculate correct amount of WBTC token here, don't get cWBTC balance.

mas0610: Sorry but I'm a bit confused about why the WBTC to cWBTC ratio isn’t 1:1. When I supply 1 WBTC on Colend, I receive 1 cWBTC. And when I withdraw 1 cWBTC, I get back 1 WBTC. So why would the ratio ever change?

PR #14954, opened 2025-06-06
yala, add the tvl of bfBTC stored in the AssetWrapper

yala-hub:
add the tvl of bfBTC stored in the AssetWrapper, so that users can collateralize bfBTC to borrow yu and obtain the related yield of yu


ghost:
The adapter at projects/yala exports TVL: 
        
 
 
    bitcoin                   31.06 M
    ethereum                  2.51 M
    
    total                    33.58 M

projects/yala/index.js
```diff
@@ -32,6 +32,31 @@ async function tvlInEthereum(api) {
...
+      (await api.call({
+        abi: "uint256:latestAnswer",
+        target: res.ethereum.AssetWrapper.bfBTC.oracle,
+      })) /
+      10 ** res.ethereum.AssetWrapper.bfBTC.decimalsOracle;
+
+    // convert to BTC
+    amt *= price;
+
+    api.addCGToken(
```
waynebruce0x: we track the price of bfBTC already so there should be no need for this conversion

ghost:
The adapter at projects/yala exports TVL: 
        
 
 
    bitcoin                   31.25 M
    ethereum                  2.52 M
    
    total                    33.76 M

ghost:
The adapter at projects/yala exports TVL: 
        
 
 
    bitcoin                   31.25 M
    ethereum                  2.52 M
    
    total                    33.76 M

projects/yala/index.js
yala-hub: got it, i have updated this.

projects/yala/index.js
```diff
@@ -32,6 +32,17 @@ async function tvlInEthereum(api) {
       permitFailure: true,
     });
   }
+
+  // bfBTC in AssetWrapper
+  {
+    // 1:1 bfBTC amt
+    let amt = await api.call({
+      abi: "erc20:totalSupply",
```
waynebruce0x: is there a reason why we cant count a balance of bfBTC here?

yala-hub: It will be on next week.

yala-hub: There are no user deposits yet.

waynebruce0x: why are we using vault.totalSupply instead of underlying.balanceOf ?

yala-hub: In fact, it's the same. Users will store bfBTC in AssetWrapper, and then deposit the wrapper as collateral into cdp. Using totalSupply is a more intuitive way.

waynebruce0x: Can we change to use underlying.balanceOf instead please

yala-hub: In fact, when deploying AssetWrapper, the underlying has already been set as immutable. The bfBTC that users deposited is indeed our TVL. This is the contract of AssetWrapper. Please take a look.

I have already invited you to read this GitHub project. <https://github.com/yalaorg/asset-wrapper/blob/b1b41adad46ffa1c1f20172fa1f8328ccf78dd72/contracts/AssetWrapper.sol#L16>

waynebruce0x: The reason we prefer to use balanceOf over totalSupply is that in the case of an exploit, totalSupply may not be an accurate representation of TVL. Is there a reason why we can't use balanceOf?

yala-hub: If we use the balanceOf, the timing for the TVL is a bit delayed. Our partners would like to be able to see the increase in TVL as soon as they deposit BFBTC.

ghost:
The adapter at projects/yala exports TVL: 
        
 
 
    bitcoin                   32.98 M
    ethereum                  2.60 M
    
    total                    35.58 M

ghost:
The adapter at projects/yala exports TVL: 
        
 
 
    bitcoin                   32.98 M
    ethereum                  2.60 M
    
    total                    35.58 M

projects/yala/index.js
yala-hub: It has been updated. Previously, I misunderstood.

PR #14977, opened 2025-06-08
Fixed a bug and added more pools for dre finance

drecronye:
There was a line that referred to `tokens.sonic.USDC_e` instead of `ADDRESSES.sonic.USDC_e`. 

And also we added scUSD (Sonic USD) as collateral and the scUSD/DRE LP pool into pool2.


ghost:
The adapter at projects/drefinance exports TVL: 
        
 
 
    sonic                     64.88 k
    sonic-pool2               62.20 k
    pool2                     62.20 k
    sonic-staking             28.96 k
    staking                   28.96 k
    
    total                    64.88 k

PR #14999, opened 2025-06-10
RealtyX add plume_main tvl support

HomuraAcc:
Because of the launch of the plume mainnet, we also updated the adapter on the plume mainnet. Now we merge the tvl on plume into the original project.


ghost:
The adapter at projects/realtyx exports TVL: 
        
 
 
    base                      881.62 k
    plume_mainnet             471.45 k
    
    total                    1.35 M

PR #15053, opened 2025-06-13
civitia

waynebruce0x:
Initia app chain - not sure if chain should be civitia or initiia?

https://x.com/civitiaorg


ghost:
The adapter at projects/civitia exports TVL: 
        
 
 
    civitia                   268.29 k
    
    total                    268.29 k

PR #15059, opened 2025-06-13
solana: kamino maple pools

treeoflife2:
Maple Markets
https://app.kamino.finance/lending


treeoflife2:
can't count kamino market as maple TVL, as liquidity resides in Kamino’s vaults, not directly in Maple’s contracts.

PR #15120, opened 2025-06-17
Fix: Sommelier

0xpeluche:
The `totalAssets` method is throwing an error. It seems that the vault's oracle (`0x7AcDB8096E51b2730387977Bad340B9EFDE61342`) is no longer able to return the asset price via the `getLatest` method, which now indicates `notSafeToUse` as `true`

I'm commenting out the vault for now to prevent a potential break in the code. Also, commenting out the vault doesn't seem to impact the TVL (no drop observed)

### DIFF
--- a/projects/sommelier/cellar-constants.js
+++ b/projects/sommelier/cellar-constants.js
@@ -44,7 +44,7 @@ const cellarsV2 = [
 ];
 
 // v2.5 Cellars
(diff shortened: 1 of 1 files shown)


ghost:
The adapter at projects/sommelier exports TVL: 
        
 
 
    ethereum                  12.47 M
    optimism                  1.02 M
    arbitrum                  816.69 k
    
    total                    14.30 M

PR #15194, opened 2025-06-23
Updating contracts for TVL to include Hyperliquid EVM

tribesxapp:
We added a game contract to hyperliquid EVM and would like to add USDT0 TVL from that contract


ghost:
The adapter at projects/perfect-pool exports TVL: 
        
 
 
    base                      4.38 k
    hyperliquid               638
    
    total                    5.01 k

PR #15372, opened 2025-07-04
fix: compute terminal-fi-predeposit tvl based on receipt tokens

DevTerminalFi:
The previous implementation counted only the available liquidity in the redemption vaults. Since this liquidity will be bridged to Converge at launch, the current adapter would break after bridging.

The new implementation uses receipt tokens as the source of TVL. This approach is more accurate because it removes discrepancies caused by conversion delays on the MPC side and is future-proof for the post-bridge phase.


ghost:
The adapter at projects/terminal-fi-predeposit exports TVL: 
        
 
 
    ethereum                  65.78 M
    
    total                    65.78 M

waynebruce0x:
When the underlying assets are bridged to Converge we should count the token balances on converge. We only use receipt tokens as a proxy for TVL where counting the locked assets isn't possible

PR #15392, opened 2025-07-07
feat: add WHLP to looping collective vaults

DashanMcCain:
### Changes
- Added `WHLP` vault to the looping collective vaults.
- Note, since this vault utilizes strategies on hyper core, there is no on-chain method to fetch the balances in the `HLP` vault at the moment until the read precompiles go live.

### DIFF
--- a/projects/looped-hype/index.js
+++ b/projects/looped-hype/index.js
@@ -7,26 +7,457 @@ const sanitizeAndValidateEvmAddresses = (addresses) => {
     .filter((address) => /^0x[a-fA-F0-9]{40}$/.test(address));
 };
 
-const LHYPE_VAULT_ADDRESS = ['0x5748ae796AE46A4F1348a1693de4b50560485562'];
+const LHYPE_VAULT_ADDRESS = '0x5748ae796AE46A4F1348a1693de4b50560485562';
+const WHLP_HLP_CONTROLLER_ADDRESS = '0x70bcc95d203920971c0d4C3EE8eC080c7ac29f18'
+const USDhl = '0xb50A96253aBDF803D85efcDce07Ad8becBc52BD5';
+
+const WHLP_VAULT_ADDRESS = '0x1359b05241cA5076c9F59605214f4F84114c0dE8'
+const HLP_VAULT_ADDRESS = '0xdfc24b077bc1425ad1dea75bcb6f8158e10df303'
+const WHLP_HYPER_CORE_MULTISIGS = [
+  '0x9fcB7066C8AeEe704f9D017996b490873b306E51',
+  '0x41f45A847bB6c8bFf1448FEE5C9525875D443b9E',
+  '0x296B1078D860c69C94CA933c4BcD2d6f192DD86e',
+  '0x31Cbd708B505d3A9A0dae336BC9476b694256e74',
+  '0xFBB47621086901487C7f3beC6F23205738d59e27',
+]
+
+// CoreReader addresses and constants
+const CORE_READER_VAULT_EQUITY = '0x0000000000000000000000000000000000000802' 
+const CORE_READER_SPOT_BALANCE = '0x0000000000000000000000000000000000000801' 
+const CORE_READER_MARGIN_SUMMARY = '0x000000000000000000000000000000000000080F'
+const USDHL_TOKEN_ID = 291n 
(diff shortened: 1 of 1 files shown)


ghost:
        
 
 
    hyperliquid               61.70 M
    
    total                    61.70 M

DashanMcCain:
Not sure why the `api.addUSDValue` is defaulting to `USDT`. Not sure if there's some standard here when a chain does not have a native stablecoin yet.

projects/looped-hype/index.js
```diff
@@ -1,29 +1,155 @@
...
+    },
+    body: JSON.stringify({
+      type: 'vaultDetails',
+      vaultAddress: HLP_VAULT_ADDRESS, // HLP vault on Hyper Core
+      user: userAddress, // Each multisig address that deposits into HLP for WHLP
+    }),
+  }
+
+  try {
+    const data = await fetch(HYPERLIQUID_INFO_URL, payload)
```
waynebruce0x: please use our SDK to fetch data on chain

projects/looped-hype/index.js
```diff
@@ -1,29 +1,155 @@
...
+    }
+    // Return 0 if followerState is null/undefined
+    const dataJson = await data.json()
+    
+    if (!dataJson.followerState) {
+      return 0
+    }
+    const vaultEquity = Number(dataJson.followerState.vaultEquity)
+    return vaultEquity
+  } catch (error) {
```
waynebruce0x: please remove all try/catch

waynebruce0x:
> Not sure why the `api.addUSDValue` is defaulting to `USDT`. Not sure if there's some standard here when a chain does not have a native stablecoin yet.

If you can return token balances that'd be preferred. USD balances are in USDT and you should use `misrepresentedTokens: true` in your exports

ghost:
        
 
 
    hyperliquid               60.64 M
    
    total                    60.64 M

DashanMcCain:
> > Not sure why the `api.addUSDValue` is defaulting to `USDT`. Not sure if there's some standard here when a chain does not have a native stablecoin yet.
> 
> If you can return token balances that'd be preferred. USD balances are in USDT and you should use `misrepresentedTokens: true` in your exports



Gotcha, added this and removed the try catch blocks

projects/looped-hype/index.js
waynebruce0x: we need to use RPC calls rather than https://api.hyperliquid.xyz/info for reading token balances and TVL

DashanMcCain:
Hi, @waynebruce0x. Are there more changes required to get this merged?

projects/looped-hype/index.js
waynebruce0x: please check this

bitcoinbender:
Can we merge this? @waynebruce0x

waynebruce0x:
Hey sorry for the hold up on this - where can this product/pool be seen on your UI? How come TVL here is now greater than that shown on the dapp UI? Thanks

bitcoinbender:
@waynebruce0x you can see the homepage here https://www.loopingcollective.org/. Total TVL across LHYPE & WHLP should be around $57M TVL currently.

waynebruce0x:
Please check the conflicts and then we can merge this

ghost:
        
 
 
    hyperliquid               37.98 M
    
    total                    37.98 M

ghost:
        
 
 
    hyperliquid               37.98 M
    
    total                    37.98 M

DashanMcCain:
@waynebruce0x, is there any way I can static call using the `api` object? We're migrating our `WHLP` vault to use `CoreWriter` and would also like to use `CoreReader` to enable fetching data purely from on-chain sources. However, in order to do this, I need to be able to make a static call to the system address.

waynebruce0x:
I think static calls should work in our SDK in the same way as normal reads

DashanMcCain:
> I think static calls should work in our SDK in the same way as normal reads

I see, but when I use `api.call` I'm met with an error stating that the abi field is required. I need to be able to just send an `eth_call` with arbitrary calldata

Note: I'll remove the try catch blocks as well once I can get the CoreReader implementation working

waynebruce0x:
I dont think we have any way of passing arbitrary calldata. Can you use ethers? Is there another way to achieve the same result?

PR #15425, opened 2025-07-10
Update River inc (prev: Satoshi protocol)

imfeng:
River inc add new Vault smart contract on Hemi
And remove xlayer chain TVL.


projects/satoshi-protocol/index.js
```diff
@@ -262,15 +273,5 @@ module.exports = {
       fromBlock: 33129825,
     }],
   }),
-  xlayer: createExports({
```
waynebruce0x: can we keep this? For historical accuracy

imfeng: @waynebruce0x 
Sure, I added it back

PR #15444, opened 2025-07-11
Fix River Omni CDP

0xpeluche:
Fix: a push yesterday broke the code with `safeVaultManagerList.forEach`, while `safeVaultManagerList` wasn’t an array, which naturally results in a `is not a function' error`

### DIFF
--- a/projects/satoshi-protocol/index.js
+++ b/projects/satoshi-protocol/index.js
@@ -212,10 +212,10 @@ module.exports = {
     ],
   }),
   hemi: createExports({
-    safeVaultManagerList: {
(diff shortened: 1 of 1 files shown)


ghost:
The adapter at projects/satoshi-protocol exports TVL: 
        
 
 
    hemi                      192.14 M
    bsquared                  40.70 M
    bob                       667.68 k
    bevm                      553.76 k
    btr                       48.73 k
    bsc                       14.00 k
    arbitrum                  13.34 k
    base                      480
    sonic                     70
    xlayer                    59
    
    total                    234.14 M

PR #15481, opened 2025-07-14
Fix: Dymension-dex

0xpeluche:
> Fix for Dymension-dex, their API was returning a malformed token that was preventing the code from executing


ghost:
The adapter at projects/dymension-dex.js exports TVL: 
        
 
 
    dymension                 3.36 M
    
    total                    3.36 M

PR #15540, opened 2025-07-17
Fix: BadgerDao

0xpeluche:
BSC has had no TVL since 2022, but it seems to have been breaking the code for the past few days because neither the API nor the cache return any data. Added a return if data is empty to allow the rest of the chains, which are still functional, to continue executing


ghost:
The adapter at projects/badgerdao.js exports TVL: 
        
 
 
    ethereum                  34.24 M
    arbitrum                  169.44 k
    polygon                   763
    bsc                       0
    fantom                    0
    
    total                    34.41 M

PR #15543, opened 2025-07-17
Fix Otsea

0xpeluche:
The adapter is no longer updating; there's an issue with the GraphQL endpoint it relies on. Switch the adapter to fetch data directly from the contracts instead of using GraphQL


ghost:
The adapter at projects/otsea exports TVL: 
        
 
 
    ethereum                  660.42 k
    base                      169.45 k
    blast                     10
    ethereum-staking          5
    staking                   5
    
    total                    829.88 k

PR #15644, opened 2025-07-24
fix: incorrect proton loan staking amount

KPHEMRAJ:
Even though price of LOAN token is 0.0006 , it was getting rounded to 0.001 and sometimes misbehaving
image


ghost:
The adapter at projects/proton-loan exports TVL: 
        
 
 
    proton-staking            42.79 M
    staking                   42.79 M
    proton                    23.00 M
    proton-borrowed           0
    borrowed                  0
    
    total                    23.00 M

PR #15674, opened 2025-07-28
fix: filter pools by hook type

abdurrwansui:
## Summary

This PR fixes an issue where TVL stopped updating due to the presence of a new hook type (type 5), which is not associated with any liquidity pool.

## Details

- We previously assumed all hooks were pool-related, but a new `hookType = 5` was introduced.
- Hook type 5 does not represent a pool and should be excluded from TVL calculation.
- This fix updates the adapter to only include hook types:
  - `2`, `3`, and `4` — valid pool hooks
- Additionally, we continue to ignore:
  - `hookType = 1` which is used for platform fees
  - `hookType = 5` (new) which is non-pool logic

## Outcome

- TVL will now accurately reflect only active pools
- This resolves the issue where TVL updates stopped after July 21 due to invalid data structures

### DIFF
--- a/projects/tapp-exchange/index.js
+++ b/projects/tapp-exchange/index.js
@@ -35,10 +35,9 @@ module.exports = {
     timetravel: false,
     aptos: {
         tvl: async (api) => {
-            const pools = (await getPools()).filter(pool=> pool.hook_type == 2 || pool.hook_type == 3 || pool.hook_type == 4);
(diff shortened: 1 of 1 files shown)


ghost:
The adapter at projects/tapp-exchange exports TVL: 
        
 
 
    aptos                     5.71 M
    
    total                    5.71 M

PR #15754, opened 2025-08-01
New adapter / Verified Network

kallolborah:
Verified_Logo_final-06

### DIFF
--- a/projects/edge-capital/index.js
+++ b/projects/edge-capital/index.js
@@ -1,0 +1,15 @@
+const { getCuratorExport } = require("../helper/curators");
+
+const configs = {
+  methodology: 'Count all assets are deposited in all vaults curated by Edge Capital.',
+  blockchains: {
+    tac:{
+      eulerVaultOwners: [
+        '0xB2b9a27a6160Bf9ffbD1a8d245f5de75541b1DDD',
+        '0x1280e86Cd7787FfA55d37759C0342F8CD3c7594a',
+      ],
+    },
+  }
+}
+
+module.exports = getCuratorExport(configs)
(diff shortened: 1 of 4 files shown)


ghost:
The adapter at projects/verified exports TVL: 
        
 
 
    ethereum                  126.70 k
    base                      0
    
    total                    126.70 k

waynebruce0x:
Can we read token balances on chain using our SDK instead?

ghost:
The adapter at projects/verified exports TVL: 
        
 
 
    ethereum                  126.70 k
    base                      0
    
    total                    126.70 k

ghost:
The adapter at projects/verified exports TVL: 
        
 
 
    ethereum                  126.70 k
    base                      0
    
    total                    126.70 k

kallolborah:
> Can we read token balances on chain using our SDK instead?

yes, we have updated the adapter code now to use the sdk

waynebruce0x:
We're still reading token balanaces from the subgraph - can this be done using the SDK and erc20:balanceOf instead?

Mohzcrea8me:
> We're still reading token balanaces from the subgraph - can this be done using the SDK and erc20:balanceOf instead?

We cannot use erc20:balanceOf as all of our pools are Balancer pools.

To calculate TVL, there are two available approaches:

1. **Using the Subgraph**(This is our current implementation.)

2. **Using getLogs via the SDK** (This would involve querying events for three different topics across various smart contracts)

We opted for the subgraph approach because it is more performant than using getLogs, and both methods yield the same values, as the subgraph is populated using those same events.

We are currently using the version/latest endpoint for the subgraphs to ensure we always access the most up to-date data

PR #15755, opened 2025-08-01
fix: wbeth tvl

KPHEMRAJ:
We were calculating tvl= wbeth supply * eth price, which is incorrect


ghost:
The adapter at projects/binance-staked-eth exports TVL: 
        
 
 
    ethereum                  10.84 B
    bsc                       741.63 M
    
    total                    11.58 B

PR #15758, opened 2025-08-01
feat: fix mento reserve query

bowd:
Fix mento reserve addresses and tracked assets, can be cross-checked with
https://reserve.mento.org/?tab=reserve-addresses

Changes:
- Add new reserve addresses
- Remove old reserve addresses
- Ensure all tokens held by reserve are tracked

---
There's still a delta compared to our own tracking, but those are assets held in Uniswap pools, that I believe won't be tracked on DefiLama


ghost:
The adapter at projects/mento exports TVL: 
        
 
 
    celo                      36.51 M
    ethereum                  19.53 M
    bitcoin                   1
    
    total                    56.03 M

bowd:
The linting error causing CI to fail wasn't introduced here and it looks also more like a bug than a pure linting thing, happy to fix it but somebody might want to look at it.

PR #15762, opened 2025-08-02
Add Bodega Market Creation Script Address

whankinsiv:
Adds the Bodega Market V3 Market Creation Address where users lock funds in order for the right to create a prediction market.


ghost:
The adapter at projects/bodega-market exports TVL: 
        
 
 
    cardano                   402.56 k
    
    total                    402.56 k

PR #15833, opened 2025-08-07
Add latest Neutral.Trade vaults

NeutralTrade:
- refactored logic to fetch drift vaults tvl
- add utils to fetch tvl of our kamino vault, hyperliquid vaults and NT Vaults
- update constants with all missing vaults


projects/neutral-trade/utils/drift.js
```diff
@@ -0,0 +1,112 @@
...
+
+function getProgram() {
+  const idl = require("../idl/drift_idl.json");
+  const programId = new PublicKey('dRiftyHA39MWEi3m9aunc5MzRF1JYuBsbn6VPcn33UH');
+  const provider = getProvider();
+  return new Program(idl, programId, provider);
+}
+
+async function getTvl(api) {
+  const vaultAddresses = DRIFT_VAULTS.map(vault => vault.address);
```
waynebruce0x, line 83: would `projects/helper/chain/rpcProxy.js` drift.vaultTvl() work here?

projects/neutral-trade/utils/hyperliquid.js
```diff
@@ -0,0 +1,29 @@
...
+      {
+        headers: {
+          "Content-Type": "application/json",
+        },
+      }
+    );
+    const accountValueHistory = data.portfolio[0][1].accountValueHistory;
+    const dp = 1e6;
+    return Number(accountValueHistory[accountValueHistory.length - 1][1]) * dp;
+  } catch (error) {
```
waynebruce0x: please remove all try/catch

waynebruce0x:
Are NT Vaults native to Neutral Trade? Is there any potential for double counting here?

projects/neutral-trade/utils/drift.js
NeutralTrade: I was using it before but it seems a bit buggy, the data would not update daily on our page.
So I copied the code from here https://github.com/DefiLlama/DefiLlama-Adapters/pull/15519/files

projects/neutral-trade/utils/hyperliquid.js
NeutralTrade: done

NeutralTrade:
> Are NT Vaults native to Neutral Trade? Is there any potential for double counting here?

Yes NT Vaults are using our own program. Regarding double counting, so far only NT Earn vault (nt vault) might have some double counting, because it may dynamically move funds to our usdc savings vault (drift) and our USDC Max Yield vault (kamino). I've disabled the usdc savings vault since we've deprecated direct deposits to it already, so it wont be double counted. Let's keep the USDC Max Yield vault counted as it is open to direct deposits on Kamino app 🙏

waynebruce0x:
We need a way to know how much of the NT Earn funds are deposited to USDC Max Yield since we can't double count deposits in TVL

NeutralTrade:
> We need a way to know how much of the NT Earn funds are deposited to USDC Max Yield since we can't double count deposits in TVL

I've disabled the Kamino vault as well, no double counting now

PR #15838, opened 2025-08-07
Add hallmark for MEV Capital vault expansion

Sithey:
## Summary
  - Added hallmark for August 2, 2025 documenting the expansion of MEV Capital's tracking beyond Morpho & Euler vaults

  ## Changes
  - Added hallmark entry explaining the significant TVL tracking expansion that occurred when the adapter was enhanced to support additional vault protocols (Terminal, Midas, Mizu, Napier, Hyperbeat, etc.)

  ## Context
  This hallmark provides transparency about when MEV Capital's adapter expanded its scope to track a broader range of vault types, which would have caused a notable increase in reported TVL.


ghost:
The adapter at projects/mev-capital exports TVL: 
        
 
 
    ethereum                  834.07 M
    hyperliquid               209.30 M
    avax                      35.16 M
    sonic                     32.05 M
    plume                     15.88 M
    berachain                 3.00 M
    unichain                  62.42 k
    bsc                       9.21 k
    
    total                    1.13 B

waynebruce0x:
Does this adapter support backfilling? Has anyone requested a backfill instead of adding the hallmark? Would this be a better solution?

Sithey:
> Does this adapter support backfilling? Has anyone requested a backfill instead of adding the hallmark? Would this be a better solution?

Thanks for the suggestion! Unfortunately, not all of the newly tracked tokens are currently supported by the DefiLlama coins API, so a full and accurate backfill isn’t possible yet.

I’ve added the hallmark for now to clearly indicate the TVL expansion date, and I’ll work on getting the missing price feeds integrated. Once those are available, I can remove the hallmark and run a proper backfill so the MEV Capital page looks cleaner and more professional.

PR #15896, opened 2025-08-12
fix: so-so value tvl

KPHEMRAJ:
Remove mag7 index from basis trading , as they dont fall under basis trading


ghost:
The adapter at projects/ssi-basis-trading exports TVL: 
        
 
 
    base                      8.22 M
    
    total                    8.22 M

waynebruce0x:
Does this code not count MEME.ssi etc?

KPHEMRAJ:
> Does this code not count MEME.ssi etc?

It only counts USSI , none of the others fall under basis trading.
Maybe we can put meme, mag7 etc under indexes

PR #15920, opened 2025-08-14
Update Flow Lending

whankinsiv:
This PR adds `trySumTokens` to cardano.js allowing 404 error when querying addresses. This is used by Flow as their API returns script addresses for pools that have not yet been funded.

### DIFF
--- a/projects/helper/chain/cardano.js
+++ b/projects/helper/chain/cardano.js
@@ -1,99 +1,196 @@
-const sdk = require('@defillama/sdk')
-const { post } = require('../http')
-const { getUniqueAddresses, nullAddress } = require('../tokenMapping')
-const { getAssets, getAddressesUTXOs, } = require('./cardano/blockfrost')
-const { PromisePool } = require('@supercharge/promise-pool')
+const sdk = require("@defillama/sdk");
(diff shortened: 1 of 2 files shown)


ghost:
The adapter at projects/flow exports TVL: 
        
 
 
    cardano                   8.68 M
    
    total                    8.68 M

projects/helper/chain/cardano.js
```diff
@@ -1,99 +1,196 @@
...
+    e?.status === 404;
+
+  // Wrap handlers so a 404 returns empty results instead of erroring
+  const safeGetAssets = async (addr) => {
+    try {
+      return await getAssets(addr);
+    } catch (e) {
+      if (is404(e)) {
+        // optional: console.warn(`Ignoring 404 for owner ${addr}`)
+        return [];
```
waynebruce0x, line 142: all these try/catches could cause under reported TVL and we wont be notified of it since our monitoring systems wont pick up the errors.

whankinsiv: Hi Wayne. This is intentional to provide a variant of SumTokens2 that is allowed to fail if an address provided by a project's API is invalid.

PR #15930, opened 2025-08-14
update token address

havana-earnium:
This PR fixes the token address retrieval, some tokens get fungible address so there is no price, so we fix it to get all address according to original address.


ghost:
The adapter at projects/earnium exports TVL: 
        
 
 
    aptos                     387.38 k
    
    total                    387.38 k

PR #16012, opened 2025-08-20
Update Superfluid Protocol adapter

elvijsTDL:
During backfilling of data the realShaman from Discord noticed that after Aug 2023 Polygon and Gnosis TVL hit 0$ , which is innacurate

The issue was the onchain resolvers that were used to map isListed in the graph query were deployed way after the protocol deployment , thus all of the tokens were returned as not listed and not picked up before the deployment of the resolver

https://gnosisscan.io/address/0x9634Fa35188C59DD3B0ADa0Dd9adD7e4ECefc0a9
https://polygonscan.com/address/0x8bDCb5613153f41B2856F71Bd7A7e0432F6dbe58

Also while running the tests for way back noticed that some of the chains would throw an error , because the api would try to fetch timestamp of a block before genesis block 

So in this PR i've done this:

- Lookup all tokens , not just the listed ones
- Paginate the token search results from subgraph as now it will go over 1k results on Base network
- Define subgraph start blocks to prevent failing queries
- Catch the error when api tries to fetch a block timestamp pre genesis block

I've ran the tests up to the start of the protocol deployment 2021-3-6 , and it would be great if it could backfilled to see the whole picture


ghost:
The adapter at projects/superfluid.js exports TVL: 
        
 
 
    optimism                  10.04 M
    base                      3.78 M
    arbitrum                  2.89 M
    xdai                      1.96 M
    polygon                   1.15 M
    celo                      778.43 k
    ethereum                  606.03 k
    avax                      268.01 k
    scroll                    40.36 k
    bsc                       15.46 k
    
    total                    21.53 M

majidrazaee2:
طی ۴سال فقط پرداخت کردم بی ان بی وپولیگان و اتریوم وبیشترین پرداخت که انجام دادم تتر وtrx بوده

majidrazaee2:
۴سال فقط به من گفته شد پرداخت کنم بی ان بی واتریوم وپولیگان بیشترین پرداخت من تتر وtrxبود

PR #16038, opened 2025-08-22
chore(ci): upgrade checkout to v5

ozrg:
Bumps checkout to v5 for future-proofing against Node 24 runner updates.
Requires runner v2.327.1+. Workflows compile the same.

ref: https://github.com/actions/checkout/releases/tag/v5.0.0


waynebruce0x:
I dont think theres any need for this change right now

PR #16040, opened 2025-08-22
Fix Hyperion pool tvl tracks

fantasyni:
Fix Hyperion some pools tvl not tracked on defilama


ghost:
The adapter at projects/hyperion exports TVL: 
        
 
 
    aptos                     103.86 M
    
    total                    103.86 M

PR #15995, opened 2025-08-19
Add Berachain support for Nabla Finance TVL tracking

Dakavon:
**Objective**
Enable TVL tracking of the Nabla Finance Crypto Hub on Berachain.

**Modifications**
- Add the Berachain router address to the config object to support TVL calculation.
- Rename addresses in the config object to clearly indicate they are related to the NablaRouter.

Link to protocol: https://app.nabla.fi/pools

image

PR #16067, opened 2025-08-25
Fix Berachain TVL adapter for Nabla Finance

Dakavon:
#### Objective
Fix for adding support for Berachain to the Nabla Finance TVL adapters.  
Preserves the existing methodology of including all pools ever backed by the backstop pool / included in the pool hub (even if later inactive, since users can still withdraw), matching the current logic used on Arbitrum and Base.

#### Why fix?
TVL did not show for Berachain. From discord:

> the berachain fromBlock is too old and there are RPC limitation that prevent us from fetching logs from too far back, we might have to use a subgraph to solve it or possible a different rpc. Could you update it to use a subgraph instead?

#### Approach
Use on-chain calls instead of event listeners to generate the swap pool list and fetch assets for TVL calculation.

Previous PR: https://github.com/DefiLlama/DefiLlama-Adapters/pull/15995

### DIFF
--- a/projects/nabla/index.js
+++ b/projects/nabla/index.js
@@ -1,51 +1,116 @@
 const { getLogs2 } = require("../helper/cache/getLogs");
 
 const config = {
-    base: {
-        router: "0x791Fee7b66ABeF59630943194aF17B029c6F487B",
-        fromBlock: 19980311,
-    },
     arbitrum: {
         router: "0x7bcFc8b8ff61456ad7C5E2be8517D01df006d18d",
         fromBlock: 240797440,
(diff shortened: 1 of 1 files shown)


ghost:
The adapter at projects/nabla exports TVL: 
        
 
 
    base                      184.67 k
    arbitrum                  154.25 k
    berachain                 107.12 k
    
    total                    446.04 k

ghost:
The adapter at projects/nabla exports TVL: 
        
 
 
    base                      183.89 k
    arbitrum                  153.59 k
    berachain                 107.30 k
    
    total                    444.78 k

Dakavon (commented):
LGTM

PR #16069, opened 2025-08-26
fix: Apply cumulative deposit interest in drift vault

YuenSC:
This PR fixes the calculation of spot position balances in Drift vaults.

Previously, deposits returned only the raw balance without applying cumulative deposit interest.
The change introduces getSpotMarketCumulativeDepositInterest and updates the calculation to properly apply the interest, ensuring balances accurately reflect the real deposited value.


ghost:
        
 
 
    solana                    19.35 M
    
    total                    19.35 M

PR #16071, opened 2025-08-26
New collateral contract added on Ostrich

ostrich-exchange:
We are supporting a new collateral on ostrich, the new tvl calculation will include collateral from the new contract as well


ghost:
The adapter at projects/ostrich exports TVL: 
        
 
 
    arbitrum                  159.05 k
    
    total                    159.05 k

PR #16072, opened 2025-08-26
Add support for new collateral on ostrich

ostrich-exchange:
We are adding support for a new collateral on ostrich, the current tvl calculation will include collateral from the new contract as well


ghost:
The adapter at projects/ostrich exports TVL: 
        
 
 
    arbitrum                  160.00 k
    
    total                    160.00 k

PR #16096, opened 2025-08-28
fix swap.coffee DEX TVL

SHADOWDANCH:
Split swap.coffee adapter into 2. Staking TVL and DEX TVL.
https://swap.coffee/

### DIFF
--- a/projects/helper/chain/ton.js
+++ b/projects/helper/chain/ton.js
@@ -68,7 +68,11 @@ async function getTokenRates({ tokens = [] }) {
 }
 
 async function getJettonsInfo(tokens){
-  return (await post("https://tonapi.io/v2/jettons/_bulk", {"account_ids": tokens}))["jettons"]
+  const result = []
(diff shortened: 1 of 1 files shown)


ghost:
The adapter at projects/swap-coffee-dex exports TVL: 
        
 
 
    ton                       401.21 k
    
    total                    401.21 k

projects/swap-coffee-stacking/index.js
```diff
@@ -1,8 +1,4 @@
-const { sumTokensExport, getJettonsInfo } = require('../helper/chain/ton')
```
waynebruce0x: please dont change the file name

waynebruce0x:
What is the reason for splitting this? Generally we would combine dex and staking where the products are of the same version

SHADOWDANCH:
> What is the reason for splitting this? Generally we would combine dex and staking where the products are of the same version

Alright, i rolledback split.

PR #16147, opened 2025-09-02
Olympus: Add Hallmarks for Major TVL Impacting Milestones

o0j4yj4y0o:
Submitted key inflection points that lead to shifts in TVL since launch including migrations and and introduction of new core contracts that enabled TVL to shift into other sister protocols (Cooler Loans to Origami and Peapod Finance, Etc.)


waynebruce0x:
Temporarily reverting this. Treasuries currently dont support hallmarks, I'm checking if we can add this

PR #16215, opened 2025-09-09
[Bucket] Remove depreciated Bucket Farm project

Jarekkkkk:


waynebruce0x:
we dont remove adapter files. Would you like us to mark this project as dead?

Jarekkkkk:
> we dont remove adapter files. Would you like us to mark this project as dead?

yes, we would like to depreciate it

PR #16226, opened 2025-09-09
feat: improve zivoe's TVL calculations

alexsserban:
Current TVL on [DefiLllama](https://defillama.com/protocol/zivoe?borrowed=true) shows as 5M (borrowed). 

This PR updates the TVL calculations. The expected values (based on running `node test.js projects/zivoe/index.js`) is to have 5M as borrowed and 1.36 M as TVL (non-borrowed). The expectation is that the historic TVL in the chart should reflect the changes.


ghost:
The adapter at projects/zivoe exports TVL: 
        
 
 
    ethereum-borrowed         5.00 M
    borrowed                  5.00 M
    ethereum                  1.36 M
    
    total                    1.36 M

PR #16241, opened 2025-09-10
Add Spark Savings `sUSDC` TVL

b-tarczynski:
Hey 👋, we would like to add new category under Spark (Combined) called Spark Savings. Here we would display sUSDC tvl read from on-chain data. The returned data matches https://data.spark.fi/savings/sUSDC values

### DIFF
--- a/projects/spark-savings/index.js
+++ b/projects/spark-savings/index.js
@@ -1,31 +1,37 @@
 const CONFIG = {
-    ethereum: {
-        susdc: '0xBc65ad17c5C0a2A4D159fa5a503f4992c7B545FE',
-    },
-    base: {
-        susdc: '0x3128a0F7f0ea68E7B7c9B00AFa7E41045828e858',
-    },
-    arbitrum: {
-        susdc: '0x940098b108fB7D0a7E374f6eDED7760787464609',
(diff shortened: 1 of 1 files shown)


ghost:
The adapter at projects/spark-savings exports TVL: 
        
 
 
    ethereum                  330.52 M
    optimism                  64.41 M
    unichain                  36.21 M
    arbitrum                  35.81 M
    base                      7.03 M
    
    total                    473.98 M

waynebruce0x:
How does Spark Savings interact with the other Spark functions? Are these funds seperate? Can we count the underlying asset value instead of using totalSupply?

b-tarczynski:
> How does Spark Savings interact with the other Spark functions?

Currently, Spark Savings sUSDC is a Spark brand savings product. When user deposits into USDC into the vault, the vault will swap this USDC into USDS via PSM then deposited into sUSDS. That's how it's tapping onto SSR.
> Are these funds separate?

Yes, these funds are separate since it's tapping fully into SSR, not on SLL.
> Can we count the underlying asset value instead of using totalSupply?

We can't as the USDC in the vault will be converted into USDS then sUSDS.

PR #16281, opened 2025-09-15
add mmarket tvl

louiseaux:
Updated Rysk V12 adapter to include MMarket contract in TVL calculations


ghost:
The adapter at projects/rysk-v12 exports TVL: 
        
 
 
    hyperliquid               12.72 M
    
    total                    12.72 M

PR #16347, opened 2025-09-19
add partner vault

0xdevtrn:
We have launched a vault for institutional users in addition to the regular user vault, and its TVL will also be tracked.
We are using Fireblocks and Ceffu for this vault as well, and will submit a PR with updated addresses as they become available.

Currently, the following two vaults are in use:
- rDkfnysh6MmpitowwYm3cEyYKzLG2zEWEX: Fireblocks, Partner Vault 1
- rHsvjnyYoAZFajhti4mb5bhcXTxLebqpKK: Fireblocks, Partner Vault 2


ghost:
The adapter at projects/doppler-finance exports TVL: 
        
 
 
    ripple                    69.86 M
    ethereum                  17.91 k
    
    total                    69.88 M

PR #16214, opened 2025-09-09
Update tokenMapping.js: re-added ankr-staked-flow

bz-hashtag-0780:
re-added 'ankr-staked-flow' which had been removed by another PR

Missing balance for the token:
Ankr Staked FLOW EVM: flow:0x1b97100ea1d7126c4d60027e231ea4cb25314bdb

PR reference: https://github.com/DefiLlama/DefiLlama-Adapters/pull/15558

PR #16356, opened 2025-09-22
Re-added ankr-staked-flow

bz-hashtag-0780:
re-added 'ankr-staked-flow' which had been removed by another PR

Missing balance for the token:
Ankr Staked FLOW EVM: flow:0x1b97100ea1d7126c4d60027e231ea4cb25314bdb

Previous [PR](https://github.com/DefiLlama/DefiLlama-Adapters/pull/16214) had been rejected as the token should be priced via [this](https://coins.llama.fi/prices/current/flow:0x1b97100ea1d7126c4d60027e231ea4cb25314bdb), but that doesn't seem to be the case anymore and ~$20m ANKRFLOWEVM TVL is not being counted in DeFi Llama currently.

Ref: https://coins.llama.fi/prices/current/flow:0x1b97100ea1d7126c4d60027e231ea4cb25314bdb
Screenshot 2025-09-22 at 09 45 33

### DIFF
--- a/projects/helper/tokenMapping.js
+++ b/projects/helper/tokenMapping.js
@@ -43,6 +43,9 @@ const fixBalancesTokens = {
   ozone: {
     // '0x83048f0bf34feed8ced419455a4320a735a92e9d': { coingeckoId: "ozonechain", decimals: 18 }, // was mapped to wrong chain
   },
+  flow: {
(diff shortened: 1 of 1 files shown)


waynebruce0x:
This seems fixed now not sure what happened there

PR #16358, opened 2025-09-22
Added oxium vaults v2 TVL

maxencerb:
## Updated Oxium: added vaults V2 TVL

Oxium recently created Vaults V2, and we need to index these new vaults TVL. The integration is mostly the same as on vaults V1 for the adapter, So I created a function to avoid code repetition.

The same way as before, our indexer is only used here to gather the list of vaults, but the TVL is then queried on-chain.


ghost:
The adapter at projects/oxium exports TVL: 
        
 
 
    sei                       2.81 M
    
    total                    2.81 M

PR #16417, opened 2025-09-25
Unified collateral vault

lukmaan-k:
All markets now share same collateral vault


ghost:
The adapter at projects/optfun exports TVL: 
        
 
 
    hyperliquid               8.00 k
    
    total                    7.99 k

PR #16428, opened 2025-09-26
fix: add second soneium vault on wavex

LeeWangKwang:
add second soneium vault(EL Vault) on wavex


ghost:
The adapter at projects/wavex exports TVL: 
        
 
 
    soneium                   578.50 k
    
    total                    578.50 k

PR #16458, opened 2025-09-29
Replace portal balances with synth balances for CrossCurve tvl

spiehdid:


ghost:
The adapter at projects/crosscurve exports TVL: 
        
 
 
    sonic                     3.90 M
    arbitrum                  2.62 M
    taiko                     961.61 k
    fantom                    151.84 k
    bsc                       7.30 k
    ethereum                  751
    base                      0
    xdai                      0
    celo                      0
    fraxtal                   0
    kava                      0
    mode                      0
    manta                     0
    polygon                   0
    optimism                  0
    blast                     0
    mantle                    0
    metis                     0
    linea                     0
    avax                      0
    
    total                    7.64 M

waynebruce0x:
Can we continur counting the deposit value instead of the total supply of minted assets?

spiehdid:
> Can we continur counting the deposit value instead of the total supply of minted assets?

It's more correct to count minted assets - in that way we get correct TVL in all chains where we have a bridge. Otherwise, the locked deposit is only visible to the deposit network, and the target network doesn't see the TVL increase, even though it's occurring (it always occuringe when we bridge some token from chain A to B).

waynebruce0x:
Value locked is on the source chain though so we should count it there

PR #16510, opened 2025-10-01
Add Vesu V2 pools

vesu-dev:


ghost:
The adapter at projects/vesu exports TVL: 
        
 
 
    starknet                  57.15 M
    starknet-borrowed         20.98 M
    borrowed                  20.98 M
    
    total                    57.15 M

PR #16544, opened 2025-10-03
Update Burve Protocol TVL 

itos-brian:
Updating the tvl calculation methodology. Added support for additional closures. 

Question: 
We support MEAD (https://www.coingecko.com/en/coins/mead-2) but it doesn't get valued as a stable. Is this on purpose, or is it something we'd need to add support for?

### DIFF
--- a/projects/burve/index.js
+++ b/projects/burve/index.js
@@ -1,32 +1,11 @@
-const ADDRESSES = require("../helper/coreAssets.json");
 const BERACHAIN = "0xa1beD164c12CD9479A1049f97BDe5b3D6EC21089";
-
-const tokens = [
-	ADDRESSES.berachain.USDC,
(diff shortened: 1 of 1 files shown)


ghost:
The adapter at projects/burve exports TVL: 
        
 
 
    berachain                 39.85 k
    
    total                    39.85 k

ghost:
The adapter at projects/burve exports TVL: 
        
 
 
    berachain                 59.88 k
    
    total                    59.89 k

waynebruce0x:
Does this value account for the MEAD deposits?

itos-brian:
> Does this value account for the MEAD deposits?

No, but they are very minimal at this point. 

EDIT: 
Once the price of MEAD is supported in the official gecko api, the price will be added.

PR #16564, opened 2025-10-06
Fix Moar TVL Calculations

lumos42:
Updated the logic to calculate TVL. Some credit accounts held Goblin Vaults' LP tokens which cannot be priced directly. This change converts the LP tokens into the underlying tokens with their balances.


ghost:
The adapter at projects/moar exports TVL: 
        
 
 
    aptos                     4.15 M
    aptos-borrowed            1.91 M
    borrowed                  1.91 M
    
    total                    4.15 M

waynebruce0x:
We have added pricing support for goblin vault tokens so there should no longer be any need to unwrap them here

lumos42:
Thank you @waynebruce0x. I can confirm that the TVL on the app looks correct now. I'll close the PR now.

PR #16571, opened 2025-10-06
fix/blacklisted-tokens

blablalf:
This PR removes the blacklist of SDEX tokens used as collateral in the [P2P lending protocol](https://defillama.com/protocol/smardex-p2p-lending).

Rationale:

- SDEX tokens are not staked but serve as collateral for active debt in other tokens.

- Blacklisting them misrepresents the protocol’s state, making it appear as if there is more borrowing than lending, and causing TVL to misleadingly drop from 800k to 10k.

- Comparable protocols, like AAVE, include both governance and governance debt tokens as collateral in their TVL calculations.


g1nt0ki:
> This PR removes the blacklist of SDEX tokens used as collateral in the [P2P lending protocol](https://defillama.com/protocol/smardex-p2p-lending).
> 
> Rationale:
> 
>     * SDEX tokens are not staked but serve as collateral for active debt in other tokens.
> 
>     * Blacklisting them misrepresents the protocol’s state, making it appear as if there is more borrowing than lending, and causing TVL to misleadingly drop from 800k to 10k.
> 
>     * Comparable protocols, like AAVE, include both governance and governance debt tokens as collateral in their TVL calculations.

@blablalf what percentage of total collateral is their own gov token in other protocols?

blablalf:
> > This PR removes the blacklist of SDEX tokens used as collateral in the [P2P lending protocol](https://defillama.com/protocol/smardex-p2p-lending).
> > 
> > Rationale:
> > 
> >     * SDEX tokens are not staked but serve as collateral for active debt in other tokens.
> > 
> >     * Blacklisting them misrepresents the protocol’s state, making it appear as if there is more borrowing than lending, and causing TVL to misleadingly drop from 800k to 10k.
> > 
> >     * Comparable protocols, like AAVE, include both governance and governance debt tokens as collateral in their TVL calculations.
> 
> @blablalf what percentage of total collateral is their own gov token in other protocols? 

Hey, there is a rule about this?

g1nt0ki:
yes, closest example I can recall is, parallel team was counting their own gov token as steaming tvl:  https://defillama.com/protocol/parallel-polkadot-daofi you will see a sharp drop at the start from 80M to zero, this was when we excluded their own tokens from streaming app tvl

blablalf:
> yes, closest example I can recall is, parallel team was counting their own gov token as steaming tvl: https://defillama.com/protocol/parallel-polkadot-daofi you will see a sharp drop at the start from 80M to zero, this was when we excluded their own tokens from streaming app tvl

I checked Aave’s historical data to see how governance token proportions evolved (at first, early in its lifecycle, similar to P2P’s current stage).

- For example, the LEND token (Aave’s original governance token) was the top 1 collateral counted toward Aave’s TVL at the start of the protocol.
screenshot_aave_v1 

- In our case, the SDEX token would represent approximately 5% of the SMARDEX ecosystem’s TVL if included.

- I also reviewed Aave V2 and found the same pattern:
screenshot_aave_v2

- If there’s no explicit rule against it, it appears accepted practice for Aave to include its governance token as collateral in TVL — since it’s used for active borrowing, meaning there’s real liquidity behind it. 

- As TVL and the number of assets grow, the share of governance tokens naturally decreases. Still, Aave historically had more than 5% of its governance token represented in its TVL, as shown below:
screenshot_aave_v3 

- Additionally, GHO is minted against AAVE but still counted as collateral, even though it’s a debt token. 

- About the Parallel protocol, they seem to only use their own governance token, the P2P protocol is permissionless — anyone can create a proposal with any token as collateral and borrowed token. The current proposals include a lot of USDT, which is unrelated to the SMARDEX ecosystem and is highly liquid, so liquidity risk is not a concern here.  

I hope this analysis clarifies our stance and contributes to a fair resolution regarding the blacklist.

PR #16584, opened 2025-10-07
update vault addresses and add btc multisig address checks

TerrenceAddison:
Just updating addresses for newly deployed


ghost:
        
 

PR #16613, opened 2025-10-09
Update Teller Adapter to include Lender Principal Supplied to Pools 

ethereumdegen:
This PR updates the Teller Protocol adapter to use Subgraphs for scraping data instead of using getLogs.  It computes TVL and Borrowed in the same way, except  for the TVL computation, it now also includes principal that lenders have supplied in to the new Teller Pools.   When this Adapter was originally created, Teller Pools did not exist; Teller only allowed users to create 1:1 OTC loans with one another.  Since Teller Pools now make up a majority of activity on the protocol, it makes sense to track and include lender supplied principal for these pools on DefiLlama now.

### DIFF
--- a/projects/teller/index.js
+++ b/projects/teller/index.js
@@ -142,13 +142,20 @@ async function bidCollateralTvl(api) {
     let skip = 0
     const pageSize = 1000
 
-    while (true) {
+    let hasMore = true
+    while (hasMore) {
       const data = await request(tellerGraphUrl, bidCollateralQuery, { skip })
-      if (!data.bidCollaterals || data.bidCollaterals.length === 0) break
+      if (!data.bidCollaterals || data.bidCollaterals.length === 0) {
+        hasMore = false
+        break
+      }
 
       allCollaterals = allCollaterals.concat(data.bidCollaterals)
 
-      if (data.bidCollaterals.length < pageSize) break
+      if (data.bidCollaterals.length < pageSize) {
+        hasMore = false
+        break
+      }
       skip += pageSize
     }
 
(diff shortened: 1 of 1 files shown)


waynebruce0x:
Where can users deposit to the Teller pools? On the UI at https://app.teller.org/ethereum/earn I can see around 6M$ TVL on ethereum, where the adapter returns more like 15M$ - do you know where this discrepancy comes from?

ethereumdegen:
Yes so there are two layers to Teller Protocol : the base OTC layer and the Pools layer .   Th e discrepancy is due to the base OTC layer.   

As you can see, this adapter works the same as the one that is published to DefiLlama now, except instead of using eventlogs it uses subgraph that is scraping those same eventlogs and now it is also including lender-supplied-principal in pools .  The previous iteration of the adapter didnt include that because pools didnt exist.  It only tracked the OTC activity .

waynebruce0x:
Where can I find more information about the lender-supplied-principal? How are OTC layer deposits made if they arent shown on the UI?

ethereumdegen:
Hello, you can pull the full list of transactions on Etherscan,  read our open source subgraph code that is powering this Adapter, or you can look on teller.org.  On Teller.org, the OTC loans are a recently deprecated feature on the frontend and the majority of the OTC loans use the MASA token.  We are switching all of our activity over to pools now as it is a significantly better UX for lenders. 


So the way this TVL is collected is using loan collaterals.  

```
const bidCollateralQuery = gql`
  query ($skip: Int!) {
    bidCollaterals(first: 1000, skip: $skip, where: { status: "Deposited" }) {
      id
      amount
      collateralAddress
      token {
        address
        symbol
        decimals
      }
      bid {
        status
      }
    }
  }
...
 let allCollaterals = []
    let skip = 0
    const pageSize = 1000

 while (true) {
      const data = await request(tellerGraphUrl, bidCollateralQuery, { skip })
      if (!data.bidCollaterals || data.bidCollaterals.length === 0) break

      allCollaterals = allCollaterals.concat(data.bidCollaterals)

      if (data.bidCollaterals.length < pageSize) break
      skip += pageSize
    }

   allCollaterals.forEach(collateral => {
      api.add(collateral.token.address, collateral.amount)
    })


`


```



You can read the source code of our subgraph here and it has remained unchanged for a very long time, it is unchanged compared to how it functioned with our currently published DefiLlama adapter.  Furthermore, the way that the subgraph collects this data is the same as the way the currently published DefiLlama collected it using EventLogs.  

https://github.com/teller-protocol/teller-protocol-v2/blob/develop/packages/subgraph/src/collateral-manager/mapping.ts


The only significant change that this new adapter makes, mathematically, is now it includes principal supplied by lenders in pools as TVL.   It includes total collateral deposited by borrowers the same as it did before and by running a test on the previous adapter and this new adapter I proved that this is true; the number is the same.

PR #16714, opened 2025-10-16
Remove delisted markets and add throttling for requests to node for Nolus

metodi96:


ghost:
        
 
ghost:
        
 
 
    osmosis                   1.99 M
    nolus                     950.18 k
    neutron                   47.28 k
    
    total                    2.99 M

projects/nolus/index.js
```diff
@@ -1,21 +1,67 @@
...
+  const results = new Array(leaseAddresses.length).fill(null)
+
+  for (let i = 0; i < leaseAddresses.length; i += BATCH.size) {
+    const start = i
+    const end = Math.min(i + BATCH.size, leaseAddresses.length)
+    const chunk = leaseAddresses.slice(start, end)
+
+    let attempt = 0
+    while (attempt <= BATCH.maxRetries) {
+      try {
```
waynebruce0x: is try / catch here necessary? This is likely to cause erratic TVL values

ghost:
        
 
 
    osmosis                   1.75 M
    nolus                     915.97 k
    neutron                   49.95 k
    
    total                    2.71 M

projects/nolus/index.js
metodi96: Hey, I've pushed a new commit where, instead of potentially allowing erratic TVL values, it simply fails if all attempts on a batch don't go through (I believe the most likely reason for this would be rate limiting). It behaves well currently). I believe that's also a viable option, right? Kept the try / catch for that reason.

PR #16741, opened 2025-10-18
feat: add USDe and WETH TVL support for terminal-fi-predeposit

DevTerminalFi:
- add USDe TVL for tUSDe redemption vault
- add WETH TVL for tETH redemption vault

As part of Terminal Finance preparations for launch, which will require unstaking part of sUSDe & weETH to setup the sUSDe/USDe & WETH/weETH liquidity pools, we need to update the TVL methodology to account for the unstaked version of those assets.

PR #16791, opened 2025-10-22
feat: update terminal-fi-predeposit TVL methodology

DevTerminalFi:
This PR is a generalization of https://github.com/DefiLlama/DefiLlama-Adapters/pull/16741 that computes `terminal-fi-predeposit` TVL based on the receipt tokens total supply & the DefiLlama pricing of those tokens.

This implementation is directly inspired by the way Midas compute their TVL.

IMPORTANT :

Before considering merging this PR, the DefiLlama pricing of `tUSDe` (`0xA01227A26A7710bc75071286539E47AdB6DEa417`) must be corrected. It is currently considered 1:1 with `sUSDe`, but in reality it is 1:1 with `USDe`.

### DIFF
--- a/projects/terminal-fi-predeposit/index.js
+++ b/projects/terminal-fi-predeposit/index.js
@@ -1,16 +1,20 @@
-const ADDRESSES = require('../helper/coreAssets.json')
-const { sumTokensExport } = require('../helper/unwrapLPs');
-
-module.exports = {
+const config = {
   ethereum: {
-    tvl: sumTokensExport({
-      tokensAndOwners: [
-        [ADDRESSES.ethereum.sUSDe, '0xFaAE52c6A6d477f859a740a76B29c33559ace18c'],
-        [ADDRESSES.ethereum.USDe, '0xFaAE52c6A6d477f859a740a76B29c33559ace18c'],
-        [ADDRESSES.ethereum.WEETH, '0xe042678e6c6871fa279e037c11e390f31334ba0b'],
-        [ADDRESSES.ethereum.WETH, '0xe042678e6c6871fa279e037c11e390f31334ba0b'],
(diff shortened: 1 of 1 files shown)


ghost:
The adapter at projects/terminal-fi-predeposit exports TVL: 
        
 
 
    ethereum                  319.47 M
    
    total                    319.47 M

waynebruce0x:
Why cant we continue counting the value of the token balances?

DevTerminalFi:
For security and accounting reasons, we prefer not to mix different assets within vaults that were initially designed to hold a single underlying.

In addition, computing TVL based on the receipt token supply is more future-proof: it remains valid regardless of where the underlying assets are held, whether directly in the vaults or later in the AMM (as will be the case post-launch).

waynebruce0x:
We cant use total supply here, it needs to continue reading token balances. TVL isnt denominated in tXXX tokens

DevTerminalFi:
What we are proposing is exactly the same methodology that Midas (our vault provider) [uses](https://github.com/DefiLlama/DefiLlama-Adapters/blob/main/projects/midas/index.js) to compute their TVL on DefiLlama.

This is not a novel or arbitrary approach, it’s an accepted method for vault-based systems where deposit tokens (receipt tokens) represent a proportional claim on the underlying assets.

TVL here is not “denominated in tTokens”, it is derived from them, using their DefiLlama-defined prices, exactly as done for Midas’ vaults and other similar adapters.

waynebruce0x:
Over the next couple of days we will refactor Midas and translate the new method over to terminal fi

waynebruce0x:
Hey so we've started to use an API from deltaY to track Midas TVL. They have a /sankey endpoint that shows us where all the funds are deployed so that we can verify balances. Is this feasible for Terminal Fi?

g1nt0ki:
closing this as stale, can re-open when there is some info to fetch /verify tvl onchain

PR #16825, opened 2025-10-24
fix: incorrect reporting of remora tvl

aaronovz1:
TVL is being incorrectly reported by the current adapter. There is no reason why tokens sitting in a FusionAMM pool should be excluded from the TVL. This methodology is not applied to other RMA providers such as xStocks.

I have removed the issuer logic so that it is now simply (token supply * token price). The same as how other RWA issuers are being represented.

### DIFF
--- a/projects/remora-markets/index.js
+++ b/projects/remora-markets/index.js
@@ -4,35 +4,29 @@ const data = [
     {
         "name": "Circle Internet Group rStock",
         "address": "5fKr9joRHpioriGmMgRVFdmZge8EVUTbrWyxDVdSrcuG", 
-        "issuer": "4A3psCthh7VBouNK24QSqhXmgZQA7BB9ReS8qruD8ZnF"
     },
     {
         "name": "MicroStrategy Inc rStock",
-        "address": "B8GKqTDGYc7F6udTHjYeazZ4dFCRkrwK2mBQNS4igqTv", 
-        "issuer": "CWcebtehmM2ijRXdhv37PkLhZTuteWp785kwxdB5QF2a"
+        "address": "B8GKqTDGYc7F6udTHjYeazZ4dFCRkrwK2mBQNS4igqTv",
     },
     {
         "name": "NVIDIA Corporation rStock",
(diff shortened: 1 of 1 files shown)


ghost:
The adapter at projects/remora-markets exports TVL: 
        
 
 
    solana                    3.42 M
    
    total                    3.42 M

waynebruce0x:
Are the tokens in the fusion pool owned by users? Have they been purchased? The logic here was that the issuer holds the majority of the total supply and it looked like they were uncirculating.

aaronovz1:
> Are the tokens in the fusion pool owned by users? Have they been purchased? The logic here was that the issuer holds the majority of the total supply and it looked like they were uncirculating.

I understand that view but this view is not being applied to xStocks or any other RWA issuer. How have you confirmed that all $80m of xStocks is not owned by them and is "circulating", when their liquidity pools only account for maybe 10% of that value?

All supply of rStocks is circulating. They are in liquidity pools which are tradable by users. There is no other way to get the tokens into users hands currently, other then fronting the liquidity (like we have done), or doing large OTC deals. This is the same for xStocks, we know this because we use the same broker. A user cannot mint/redeem with them, the same as with us currently. This will change in near future with updates on the broker side coming.

waynebruce0x:
We exclude xStock tokens from TVL where they are owned by the issuer, about 40M$. The issue is that for both Remora and xStocks is that the protocol can mint an arbitrary number of tokenised stocks which isnt representative of defi activity. There are many cases across our integrations where we exclude redundant assets from TVL.

aaronovz1:
> We exclude xStock tokens from TVL where they are owned by the issuer, about 40M$. The issue is that for both Remora and xStocks is that the protocol can mint an arbitrary number of tokenised stocks which isnt representative of defi activity. There are many cases across our integrations where we exclude redundant assets from TVL.

Yes, their "pre mint" address is excluded but I don't think you can really know for sure thats the extent of what they own. As I say - considering only 10% or less is actually usable liquidity. I'd argue there is more owned by them which should be excluded. Otherwise, please treat Remora the same and give RWA issuers the benefit of the doubt in this new and evolving industry. Perhaps disclaimers could be placed in the information section or make better use of the "methodology" label.

PR #16837, opened 2025-10-27
Fix pool asset iteration and ensure all multi-asset pools are included in TAPP Exchange TVL calculation

abdurrwansui:
This PR fixes the TVL calculation logic for TAPP Exchange by refactoring how pool assets and reserves are aggregated.

Issue

- Previously, the adapter iterated over filteredPools and:
- Added the first two tokens (coinA, coinB) separately
- Then iterated again through all pool.assets, re-adding tokens inconsistently
- As a result, multi-asset pools (e.g. USD1-USDT-USDC) and some secondary tokens were either undercounted or skipped, depending on the pool structure.
- This caused DeFiLlama’s displayed TVL to be significantly lower than the on-chain value.

Fix

- Simplified iteration to handle any pool length (2, 3, or more assets) in a single loop.
- Ensures each asset/reserve pair is properly mapped via getPairedCoin() before being added to the total.
- Removes duplicate api.add() calls for coinA and coinB, preventing double-counting.
- Improves compatibility with new TAPP tri-pools (e.g. USD1-USDT-USDC, WBTC-XBTC-ABTC).


ghost:
The adapter at projects/tapp-exchange exports TVL: 
        
 
 
    aptos                     10.57 M
    
    total                    10.57 M

PR #16842, opened 2025-10-27
Fix: Gaming-dex

0xpeluche:
> There doesn’t seem to be any liquidity on the DefiVerse chain, and even the block explorer itself doesn’t appear to be working


ghost:
The adapter at projects/gaming-dex exports TVL: 
        
 
 
    oas                       1.83 M
    defiverse                 0
    
    total                    1.83 M

PR #16870, opened 2025-10-28
Update TVL function to use new contract addresses

unhappyben:
This is an update to zkp2p as we will be releasing our new V3 tomorrow, with a new contract. Tracking is exactly the same before, let me know if any questions!


ghost:
The adapter at projects/zkp2p exports TVL: 
        
 
 
    base                      77.56 k
    
    total                    77.56 k

PR #16880, opened 2025-10-29
Fix: add block-scoped cache key for SFPM chunks (v1 & v1.1)

BrandonL20:
V1 and v1.1 adapters will need to be re-ran for the entire historical period to ensure accuracy of TVL data:
- **For V1: Re-run from Dec 15, 2024**
- **For V1.1: Re-run from Feb 22, 2025**

This PR fixes an issue where cachedGraphQuery reused the same cached SFPM chunk data across multiple days when running TVL over multiple days.

Previously, both the v1 and v1.1 adapters used a static cache key like
`panoptic/v1/${chain}/sfpm-chunks`
even when `useBlock: true` was enabled.

Because the cache key didn’t include the block number, all subsequent days in a multi-day run reused the first day’s cached response, leading to incorrect TVL values in range mode.

This change updates both adapters to include the current block in the cache key:
`panoptic/v1/${chain}/sfpm-chunks@${api.block}`
`panoptic/v1_1/${chain}/sfpm-chunks@${api.block}`

This ensures each day’s subgraph query is cached separately and matches single-day snapshot results.

Impact
* Fixes TVL discrepancies between single-day and multi-day runs

### DIFF
--- a/projects/panoptic-v1_1/index.js
+++ b/projects/panoptic-v1_1/index.js
@@ -67,8 +67,8 @@ async function tvl(api) {
     api.addTokens(tokens, balancesRaw)
   }
 
-
-  const chunks = await cachedGraphQuery(`panoptic/v1_1/${chain}/sfpm-chunks`, graphUrl, SFPMChunksQuery, { api, useBlock: true, fetchById: true, safeBlockLimit, })
+  const block = api.block ?? 0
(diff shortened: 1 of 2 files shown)


ghost:
The adapter at projects/panoptic-v1_1 exports TVL: 
        
 
 
    base                      3.07 k
    unichain                  89
    ethereum                  8
    
    total                    3.17 k

ghost:
The adapter at projects/panoptic exports TVL: 
        
 
 
    ethereum                  13.02 k
    unichain                  638
    base                      274
    
    total                    13.94 k

waynebruce0x:
We can only backfill to start of this month for both V1 and V1.1

PR #16883, opened 2025-10-30
Hydrex Omni TVL Integration

larrettgee:
As Hydrex moves into the direction of sourcing liquidity across multiple protocols on Base, we are suggesting to categorize anything outside of Hydrex's own LPs under this single "Hydrex Omni" adapter.

We have taken our previous Hydrex-UniV4 adapter & modified enhanced it by adding our Morpho gauges as well. We have removed the previous adapter & will plan to add onto Hydrex Omni as we go forward.

The new Morpho methodology queries Morpho's vault contracts onchain directly with the balance of the Hydrex gauge. UniV4 methodology remains unchanged.


ghost:
The adapter at projects/hydrex-omni exports TVL: 
        
 
 
    base                      1.72 M
    
    total                    1.72 M

waynebruce0x:
Please use the original file path, we can update the metadata shown on the front end

ghost:
The adapter at projects/hydrex-univ4 exports TVL: 
        
 
 
    base                      2.83 M
    
    total                    2.83 M

larrettgee:
Sounds good thank you @waynebruce0x. I've updated this.

My ideal case would be calling the `hydrex-univ4` adapter "Hydrex Omni" on the frontend.

If we could also call the `hydrex` adapter "Hydrex Integral" that would be great too.

PR #16894, opened 2025-10-30
restaking milktia 

waynebruce0x:


ghost:
The adapter at projects/milky-way-restaking exports TVL: 
        
 
 
    milkyway                  507.01 k
    
    total                    507.01 k

PR #16165, opened 2025-09-04
Add WinnieSwap on Berachain

zdeadex:
https://github.com/DefiLlama/dimension-adapters/pull/4160

---
##### Name (to be shown on DefiLlama): 
WinnieSwap

##### Twitter Link:
https://x.com/WinnieSwap

##### List of audit links if any:
univ3
arrakisv1
more incoming

##### Website Link:
https://winnieswap.com

##### Logo (High resolution, will be shown with rounded borders):
https://ibb.co/WN7V3dZj

##### Current TVL:
10k

##### Treasury Addresses (if the protocol has treasury)
N/A

##### Chain:
Berachain

##### Coingecko ID (so your TVL can appear on Coingecko, leave empty if not listed): (https://api.coingecko.com/api/v3/coins/list)
We don't have a coin

##### Coinmarketcap ID (so your TVL can appear on Coinmarketcap, leave empty if not listed): (https://api.coinmarketcap.com/data-api/v3/map/all?listing_status=active,inactive,untracked&start=1&limit=10000)
We don't have a coin

##### Short Description (to be shown on DefiLlama):
Supercharging DeFi with Berachain Proof of Liquidity - enjoy deeper liquidity, smarter capital efficiency & sky-high APRs

##### Token address and ticker if any:
N/A

##### Category (full list at https://defillama.com/categories) *Please choose only one:
[Dexs](https://defillama.com/protocols/dexs)

##### Oracle Provider(s): Specify the oracle(s) used (e.g., Chainlink, Band, API3, TWAP, etc.):
N/A

##### Implementation Details: Briefly describe how the oracle is integrated into your project:
N/A

##### Documentation/Proof: Provide links to documentation or any other resources that verify the oracle's usage:
https://docs.winnieswap.com

##### forkedFrom (Does your project originate from another project):
univ3/arrakisv1

##### methodology (what is being counted as tvl, how is tvl being calculated):
pool tvl, addition of the two tokens on the pool

##### Github org/user (Optional, if your code is open source, we can track activity):
https://github.com/StakeLab-Zone

PR #16928, opened 2025-11-03
Winnieswap improvements

zdeadex:
Initial PR: https://github.com/DefiLlama/DefiLlama-Adapters/pull/16165

Improvements adding
- all our pools/vaults considered (our NECT/HONEY one wasn't in the list)
- autowin vaults

### DIFF
--- a/projects/winnieswap/index.js
+++ b/projects/winnieswap/index.js
@@ -1,74 +1,29 @@
 const stickyFactory = '0x18B9ABf2E821E2fE7A08Dc255d5a7e77fFc0b844'
-const autoWinFactory = '0x21b35a3dcF462540cb8EEA0e5d8594FF9e93C2e1'
 
 const abis = {
   getDeployers: 'address[]:getDeployers',
   getStickyVault: 'function getStickyVaults(address deployer) view returns (address[])',
-  getAutoWinVaults: 'function getAutoWinVaults(address deployer) view returns (address[])',
-  getUnderlyingBalances: 'function getUnderlyingBalances() view returns (uint256 amount0Current, uint256 amount1Current)',
-  token0: 'address:token0',
-  token1: 'address:token1',
 }
 
-const stickyVaultsTVL = async (api) => {
+const akkarisTVL = async (api) => {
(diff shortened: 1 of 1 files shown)


waynebruce0x:
It seems like these vaults are holding dex LP tokens? Is that right? In that case this would be double counting and we cant re-count these

zdeadex:
@waynebruce0x yes indeed but for example if you look at the current winnieswap dashboard on defillama, the counting is correct even tho you have already 
- the univ3 (not lp)
- the sticky vault (sticky lp tokens)

the added autowin vaults are just new lp tokens (aw lp tokens)
it's shoudln't be counted twice

also we fixed the time to have the nect/honey counted as we don't see it counted in the dasboard and now should be visible

waynebruce0x:
So a user will deposit funds to a pool, then take that LP receipt token and deposit that to a sticky vault or autowin vault? Or are these vaults holding LP tokens from elsewhere?

zdeadex:
The user can
- deposit into a pool and get an NFT position (similar to univ3)
- deposit into a vault and get sticky tokens
- deposit into a vault and get aw tokens

these can't be stake in the same vault, you can stake them in separated vaults like on BEX or Infrared for example

zdeadex:
so it's good? @waynebruce0x

waynebruce0x:
It seems like the vaults are just double counting the same underlying assets in the liquidity pools, in which case we cant include this as TVL

zdeadex:
Alright just changed the block to have all our pools included, should be good now @waynebruce0x

ghost:
The adapter at projects/winnieswap exports TVL: 
        
 
 
    berachain                 288.29 k
    
    total                    288.29 k

zdeadex:
there is still a problem, the vault https://www.winnieswap.com/vault/0xa5b3ecafbfcc7946c11a28252c6f2bdb73d30388 isn't included in the data reported...

waynebruce0x:
I think this might be a caching issue - I will merge this now and keep an eye on the issue

PR #16937, opened 2025-11-04
vaquita protocol - add multi-strategy vaults queries

raptor0929:
UPDATE:

- Change the contract address in Base
- Add multi-strategy vaults queries for each supported asset


ghost:
The adapter at projects/vaquita-protocol exports TVL: 
        
 
 
    base                      1.03 k
    
    total                    1.04 k

PR #16939, opened 2025-11-04
Reya rhedge pool

ioanaroceanu2000:
Adding rHEDGE token that is owned by the Reya Pool


ghost:
The adapter at projects/reya-perp exports TVL: 
        
 
 
    reya                      25.72 M
    
    total                    25.72 M

PR #16971, opened 2025-11-06
Lombard/BTC.b as several adapter, correct tvl for LBTC and BTC.b, names update

VaitaR:
- BTC.b as several adapter under Lombard
- Correct TVL distr. for LBTC and BTC.b


projects/lombard-btcb/index.js
```diff
@@ -0,0 +1,39 @@
+const { sumTokens } = require('../helper/chain/bitcoin')
```
waynebruce0x, line 1: please make these changes to `projects/avalance-btc/index.js` instead

PR #16980, opened 2025-11-07
fix: handle case when market has been deleted

SmolinPavel:
For [Templar Protocol](https://defillama.com/protocol/templar-protocol) adding a fix that will ignore deleted market. 

Before this PR adapter would crash if the market has been deleted.


ghost:
The adapter at projects/templarfi exports TVL: 
        
 
 
    near                      8.13 M
    near-borrowed             1.82 M
    borrowed                  1.82 M
    
    total                    8.13 M

PR #16999, opened 2025-11-10
Remove API Endpoint Dependency

DashanMcCain:
Removes the dependency on API endpoints for `looped-hype` and `wrapped-hlp` adapters.

### DIFF
--- a/projects/looped-hype/index.js
+++ b/projects/looped-hype/index.js
@@ -1,29 +1,25 @@
-const { getConfig } = require("../helper/cache")
-
 const LHYPE_VAULT = '0x5748ae796AE46A4F1348a1693de4b50560485562'
-const CHAIN_ID_STR = '999'
-const HYPE_TOKEN = '0x5555555555555555555555555555555555555555'
+
+const LHYPE_ACCOUNTANT = '0xcE621a3CA6F72706678cFF0572ae8d15e5F001c3'
+
+const WHYPE = '0x5555555555555555555555555555555555555555'
+
+const totalSupplyAbi = 'function totalSupply() view returns (uint256)'
+const exchangeRateAbi = 'function getRate() view returns (uint256)'
 
 const tvl = async (api) => {
-  const data = await getConfig(
-    'lhype-tokens',
-    `https://backend.nucleusearn.io/v1/vaults/underlying_strategies?vault_address=${LHYPE_VAULT}&chain_id=${CHAIN_ID_STR}`
-  )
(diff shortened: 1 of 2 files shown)


ghost:
        
 
 
    hyperliquid               41.45 M
    
    total                    41.45 M

ghost:
The adapter at projects/wrapped-hlp exports TVL: 
        
 
 
    hyperliquid               6.39 M
    
    total                    6.39 M

projects/looped-hype/index.js
```diff
@@ -1,29 +1,25 @@
...
+  const [lhypeTotalSupply, lhypeExchangeRate,] = await Promise.all([
+    api.call({ target: LHYPE_VAULT, abi: totalSupplyAbi }),
+    api.call({ target: LHYPE_ACCOUNTANT, abi: exchangeRateAbi, }),
+  ])
 
-  let totalHype = 0
-  for (const pos of Object.values(strat)) {
-    totalHype += Number(pos.valueInBase || 0)
-  }
+  const lhypeTotalValueInHype = lhypeTotalSupply * lhypeExchangeRate / 1e18
```
waynebruce0x, line 16: is it possible to count the underlying deposits rather than using total supply of LHYPE?

projects/wrapped-hlp/index.js
```diff
@@ -1,25 +1,21 @@
...
-  const strat = data?.[CHAIN_ID_STR]
-  if (!strat) return
+  const [whlpTotalSupply, whlpExchangeRate,] = await Promise.all([
+    api.call({ target: WHLP_VAULT, abi: totalSupplyAbi }),
+    api.call({ target: WHLP_ACCOUNTANT, abi: exchangeRateAbi, }),
+  ])
 
-  let totalUsd = 0
-  for (const pos of Object.values(strat)) totalUsd += Number(pos.valueInBase || 0)
+  const whlpTotalValueInUsdt0 = whlpTotalSupply * whlpExchangeRate / 1e6
```
waynebruce0x, line 15: same question here

projects/looped-hype/index.js
DashanMcCain: If there's a getter for fetching event logs I'd be able to fetch deposit events and subtract the withdrawal events to give a somewhat better estimate on the value.

I'd prefer to just use the tokens and owners approach but the underlying strategies of LHYPE changes over time. Meaning I'd have to update these adapters every time we go into a new pendle market or things of that nature. There's also the problem of the tokens and owners approach not supporting all of the strategies that we could go into.

PR #17010, opened 2025-11-11
Fix origyn

0xpeluche:
> issue with feed price → CG price


ghost:
The adapter at projects/origyn exports TVL: 
        
 
 
    icp                       51.31 M
    icp-staking               3.97 M
    staking                   3.97 M
    
    total                    51.31 M

PR #17026, opened 2025-11-12
chore: update CU url

ankushKun:
The updated Compute Unit node at cu.ardrive.io is much faster and stable than the testnet Compute Unit route at cu.ao-testnet.xyz, which would soon be deprecated


ghost:
The adapter at projects/LiquidOps exports TVL: 
        
 
 
    ao                        73.38 k
    ao-borrowed               25.43 k
    borrowed                  25.43 k
    
    total                    73.38 k

PR #17044, opened 2025-11-14
fix<Vecits>: Update TVL calculation by adding direct drift user address

YuenSC:
This pr include the tvl from some drift users that is under vectis by fetching those drift user addresses from api


ghost:
        
 
 
    solana                    26.43 M
    
    total                    26.43 M

PR #17082, opened 2025-11-18
Slipstream V2 handler & Arcadia account version upgrade

j-vp:
Aerodrome Slipstream V2 was recently deployed on Base. Although code is very similar, this can be considered a fully new protocol (new factory, new pool impl, new position manager) -> handle Slipstream V2 positions correctly.

Added resolveSlipstreamV2 with unwrapSlipstreamV2NFTs as to not potentially break any other project's code that depends on the slipstream resolver. It uses the same code for the inner unwrapping logic, only the nft address checks in unwrapSlipstreamV2NFTs are different.

Added Arcadia Account version 3 and 4. Version 3 = margin account and handles the same way as version 1, similar for account version 4 to account version 2. All 4 account versions (can) coexist within the protocol.

Improved performance by batching spot_asset_data calls into batches of 5 and removing the sleep inbetween calls, as there is no rate limit for defillama.


ghost:
        
 
 
    base                      8.17 M
    
    total                    8.17 M

PR #17138, opened 2025-11-24
Update Boson protocol TVL calculation

zajck:
Updating an existing project, `Boson Protocol`. 

Previously, the adapter was showing only staking on ethereum.  

With this PR, we added the TVL calculation, which now includes RWAs tokenised in Boson Protocol and all ERC20 tokens and ETH deposited and locked in the protocol contracts.

Additionally, all chains, where the protocol is deployed, were added to the adapter.

### DIFF
--- a/projects/boson/utils.js
+++ b/projects/boson/utils.js
@@ -1,6 +1,6 @@
 const ethers = require("ethers");
 const { getLogs } = require("../helper/cache/getLogs");
-const { sumTokens2 } = require("../helper/unwrapLPs");
+const sdk = require('@defillama/sdk');
 
 const {
   protocolDiamondAddress,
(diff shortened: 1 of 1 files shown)


ghost:
The adapter at projects/boson exports TVL: 
        
 
 
    polygon                   5.33 M
    ethereum-staking          523.78 k
    staking                   523.78 k
    ethereum                  2.04 k
    arbitrum                  0.00
    base                      0.00
    optimism                  0.00
    
    total                    5.33 M

waynebruce0x:
Where on the UI can I make a deposit?

zajck:
@waynebruce0x 

The dApp is available at https://bosonapp.io/

#### As a seller:
1. You need to create a seller profile first.
2. Go to "Seller hub"
3. On the left, choose "Finances"
4. Deposit any of the tokens currently supported in the dApp
image


#### As a buyer:
1. Find an item you'd like to buy.
2. Click on "Buy with BOSON".
3. image  
4. This deposits your funds in the protocol. The funds are locked until the exchange is finalised. If everything goes well, the funds are released to the seller. But if a buyer raises a dispute, funds are locked until the final decision is made, and the funds are then distributed to the seller and/or buyer accordingly.

waynebruce0x:
Where is this TVL stored? Can we please use erc20:balanceOf() instead?

zajck:
There are two contributing factors to TVL:
1. Tokens deposited in the protocol contract (either by buyers or sellers). These can be obtained with erc20:balanceOf() and I pushed the change that does it. The protocol contract address is `0x59A4C19b55193D5a2EAD0065c54af4d516E18Cb5` on all chains. 
2. On-chain offers which tokenize real world assets. Sellers are obliged to deliver the sold asset if a buyer commits to their offer. These values cannot be obtained with erc20:balanceOf() until someone actually commits to it. We consider the offer price as the locked value. If the offer expires or it is voided, the value is removed from TVL. Once someone commits to it, and if the offer becomes unavailable for further commits, it is also removed from TVL to avoid double-counting, since the value is already captured in (1).

zajck:
@waynebruce0x any additional comments?
What can we do to progress?

projects/boson/index.js
```diff
@@ -1,17 +1,80 @@
...
+
+async function tvl(api) {
+  // Currently Deposited and Encumbered funds.
+  // Includes all buyer commitments, seller deposits and dispute resolution fees for active exchanges for both Fixed Price and Price Discovery offers.
+  // It also includes all additional commits as result of Sequential Commitments
+  await getDepositedAndEncumberedBalances(api);
+
+  // Listed offers with uncommitted quantity
+  const voidedOffers = await getVoidedOffers(api);
+  const activeOffers = await getOffers(api, voidedOffers);
```
waynebruce0x, line 18: Where is the TVL of these offers stored? Can we use balanceOf again here?

zajck: As written in [my comment](https://github.com/DefiLlama/DefiLlama-Adapters/pull/17138#issuecomment-3574426128), the offers represent tokenized RWA. The value is reported by the seller. There exists a corresponding ERC721 token, but until someone commits to buy it, there is no way to get the value using balanceOf.   

We included the offers in TVL calculation as per [the documentation](https://docs.llama.fi/list-your-project/what-to-include-as-tvl):
image

We'd be happy to report that separately (similarly to what is possible for other types, such as staking, vesting, pool2 etc.), but it seems it cannot be exported in a similar manner:
image

PR #17232, opened 2025-11-29
triggerx should be doublecounted

noateden:


ghost:
The adapter at projects/triggerx exports TVL: 
        
 
 
    ethereum                  1.92 B
    
    total                    1.92 B

PR #17307, opened 2025-12-05
feat: add native USDC for Opus

tserg:
This PR adds a new contract to support the migration of bridged USDC to native USDC on Starknet.


ghost:
The adapter at projects/opus exports TVL: 
        
 
 
    starknet                  517.85 k
    
    total                    517.84 k

PR #17403, opened 2025-12-13
update SundaeV3 to calculate based on individual token value

whankinsiv:
This PR updates the sundae-v3 adapter to calculate TVL based on individual token value instead of 2 times ADA TVL. This increases TVL as pools which do not include ADA are now accounted for. It also improves transparency of the TVL breakdown.


ghost:
The adapter at projects/sundaeswap-v3 exports TVL: 
        
 
 
    cardano                   6.21 M
    
    total                    6.21 M

PR #17409, opened 2025-12-14
Update Strike Finance adapter

whankinsiv:
This PR updates the Strike Finance adapter to fetch all addresses associated with the stake key instead of using hardcoded addresses. To support this, I've added the `getAccountAddresses` helper function to `/helper/chain/cardano/blockfrost.js`. The staked protocol token address is filtered out of the list before TVL is calculated.


ghost:
The adapter at projects/strike-finance exports TVL: 
        
 
 
    cardano-staking           10.22 M
    staking                   10.22 M
    cardano                   5.50 M
    
    total                    5.50 M

PR #17505, opened 2025-12-28
Add Injective staking TVL

davidzk3:
**Summary**
Adds an Injective staking TVL adapter.

**Methodology**
TVL equals the total amount of INJ bonded in staking.
The adapter queries the Cosmos staking pool endpoint and reads the bonded_tokens value, which represents the total INJ currently staked by validators and delegators.

**Implementation**
Uses public Injective REST LCD endpoints with simple fallback and the existing DefiLlama test runner.
No UI, wallet connection, or external dependencies required.

**Verification**
The staking TVL can be independently verified without any website UI or connecting a wallet using the following methods.

**Method 1 - DefiLlama test runner (reproduce the number locally)**
From the DefiLlama Adapters repository root, run:

```bash
node test.js projects/injective-staking/index.js
```

This prints the same staking TVL the adapter returns in CI.
Note that the USD number can change between runs because the INJ price changes over time.

**Method 2 - Direct Cosmos staking query and manual math (raw on chain value)**
**Step 1** - Query the Cosmos staking pool endpoint (public JSON):

```bash
curl "https://injective-rest.publicnode.com/cosmos/staking/v1beta1/pool"
```

You will get a JSON response containing: `pool.bonded_tokens`

**Step 2** - Convert base units to INJ
Injective base units use 1e18 decimals, so:

INJ_staked = `bonded_tokens` / 1e18

Example using the value shown in local testing:

`bonded_tokens` / 1e18 ≈ 56,789,341 INJ

**Step 3** - Convert to USD using the same price source DefiLlama uses
DefiLlama pricing for INJ comes from Coingecko (coingecko:injective-protocol).
From the test runner output, you can see the price used.

Example:

56.8M INJ × ~$4.8 ≈ ~$273M

This matches the adapter output up to normal price movement and rounding.

**Method 3 - Cross check using a different public endpoint (redundancy)**
This shows the bonded_tokens value is not tied to a single provider.

**Step 1** - Query an alternative endpoint:

```bash
curl "https://lcd.injective.network/cosmos/staking/v1beta1/pool"
```

Or:

```bash
node -e "fetch('https://lcd.injective.network/cosmos/staking/v1beta1/pool').then(r=>r.json()).then(console.log).catch(console.error)"
```

**Step 2** - Compare results
You should see the same bonded_tokens value (or extremely close).
If one endpoint is temporarily unavailable, the adapter fallback list helps ensure reliability.

**Chain**
Injective

### DIFF
--- a/projects/injective-staking/index.js
+++ b/projects/injective-staking/index.js
@@ -1,0 +1,69 @@
+// Injective staking TVL adapter
+// TVL equals total INJ bonded in staking
+// Reads Cosmos staking pool bonded_tokens via LCD endpoints with fallback
+
+const LCDS = [
+  "https://lcd.injective.network",
+  "https://injective-rest.publicnode.com",
+  "https://k8s.mainnet.lcd.injective.network:443",
+];
+
+const INJ_COINGECKO = "coingecko:injective-protocol";
+const INJ_DECIMALS = 10n ** 18n;
+
+async function fetchJsonWithFallback(path) {
+  let lastErr;
+  for (const base of LCDS) {
+    try {
+      const res = await fetch(`${base}${path}`, {
+        method: "GET",
+        headers: { accept: "application/json" },
+      });
+
+      const txt = await res.text();
+
+      if (!res.ok) {
+        lastErr = new Error(`LCD ${base} error ${res.status}: ${txt.slice(0, 160)}`);
+        continue;
+      }
+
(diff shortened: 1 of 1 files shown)


ghost:
The adapter at projects/injective-staking exports TVL: 
        
 
 
    injective                 268.23 M
    
    total                    268.23 M

PR #17518, opened 2025-12-30
fix: compute bnpl-pay borrowed tvl from on-chain loan state

davidzk3:
### Summary
This PR fixes the BNPL Pay adapter so `borrowed` TVL is computed from on-chain loan state instead of being silently overridden to `{}`.

#### What changed
- Removed the forced empty `borrowed` override
- Implemented on-chain borrowed TVL calculation by:
  - Enumerating banking nodes from the factory
  - Fetching active loan IDs per node
  - Summing `principalRemaining` for non-slashed loans
- Avoids inferring or backfilling values when no loans exist

#### Validation
- Verified **13 active banking nodes** on Ethereum
- Confirmed `getCurrentLoansCount() == 0` for all nodes
- **Borrowed currently resolves to 0 because all banking nodes report zero active loans**

#### Why this matters
Previously, borrowed TVL was always forced to zero, causing misleading flatline and decay behavior in historical charts.  
This change ensures borrowed TVL accurately reflects real protocol state and prevents silent data corruption going forward.

### DIFF
--- a/projects/bnpl-pay/index.js
+++ b/projects/bnpl-pay/index.js
@@ -1,29 +1,77 @@
 const { getUniqueAddresses } = require('../helper/utils')
 const { sumTokens2 } = require('../helper/unwrapLPs')
+
 const BNPL_FACTORY = '0x7edB0c8b428b97eA1Ca44ea9FCdA0835FBD88029'
 
+async function getNodes(api) {
+  return api.fetchList({
+    lengthAbi: abi.bankingNodeCount,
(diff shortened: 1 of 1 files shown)


ghost:
The adapter at projects/bnpl-pay exports TVL: 
        
 
 
    ethereum                  23.33 k
    ethereum-borrowed         0.00
    ethereum-staking          0.00
    borrowed                  0.00
    staking                   0.00
    
    total                    23.33 k

PR #17533, opened 2026-01-01
Updated the Genius Yield Adapter

4TT1L4:
Changed the v2 address to be using the non-shared variant (like the previous versions)


4TT1L4:
After talking once again with out blockchain engineers, it turned out that we could use a non-shared address like in the previous versions.

I have updated the adapter code accordingly.

PR #17568, opened 2026-01-06
Updated Script to dynamically fetch vault's

krishnafractol:


waynebruce0x:
`Error in sui: TypeError: Invalid URL` - does this work for you locally?

krishnafractol:
> `Error in sui: TypeError: Invalid URL` - does this work for you locally?

I tried running locally but due to RPC_PROXY_URL It's fails for me previously but works when deployed do you have any suggestion here.  how to test it locally

PR #17646, opened 2026-01-12
fix(ston): filter pools to only include those with core assets

nocell:
**Summary**

Filter STON.fi pools to only include those where at least one token is a known core asset (TON, USDT, USDC, etc.).

**Problem**

When spam tokens create pools like `SCAM1/SCAM2`, these pools:
1. Cannot be reliably priced — there's no core asset to derive a price from
2. May cause price lookup failures when the price server encounters unknown tokens

**Solution**

Added a filter that only includes pools where `token0` or `token1` is in the `coreAssets` list for TON chain. This ensures every pool has at least one token with a known price.

> TVL difference (~1M) comes from spam token pools that cannot be reliably priced anyway.

**Testing**

```bash
node test.js projects/ston/index.js
```
✅ Output: 39.16M TVL, no errors, no unknown tokens

### DIFF
--- a/projects/ston/index.js
+++ b/projects/ston/index.js
@@ -1,21 +1,24 @@
 const { get } = require('../helper/http')
 const { transformDexBalances } = require('../helper/portedTokens')
-const { getCoreAssets } = require('../helper/tokenMapping')
 
 module.exports = {
   misrepresentedTokens: true,
   timetravel: false,
   ton: {
     tvl: async () => {
-      const result = await get("https://api.ston.fi/v1/pools?dex_v2=true")
-      const coreTokens = new Set(getCoreAssets('ton'))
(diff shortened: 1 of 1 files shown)


ghost:
The adapter at projects/ston exports TVL: 
        
 
 
    ton                       39.13 M
    
    total                    39.13 M

waynebruce0x:
What is the objective here? If we price the tokens in our server then they should be included in TVLs

ghost:
The adapter at projects/ston exports TVL: 
        
 
 
    ton                       38.53 M
    
    total                    38.53 M

nocell:
That PR fix crash for scam/blacklisted tokens or tokens without value ( without `dex_usd_price` in STON.fi API ). Not need to manual exclude tokens.

PR #17667, opened 2026-01-13
Fix: Thena-v3

0xpeluche:
> The USD+ contract is no longer functional at all and reverts → blacklist


ghost:
The adapter at projects/thena-v3 exports TVL: 
        
 
 
    bsc                       1.85 M
    
    total                    1.85 M

PR #17740, opened 2026-01-20
update ea-finance tvl

ti-zzz:
1. WCC is a wrapped version of CC token (from Canton network) with a 1:1 peg. 
2. Although WCC is a token issued by our platform, it is generated by a Mint contract and is not manually controlled. We strictly control the 1:1 correspondence between Canton and BSC. We are simply providing users with a way to stake Canton CC on the BSC chain. Strictly speaking, users are staking Canton CC, which should be considered as TVL (Total Value Locked).

## Summary by CodeRabbit

* **Refactor**
  * Consolidated TVL (Total Value Locked) calculation for EA Finance to directly utilize staking logic, streamlining how metrics are computed and improving consistency across the module.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


ghost:
The adapter at projects/ea-finance exports TVL: 
        
 
 
    bsc                       7.70 k
    
    total                    7.70 k

PR #17761, opened 2026-01-21
fix(arkis): include AgreementV2Factory in TVL calculation

gabrielgrigoras-ops:
Arkis TVL on DefiLlama is currently missing agreements deployed via the AgreementV2Factory.

Root cause:
- Adapter only tracked AgreementFactory v1
- AgreementV2Factory was deployed on mainnet and emits the same AgreementCreated(address) event
- Agreements created via V2 were not included in TVL

Fix:
- Add AgreementV2Factory address (proxy) to adapter
- Use correct fromBlock for V2 deployment (23468237)
- Fetch AgreementCreated events from both v1 and v2 factories
- Merge agreement sets before token discovery and TVL calculation

This change does not affect Hyperliquid TVL logic.

## Summary by CodeRabbit

* **New Features**
  * Added support for an additional agreement version; TVL and borrowed totals now aggregate across multiple agreement types and margin accounts.

* **Bug Fixes**
  * Improved accuracy of aggregated totals by consistently combining multi-version data during log ingestion.

* **Style**
  * Minor formatting and cleanup in token processing and data handling.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/projects/arkis/index.js
+++ b/projects/arkis/index.js
@@ -45,7 +45,9 @@ const tokens_to_track = [
 
 const blacklist = [
   '0x8fccfd6404da2026eee7e4f529b45f3caaf0594e',
-  '0x4956b52ae2ff65d74ca2d61207523288e4528f96'
+  '0x4956b52ae2ff65d74ca2d61207523288e4528f96',
+  '0x7056ec39063c8eb094f5006c7e81f1fe2e7552ae',
+  '0x009848a093cb7991c1214e4655a6e8fc2f2da6cd',
 ]
 
 const fetchFactoryLogs = async (api, type, keyOverride) => {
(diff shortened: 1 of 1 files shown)


projects/arkis/index.js
```diff
@@ -6,9 +6,16 @@ const native = ADDRESSES.GAS_TOKEN_2;
...
 const factories = {
   agreement: "0xEA623eebd9c5bFd56067e36C89Db0C13e6c70ba8",
+  agreementV2: "0xe70d11D23F36826C58f30C61B4DeAf0A89a6D837",
   marginAccount: "0xbbC9c04348E093473C5b176Cb4b103fF706528bf"
 }
 
+const fromBlocks = {
+  agreement: 21069508,
+  agreementV2: 23468237,
+  marginAccount: 21069508,
```
danil-menkin: technically doesn't matter as that block is more old, but just for consistency 
```suggestion
  marginAccount: 21069559,
```

g1nt0ki (commented):
thanks for the PR

PR #17917, opened 2026-02-04
fix: use USDC_CIRCLE instead of USDC for LP pools and margin accounts

T0xicPotato:
fix: use USDC_CIRCLE instead of USDC for LP pools and margin accounts

## Summary by CodeRabbit

* **Refactor**
  * Restructured internal token and pool configurations to improve system organization. Updated mappings to support both legacy and current asset configurations.


ghost:
The adapter at projects/sharwa-finance exports TVL: 
        
 
 
    arbitrum                  49.23 k
    
    total                    49.23 k

PR #18051, opened 2026-02-16
Extend Sumtokens2 helper for historical support for solana

treeoflife2:
## Summary by CodeRabbit

* **New Features**
  * Retrieve historical token balances for dates before today via an Allium-based historical data path.
  * Automatic routing to the appropriate data path based on requested timestamp (historical vs current).
  * Improved balance retrieval with trusted-token filtering, cross-source token/account matching, and SOL vs SPL handling.
  * SQL-safe querying added and RPC-based flow preserved as a fallback.

### DIFF
--- a/projects/helper/solana.js
+++ b/projects/helper/solana.js
@@ -266,7 +266,6 @@ async function sumTokens2({
 
   // Route to Allium-based historical path when api.timestamp is before today (UTC)
   if (api && api.timestamp && (chain === 'solana' || !chain) && api.timestamp < getUniqStartOfTodayTimestamp()) {
-    sdk.log('sumTokens2: using historical Allium path for date', date);
     return sumTokens2_historical({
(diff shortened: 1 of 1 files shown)


projects/helper/solana.js
```diff
@@ -248,6 +263,19 @@ async function sumTokens2({
 }) {
 
   if (api) chain = api.chain
+
+  // Route to Allium-based historical path when api.timestamp is before today (UTC)
+  if (api.timestamp < getUniqStartOfTodayTimestamp()) {
```
g1nt0ki: api might always get passed, and we need to check if chain is empty or solana, dont think allium supports for the rest

also, can keep it simple like api.timestamp is over 12 hours ago

treeoflife2: > also, can keep it simple like api.timestamp is over 12 hours ago

The daily balances tables provide the daily end-of-day balance, so it's needed filter.

PR #18254, opened 2026-03-05
Connected convex pools to UFarm.Digital

evgen-povt:
## Summary by CodeRabbit

* **Bug Fixes**
  * Asset tracking now includes convex reward pools in account summaries for fuller visibility.
  * Enhanced token unwrapping to handle complex and layered assets, improving balance accuracy.
  * Improved asset aggregation for more comprehensive and reliable portfolio calculations.

* **Refactor**
  * Streamlined aggregation path by removing legacy logs-based approach.

* **Chores**
  * Public configuration updated to indicate double-counting behavior.

### DIFF
--- a/projects/ufarm-digital/index.js
+++ b/projects/ufarm-digital/index.js
@@ -22,6 +22,7 @@ const config = {
 
 module.exports = {
   methodology: 'Counts the AUM of all pools registered in the UFarm Protocol',
+  doublecounted: true,
 }
 
 Object.keys(config).forEach(chain => {


ghost:
The adapter at projects/ufarm-digital exports TVL: 
        
 
 
    ethereum                  935.99 k
    arbitrum                  86.30 k
    
    total                    1.02 M

ghost:
The adapter at projects/ufarm-digital exports TVL: 
        
 
 
    ethereum                  765.09 k
    arbitrum                  86.07 k
    
    total                    851.15 k

projects/ufarm-digital/index.js
RohanNero, line 24: can you please add `doublecounted: true` here since ufarm's tvl is tracking tokens staked into other protocols

RohanNero (changes requested):

projects/ufarm-digital/index.js
evgen-povt: Our TVL reflects the net assets under management within UFarm vaults and does not double count tokens.
Tokens deposited into external protocols are accounted for only once, based on their underlying value, and are not counted separately both as vault assets and as staked positions.
Internal tests also confirmed that the calculated TVL precisely matches the current on-chain TVL, which further verifies that no double counting occurs.
Therefore, doublecounted: true would not be accurate for our TVL calculation.

RohanNero: the `doublecounted` flag doesn't mean your tokens are counted twice within the adapter, but instead that the tokens are staked into another protocol (which tracks the tokens as their own inside their adapter). 

For example, if a user deposits $100 into Ufarm, which may deposit into Uniswap/convex/etc, the $100 is then being tracked for ufarm and the external protocol. Marking it with [doublecounted](https://docs.llama.fi/faqs/frequently-asked-questions#what-is-doublecount) won't affect the tvl displayed on your page, it just allows us to prevent counting the same assets twice for the aggregate tvl, e.g. by chain.

evgen-povt: got it, thank you for clarification. Added the flag.

ghost:
The adapter at projects/ufarm-digital exports TVL: 
        
 
 
    ethereum                  724.42 k
    arbitrum                  84.48 k
    
    total                    808.90 k

PR #18274, opened 2026-03-06
Fix: Beradrome

0xpeluche:


ghost:
The adapter at projects/beradrome exports TVL: 
        
 
 
    berachain-borrowed        2.99 M
    borrowed                  2.99 M
    berachain                 1.01 M
    
    total                    1.01 M

PR #18292, opened 2026-03-07
fix: update Shido DEX TVL to accurately track all 140+ pools

kuzo998:
Description: > This is a critical accuracy fix for the existing Shido DEX listing.

Reason for Update: > The previous implementation relied on RPC logs which hit search limits, causing the adapter to miss approximately 28 pools (tracking only 112 tokens instead of 140+). This resulted in underreported TVL.

New Methodology: > We have switched to a hybrid subgraph approach. We use the official subgraph to map the addresses of all 140+ factory-created pools and then fetch their real-time on-chain balances.

[x] Key Change: Enabled misrepresentedTokens: true to allow the production router to correctly price unlisted or isolated pairs (like ILCFNBR/CHINAWOK).

[x] Verified locally; successfully extracted raw balances for 140+ tokens.

## Summary by CodeRabbit

* **New Features**
  * Added Shido DEX integration that calculates and reports Total Value Locked (TVL). It fetches pool data from the DEX, aggregates token balances across pools, and exposes TVL results for inclusion in the platform’s overall metrics and dashboards. This enables monitoring of Shido liquidity and value composition alongside other integrations.

### DIFF
--- a/projects/shido-dex/index.js
+++ b/projects/shido-dex/index.js
@@ -14,12 +14,12 @@ async function tvl(api) {
   `;
   
   const response = await post(SUBGRAPH_URL, { query });
-  const pools = response.data.pools;
+  const pools = response?.data?.pools ?? [];
 
   const tokensAndOwners = [];
   pools.forEach(pool => {
-    tokensAndOwners.push([pool.token0.id, pool.id]);
-    tokensAndOwners.push([pool.token1.id, pool.id]);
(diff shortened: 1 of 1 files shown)


ghost:
The adapter at projects/shido-dex exports TVL: 
        
 
 
    shido                     676.03 k
    
    total                    676.03 k

ghost:
The adapter at projects/shido-dex exports TVL: 
        
 
 
    shido                     676.25 k
    
    total                    676.25 k

PR #18300, opened 2026-03-08
fix(sablier-legacy): normalize blacklisted tokens using getUniqueAddresses

zknpr:
## Summary
- Fixes token blacklist filtering by wrapping `blacklistedTokens` in `getUniqueAddresses()` at definition time
- Previously, case-sensitive comparison could miss blacklisted tokens (SAI address)
- Resolves existing TODO in the codebase

## Test plan
- [ ] Verify sablier-legacy TVL output is correct
- [ ] Confirm SAI tokens are properly excluded

## Summary by CodeRabbit

* **Refactor**
  * Updated address deduplication to use runtime-based generation instead of static configuration, maintaining the same functional behavior.


zknpr:

PR #18304, opened 2026-03-08
fix(dual): exclude non-existent reverse vaults from TVL

zknpr:
## Summary
- Fixes inflated TVL caused by including derived `reverseVault` PDA addresses that don't exist on-chain
- Verifies vault existence using `getMultipleAccounts` before including in TVL calculation
- Deduplicates token account fetching between `tvl()` and `staking()` into shared `getTokenAccounts()`

## Test plan
- [ ] Verify Dual TVL is no longer inflated
- [ ] Confirm only existing on-chain vaults are counted


zknpr:

PR #18315, opened 2026-03-08
perf(near): Optimize sumTokens with concurrent execution

zknpr:
## Summary
- Uses `sliceIntoChunks` and `Promise.all` (50 concurrent tasks) for NEAR RPC calls
- JSON-RPC batching is not supported by public NEAR RPCs, so concurrent individual calls is the right approach

## Test plan
- [ ] Verify NEAR adapter TVL is unchanged


zknpr:

PR #18407, opened 2026-03-11
fix(dorkfi): only suppress 404 in market box catch, rethrow other errors

mikepappalardo:
Addresses reviewer feedback on #18213.

The blanket `catch` was silently swallowing all errors including decode bugs and transient indexer failures, which would cause borrows to be silently under-reported.

This fix:
- Only suppresses HTTP 404 (box not yet initialised for a market)
- Re-throws all other errors so real failures surface

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for data fetching: not-found (404) cases are skipped while other fetch errors are surfaced.
  * Borrow calculations now occur only after successful data retrieval, preventing incorrect or partial results when a fetch fails.

### DIFF
--- a/projects/dorkfi/index.js
+++ b/projects/dorkfi/index.js
@@ -128,36 +128,39 @@ function makeBorrowed(chain) {
     const markets = await discoverAllMarkets(chain);
 
     for (const [marketAppId, poolAppId] of markets) {
+      // Only the box fetch may legitimately 404 (market not yet initialised).
+      // Keep all other logic outside the catch so errors from decoding or
+      // lookupAccount propagate instead of being silently swallowed.
+      let boxData;
       try {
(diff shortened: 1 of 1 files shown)


ghost:
The adapter at projects/dorkfi exports TVL: 
        
 
 
    algorand                  75.69 k
    voi                       24.11 k
    voi-borrowed              0.00
    algorand-borrowed         0.00
    borrowed                  0.00
    
    total                    99.80 k

ghost:
The adapter at projects/dorkfi exports TVL: 
        
 
 
    algorand                  75.70 k
    voi                       24.11 k
    voi-borrowed              0.00
    algorand-borrowed         0.00
    borrowed                  0.00
    
    total                    99.81 k

PR #18410, opened 2026-03-12
Align SoDEX TVL with ValueChain

ze-codes:
## Summary
- add `valuechain` as a supported chain in `chains.json`
- aggregate the existing SoDEX custody balances into a single `valuechain` TVL export so the chain follows the same custody-based methodology as SoDEX Bridge
- document that ValueChain TVL is broader than the spot/perp vault API and keeps SOSO balances in staking

## Test plan
- Ran `node test.js projects/sodex/index.js`
- Confirmed the adapter loads and executes the aggregated `valuechain` path
- Test currently fails on a Solana RPC timeout (`api.mainnet-beta.solana.com`), while the expected non-staking TVL remains aligned with the live `sodex-bridge` payload (~59.7M)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/projects/helper/chains.json
+++ b/projects/helper/chains.json
@@ -445,6 +445,7 @@
   "union",
   "unit0",
   "vana",
+  "valuechain",
   "vechain",
   "velas",
   "venom",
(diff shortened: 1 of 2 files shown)


ghost:
The adapter at projects/sodex exports TVL: 
        
 
 
    valuechain                59.69 M
    valuechain-staking        1.56 M
    staking                   1.56 M
    
    total                    59.69 M

PR #18462, opened 2026-03-17
Add YieldSeeker adapter

krishan711:
Adds TVL adapter for [YieldSeeker](https://yieldseeker.xyz/) on Base chain
Wallet addresses are fetched from the YieldSeeker API, token balances are read on-chain
ERC4626 vault share balances are converted to underlying asset amounts on-chain
Test output: $1.47M TVL

---

## Summary by CodeRabbit

* **New Features**
  * Added a public TVL entry for YieldSeeker to surface its balances.
  * End-to-end TVL calculation for YieldSeeker yield agents, aggregating on-chain token balances.
  * Automatic detection and normalization of ERC4626 vault shares into underlying asset values.
  * Curated vault allowlist used to improve accuracy of yield agent valuations.

### DIFF
--- a/projects/yieldseeker/yieldAgents.js
+++ b/projects/yieldseeker/yieldAgents.js
@@ -115,7 +115,7 @@ async function yieldAgentsTvl(api) {
       // Remove vault share balance and add equivalent underlying amount
       api.removeTokenBalance(vaultKey);
       const underlyingAmount = BigInt(shareBalance) * BigInt(totalAssets) / BigInt(totalSupply);
-      api.add(asset, underlyingAmount.toString());
+      api.add(asset.toLowerCase(), underlyingAmount.toString());
     }
   }
 


ghost:
The adapter at projects/yieldseeker exports TVL: 
        
 
 
    base                      1.47 M
    
ghost:
The adapter at projects/yieldseeker exports TVL: 
        
 
 
    base                      1.48 M
    
ghost:
        
 
ghost:
The adapter at projects/yieldseeker exports TVL: 
        
 
 
    base                      1.48 M
    
ghost:
The adapter at projects/yieldseeker exports TVL: 
        
 
 
    base                      1.48 M
    
ghost:
The adapter at projects/yieldseeker exports TVL: 
        
 
 
    base                      1.48 M
    
krishan711:
Pushed. The first finding was valid — fixed by lowercasing asset before api.add(). The nitpick about null-guarding normalizeToken I skipped since Object.keys() always returns strings.

ghost:
The adapter at projects/yieldseeker exports TVL: 
        
 
 
    base                      1.49 M
    
    total                    1.49 M

krishan711:
The agentWalletAddress field is a required non-nullable field in the API response schema — every record in agentAnalytics has a valid wallet address. Adding a guard here would be dead code. The snapshot.tokenBalances guard we added earlier is different because snapshot data can legitimately be incomplete for newly created agents, but an agent record without a wallet address would indicate a bug in the API itself, which we'd want to surface as an error rather than silently skip.

krishan711:
hey @RohanNero it's my first time adding to DefiLlama. Please do let me know if there's anything to update here, I'm around all day so can respond quickly 👍

projects/yieldseeker/index.js
```diff
@@ -0,0 +1,11 @@
+const { yieldAgentsTvl } = require('./yieldAgents');
+
+module.exports = {
+  methodology: 'Counts the tokens held across all YieldSeeker agent wallets. Wallet addresses are obtained from the YieldSeeker API, and token balances are read on-chain.',
```
RohanNero, line 4: hello, thank you for the PR. we typically don't allow tvl tracking from EOAs, could you clarify how they are used in yieldseeker? do you generate new wallets when users create agents? Are users able to deposit/withdraw from the agent wallets? Are users able to view their agent wallet private keys?

krishan711: hey, apologies for the late response here.
1. the EOAs are created using coinbase's server wallets and then ERC-7702 delegated onchain to a smart wallet so that our servers can operate on their behalf with a paymaster. the user can make calls directly too if they want but this rarely happens. 
2. yep, a new wallet is created whenever an agent is created. each agent has an EOA set up (just shy of 5k agents now) and they all operate autnomously.
3. users can send money directly to the agent wallet and withdraw by using the smart wallet, but usually they just do this through our ui cos it's much simpler.
4. users can't see the private keys - there are easy  operations to withdraw all funds and then the agent wallet basically doesnt get used again, but we've chosen not to allow exporting private keys because many of our users would probably do this accidentally or loose them etc, we've made the ui easy enough that nobody has needed this so far! (we could, of course, one day export the keys if we really wanted to)

RohanNero: no worries, thank you for clarifying

krishan711:
hey @RohanNero the docs say once the PR is merged in someone from the frontend team will integrate it into the defillama ui, but this hasnt happened yet. do i need to poke someone? sorry to tag here i know you must be super busy, but i couldnt see where else to try from the docs (https://docs.llama.fi/list-your-project/submit-a-project)

RohanNero:
@krishan711 Hey, no worries, thank you for letting me know, I will get it fixed.

PR #18483, opened 2026-03-19
Enable timetravel for Stacks adapters, update Zest v1 + v2 to use it

y0s0n:
### Problem

All Stacks adapters have `timetravel: false` because the shared helper `stacks-api.js` used `fetchCallReadOnlyFunction` from `@stacks/transactions`, which doesn't expose the `?tip=` query parameter. The Hiro API's `/v2/contracts/call-read/` endpoint supports `?tip={index_block_hash}` to query contract state at any historical block, but the SDK abstraction hides this.

Additionally, `@defillama/sdk`'s `getBlocks()` does not support Stacks, so `api.block` is always `undefined` for Stacks adapters - even when a historical timestamp is provided. This means there was no way for Stacks adapters to know what block height to query.

#### 1. `projects/helper/chain/stacks-api.js` - shared Stacks infrastructure

**Replaced `fetchCallReadOnlyFunction` with a direct HTTP POST.** The new implementation replicates the SDK's behavior exactly (same URL construction, same Clarity serialization via `cvToHex`, same response decoding via `hexToCV`) but adds an optional `block` parameter to `call()`.

The `block` parameter accepts either:
- A block height (integer) - internally resolved to `index_block_hash` via the Hiro extended API
- A raw block hash (64-char hex string) - used directly as the `?tip=` value

**Important detail about `?tip=`:** The Hiro API requires the `index_block_hash` field (not the regular `hash`) with the `0x` prefix stripped. Using the regular block hash or keeping the `0x` prefix causes the API to silently ignore the parameter and return current state. This was verified experimentally.

Block hash lookups are cached so multiple `call()` invocations at the same block height only hit the API once.

**Added `getBlockAtTimestamp(timestamp)`.** Since `api.block` is not populated for Stacks, adapters need to resolve timestamps to block heights themselves. This function uses interpolation search against a hardcoded table of 25 known `(height, timestamp)` anchor points derived from chain data.

How it works:
1. Find the two anchors bracketing the target timestamp (0 API calls)
2. Linear interpolate to estimate block height (0 API calls)
3. Fetch that block to verify (1 API call)
4. If off by more than 30 seconds, apply one correction using the local block rate derived from the bracket (0-1 API calls)

Total cost: **1-2 API calls per timestamp lookup**, cached for subsequent calls.

The anchor table includes dense coverage around the Nakamoto activation (block 171,355, Oct 25 2024) and the post-activation transition period where block times changed from ~600s to ~4s. Accuracy is within minutes for most of chain history. For timestamps beyond the last anchor, it fetches the latest block from the API as an upper bound.

#### 2. `projects/zest/api.js` - Zest V1

**Added pool-read version routing.** Zest V1's `pool-read` contract has been redeployed 16 times. Querying a historical block with a contract that didn't exist yet would fail. A version table maps deployment heights to contract addresses, and `getPoolRead(blockHeight)` selects the correct version for any given height.

**Added timetravel to `borrowed()`.** Resolves `api.block` or falls back to `getBlockAtTimestamp(api.timestamp)`. Each asset query is wrapped in try/catch because some assets were added after launch - querying them at a pre-listing block height returns an error, which is correctly treated as 0 borrows.

#### 3. `projects/zest-v2/api.js` - Zest V2

**Added timetravel to `borrowed()`.** Same `api.block ?? getBlockAtTimestamp(api.timestamp)` pattern. V2 vaults have not been redeployed so no version routing is needed.

#### 4. `projects/zest/index.js` and `projects/zest-v2/index.js`

Removed `timetravel: false`.

### Backwards compatibility

- When `api.block` is undefined and `api.timestamp` is not provided, `block` resolves to `undefined` and no `?tip=` is appended - identical to previous behavior
- The `call()` function signature is backwards compatible; existing callers that don't pass `block` are unaffected
- `getBlockAtTimestamp` is exported from `stacks-api.js` for use by any other Stacks adapter that wants to add timetravel support

### How to test

```bash
# Current data (no timetravel) - should match previous values
node test.js projects/zest/api.js
node test.js projects/zest-v2/api.js

# Historical data - pass a unix timestamp as second arg
node test.js projects/zest/api.js 1737586668
```

### Verified

- All 16 pool-read versions tested at historical block heights - 14 return distinct historical borrow values, 2 earliest correctly fail (wstx not yet listed, caught by try/catch)
- `getBlockAtTimestamp` tested across chain history from 2021 to 2026
- Current TVL and borrowed amounts match pre-change values (no regression)
- Historical contract reads return genuinely different values

**NOTE**

---

> - If you would like to add a `volume/fees/revenue` adapter please submit the PR [here](https://github.com/DefiLlama/adapters).

3. Please fill the form below **only if the PR is for listing a new protocol** else it can be ignored/replaced with reason/details about the PR
4. **For updating listing info** It is a different repo, you can find your listing in this file: https://github.com/DefiLlama/defillama-server/blob/master/defi/src/protocols/data2.ts, you can edit it there and put up a PR

---

## Summary by CodeRabbit

* **New Features**
  * Added block-timestamp lookup to query historical state and allow contract reads anchored to a specific block/timestamp.
  * Contract reads now accept an optional block/tip parameter.
  * Support for selecting contract versions based on block height.

* **Improvements**
  * Cached and interpolated block lookup for faster/resilient timestamp→block resolution, with sensible fallbacks.
  * Per-vault/asset fetches now continue on most errors while propagating rate-limit failures.

* **Chores**
  * Removed timetravel configuration flags from module exports.

### DIFF
--- a/projects/helper/chain/stacks-api.js
+++ b/projects/helper/chain/stacks-api.js
@@ -105,26 +105,25 @@ async function getBlockAtTimestamp(targetTimestamp) {
   let guess = Math.floor(lo.height + fraction * (hi.height - lo.height))
   guess = Math.max(lo.height, Math.min(hi.height, guess))
 
-  // Fetch and correct (1-2 calls)
-  const block = await fetchBlock(guess)
-  if (Math.abs(block.block_time - targetTimestamp) < 30) {
-    blockAtTimestampCache[targetTimestamp] = block.height
-    return block.height
+  // Fetch and correct (1-2 calls). Always return highest block at or before target.
+  let best = await fetchBlock(guess)
+
+  if (best.block_time > targetTimestamp || Math.abs(best.block_time - targetTimestamp) > 30) {
+    const localRate = (hi.time - lo.time) / (hi.height - lo.height)
+    const correction = Math.round((targetTimestamp - best.block_time) / localRate)
+    const corrected = Math.max(lo.height, Math.min(hi.height, guess + correction))
+    if (corrected !== guess) {
+      best = await fetchBlock(corrected)
+    }
   }
 
-  // One correction: use local block rate from anchors, clamped to bracket
-  const localRate = (hi.time - lo.time) / (hi.height - lo.height)
-  const correction = Math.round((targetTimestamp - block.block_time) / localRate)
-  const corrected = Math.max(lo.height, Math.min(hi.height, guess + correction))
-
-  if (corrected !== guess) {
-    const block2 = await fetchBlock(corrected)
-    blockAtTimestampCache[targetTimestamp] = block2.height
-    return block2.height
+  // Walk back if we landed after the target timestamp
+  if (best.block_time > targetTimestamp && best.height > 1) {
+    best = await fetchBlock(best.height - 1)
   }
 
-  blockAtTimestampCache[targetTimestamp] = block.height
-  return block.height
+  blockAtTimestampCache[targetTimestamp] = best.height
+  return best.height
 }
 
 function isBlockHash(value) {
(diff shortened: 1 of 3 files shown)


RohanNero:
thank you for the PR! I refactored slightly and merged in #18535

PR #18542, opened 2026-03-26
Trebleswap V2

SolarCloud14:
Good afternoon,

We previously added our project, TreableSwap. That was the v1 Algebra contract. We are now adding the v2 Algebra contract and would like to ask if you could include this new version in the existing project.

Here is the PR we submitted earlier:
https://github.com/DefiLlama/DefiLlama-Adapters/pull/13221

Additionally, we have a question regarding fees:

We would also like to address an issue regarding how our fees are currently reflected on DefiLlama.

At the moment, the reported data appears to be significantly inaccurate. For example, the platform shows approximately $73 in total fees for the entire year and $0 in revenue, which does not reflect our actual performance.

At a minimum, LP trading fees should be included in the calculation. Based on our internal data, the protocol has generated at least several hundred thousand dollars in fees over the past year. Additionally, the reported figures should ideally account for all relevant fee sources, including LP fees, routing fees (such as those from Algebra), as well as fees from integrations like 0x, Houdini, and cross-chain swaps.

For reference, our landing page currently displays total revenue of approximately $97,000. While the actual figure may be higher, this value should at least be consistently reflected on DefiLlama.

We have also noticed that our volume metrics have been inaccurate over the past two days.

We would appreciate your assistance in reviewing and correcting these discrepancies.

### DIFF
--- a/projects/trebleswap_v2/index.js
+++ b/projects/trebleswap_v2/index.js
@@ -1,0 +1,11 @@
+const { uniV3Export } = require("../helper/uniswapV3");
+
+const TREBLESWAP_ALGEBRA_FACTORY = "0x6e606Cf94A4DDc01aEed2Fce16d1b4f5B33e0A31"
+
+module.exports = uniV3Export({
+    base: {
+        factory: TREBLESWAP_ALGEBRA_FACTORY,
+        fromBlock: 39029383,
+        isAlgebra: true
+    }
+});


ghost:
The adapter at projects/trebleswap_v2 exports TVL: 
        
 
 
    base                      589.63 k
    
    total                    589.63 k

0xpeluche:
Thanks for the PR, I’m integrating the adapter into our consolidation registry for all Uniswap V3 / Algebra #18543

0xpeluche:
### Do not merge  ### 
> the adapter has already been merged. I’m leaving the PR open to continue looking into the issue

g1nt0ki:
tracking fees & volume here: https://github.com/DefiLlama/dimension-adapters/commit/53a6a398799803639a2c2e33fe29c62460f61a2c

PR #18547, opened 2026-03-26
Update the stellar-dex adapter to use coingecko pricing

taha-aziz:
This PR adjusts `stellar-dex`'s adapter to conform to using Defillama pricing, as noted through our usage of the api/functions and correctly adjusting the misrepresented tokens classification. The underlying logic that calculates tvl at its raw asset balance can be found in Stellar's public `tvl_agg` dbt model found [here](https://github.com/stellar/stellar-dbt-public/blob/master/models/marts/tvl/tvl_agg.sql).

**NOTE**

---

1. If you would like to add a `volume/fees/revenue` adapter please submit the PR [here](https://github.com/DefiLlama/dimension-adapters).

2. Once your adapter has been merged, it takes time to show on the UI. If more than 24 hours have passed, please let us know in Discord.
3. Sorry, We no longer accept fetch adapter for new projects, we prefer the tvl to computed from blockchain data, if you have trouble with creating a the adapter, please hop onto our discord, we are happy to assist you.
4. **For updating listing info** It is a different repo, you can find your listing in [this file](https://github.com/DefiLlama/defillama-server/blob/master/defi/src/protocols/data2.ts), you can edit it there and put up a PR
5. Please do not add new npm dependencies, do not edit/push `pnpm-lock.yaml` file as part of your changes

---
## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected misrepresented tokens classification (flag set to false)

* **Refactor**
  * Redesigned TVL calculation to aggregate per-asset balances, select rows by timestamp, and resolve prices via multiple on-chain/off-chain sources with intelligent fallbacks
  * Improved stablecoin price handling for current timestamps and applied asset-specific price scaling

* **Documentation**
  * Updated methodology text describing the new approach

### DIFF
--- a/projects/stellar-dex/index.js
+++ b/projects/stellar-dex/index.js
@@ -1,37 +1,270 @@
-const utils = require('../helper/utils');
-const { getApiTvl } = require('../helper/historicalApi');
+const { fetchURL } = require('../helper/utils');
 
-const TVL_URL = "https://storage.googleapis.com/defillama-stellar-tvl/stellar-tvl.json";
+const TVL_URL = 'https://storage.googleapis.com/defillama-stellar-tvl/tvl_agg.json';
 
-async function fetchTvlData() {
-  const response = await utils.fetchURL(TVL_URL, { responseType: 'text' });
-  const lines = response.data.trim().split('\n');
-  return lines.map(line => JSON.parse(line));
+// Mapping of Stellar asset codes to CoinGecko IDs
+const COINGECKO_MAPPINGS = {
+  AQUA: 'aquarius',
+  XLM: 'stellar',
+  USDC: 'usd-coin',
+  EURC: 'euro-coin',
+  BTC: 'bitcoin',
+  ETH: 'ethereum',
+  USDT: 'tether',
+  PYUSD: 'paypal-usd',
+  EURS: 'stasis-eurs',
+  VELO: 'velo',
+  TFT: 'threefold-token',
+  MOBI: 'mobius',
+  RIO: 'realio-network',
+  GYEN: 'gyen',
+  XSGD: 'xsgd',
+  TRYB: 'bilira',
+  XCHF: 'cryptofranc',
+  IDRT: 'rupiah-token',
+  QCAD: 'cad-coin',
+  AUDD: 'novatti-australian-digital-dollar',
+  SolvBTC: 'solv-btc',
+  USDY: 'ondo-us-dollar-yield',
+  USDM1: 'mountain-protocol-usdm',
+  VCHF: 'vnx-swiss-franc',
+  VEUR: 'vnx-euro',
+  LSP: 'lumenswap',
+  XXA: 'ixinium',
+  AFR: 'afreum',
+  CRDT: 'crdt',
+  // Bridged assets — map to underlying
+  abUSDC: 'usd-coin',
+  abUSDT: 'tether',
+  abBUSD: 'binance-usd',
+  abBNB: 'binancecoin',
+  aeUSDC: 'usd-coin',
+  aeUSDT: 'tether',
+  aeDAI: 'dai',
+  aeETH: 'ethereum',
+  apUSDC: 'usd-coin',
+  apUSDT: 'tether',
+  apMATIC: 'polygon-ecosystem-token',
+  asUSDC: 'usd-coin',
+  asSOL: 'solana',
+  acUSD: 'usd-coin',
+  acCELO: 'celo',
+  // Etherfuse stablebonds
+  USTRY: 'etherfuse-ustry',
+  TESOURO: 'etherfuse-tesouro',
+  // Stellar-native tokens
+  SSLX: 'starslax',
+  SCOP: 'scopuly',
+  // Yield-bearing tokens — map to underlying
+  yBTC: 'bitcoin',
+  yETH: 'ethereum',
+  yXLM: 'stellar',
+  yUSDC: 'usd-coin',
+  xSolvBTC: 'solv-btc',
+  BTCLN: 'bitcoin',
+  ARST: 'argentine-peso',
+  ARS: 'argentine-peso',
+  BENJI: 'franklin-templeton-benji',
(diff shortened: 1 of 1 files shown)


ghost:
The adapter at projects/stellar-dex exports TVL: 
        
 
 
    stellar                   16.09 M
    
    total                    16.09 M

ghost:
The adapter at projects/stellar-dex exports TVL: 
        
 
 
    stellar                   16.09 M
    
    total                    16.09 M

RohanNero:
thank you for the PR, but we typically don't like to include any pricing in the tvl adapters so the original approach is preferred here

taha-aziz:
> thank you for the PR, but we typically don't like to include any pricing in the tvl adapters so the original approach is preferred here

Hello @RohanNero, could you please elaborate on this? The original approach does include pricing and was grandfathered in, but this new adapter should conform to the standards of DefiLlama much more. You can notice this through the usage of the functions seen on many other adapters as well as misrepresented_tokens now being set to False from what was True. We are providing raw asset balances and using DefiLlama to do pricing.

RohanNero:
@taha-aziz Hello, we prefer to exclude in-adapter pricing (calling the DL price api in this case) and the hard-coded token mappings when possible. I had attempted to [refactor](https://github.com/DefiLlama/DefiLlama-Adapters/compare/main...RohanNero:DefiLlama-Adapters:stellar-dex-refactor) using raw assets after adding pricing on the server for missing tokens, but the tvl_agg endpoint is showing `2026-03-25` as the date. 

Do you know how we can fetch the list of tokens and contracts without using the `TVL_URL`?

taha-aziz:
@RohanNero, so the tvl_agg endpoint is a temporary one made while we are doing this switch. The prior one was updated daily but in USD and we are doing this adapter switch as you know. The `tvl-agg` one was only updated as of `2026-03-25` so that is correct actually. Our endpoint will be updated daily the same time this PR goes through. We should be good to keep fetching the list of tokens using this pre-established `TVL_URL`.

Knowing this, can we move forward with getting this adapter in?

RohanNero:
Thank you for the clarification, I will discuss this with the team today

RohanNero:
@taha-aziz Thanks again for the PR, I refactored and merged in #18680

taha-aziz:
@RohanNero can you revert your change? The pr is using the temporary file endpoint that would not work long term. See my comment [here](https://github.com/DefiLlama/DefiLlama-Adapters/pull/18547#issuecomment-4193608682) 

Also, your reported tvl output is 6 million higher

RohanNero:
@taha-aziz Hello, what is the correct URL? Will stellar-dex be updated to match the format of tvl_agg? Also, when testing this PR locally, it returns 22.76M currently. So is the stellar-dex's reported 15M for april 6th correct or is it closer to the refactored PRs (~22M)?

taha-aziz:
@RohanNero the correct url would be the stellar-tvl file which would be updated to match the format of tvl_agg. Yes, so this PR will also return the incorrect tvl as of right now because the file now includes AMM/liquidity_pools_tvl. There will be a second adapter PR that comes through that will report those numbers. Essentially total_tvl is adding up everything, but really what we need to be pulling from the file is the `accounts_tvl + trustlines_tvl` for just dex not `total_tvl`. The reported 15m is what it should be.

We also want to ask how the pricing works in the refactor. We noticed that contract_id gets pulled and used in the `tvl` function. Could you please elaborate on that?

RohanNero:
@taha-aziz So the "extra" 6M is from the AMM/liquidity pools, and you now want to exclude counting that here and instead move it to a second adapter? Why split the accounts/trustlines tvl from the amm/liquidity pools? 

I mapped stellar assets to their soroban contract addresses so they could be added to api directly (some are priced in server by soroban contract only)

taha-aziz:
@RohanNero So the AMM/LP portion is still in development which is why it is in the test file for now. I will get back to the team and confirm if we would like to go with the single or second adapter approach as well.

Additionally, we can provide the contract_ids for the assets (I will get that included so we don't have to hit the Horizon endpoint). We would like to know how the backend figures out the stellar asset code/issuer/contract_ids to pricing though if you can give us more details here!

RohanNero:
@taha-aziz Is the 6M being tracked in the AMM/LP real tokens, or is it only testing values? If it is real, could we continue tracking it as it is for now, then you can make a new PR to split the adapter into two in the future if you guys decide? (I can point the api back to stellar-dex, but it would need to be updated first)

That would be perfect. Alternatively, we can add server pricing for assets by their stellar id in addition to their soroban contract if necessary/desired. The [defillama-server](https://github.com/DefiLlama/defillama-server/tree/master) repo is where all of the tokens are priced, for the stellar ones specifically you can see this [tokenMapping.json](https://github.com/DefiLlama/defillama-server/blob/master/coins/src/adapters/tokenMapping.json#L11234) file (line 11,234).

taha-aziz:
@RohanNero can we still please revert the PR? we want to review the numbers internally further before we merge the refactored version in given the change of not using our mapping and other factors. We can re-visit the one or two adapters then.

What would be the process of adding an asset on to that tokenMapping.json file? What if the asset is not found in coingecko, but found in stellar-expert - what would that process be as well?

RohanNero:
@taha-aziz sure. To add pricing for an asset you can open a PR in the server repo in the same way you opened one here, if the token isnt listed on coingecko you'll need to either add an adapter to retrieve the price onchain or from another source (api)

taha-aziz:
@RohanNero thanks for reverting! will the dashboard here be fixed/reverted to the old tvl numbers for today? https://defillama.com/protocol/stellar-dex

RohanNero:
@taha-aziz yes, I will update it so it doesn't show the spike to 20M

taha-aziz:
Please see here for refactored and new PR: https://github.com/DefiLlama/DefiLlama-Adapters/pull/18878

Note: this includes an amm adapter that stellar is now owning and maintaining as we phase out of LumenSwap's adapter

cc: @RohanNero

PR #18438, opened 2026-03-16
Add Alto to DefiLlama

0xshippor:
##### Name (to be shown on DefiLlama):
Alto

##### Twitter Link:
https://x.com/alto_money

##### List of audit links if any:
https://github.com/altomoney/security-review-engagements

##### Website Link:
https://www.altofoundation.org/

##### Logo (High resolution, will be shown with rounded borders):
https://ogqctdyovgvwilkqwdxv.supabase.co/storage/v1/object/public/assets/media/defillama/alto_logo.png

##### Current TVL:
Our stablecoin DUSD is not priced yet via the API but available on [coingecko](https://www.coingecko.com/en/coins/alto-dusd). Once DUSD is priced (we assume that this is after https://github.com/DefiLlama/peggedassets-server/pull/739 is merged) the current TVL should be following
Total TVL	~$468.8k
Total Borrowed ~$131.5k

##### Treasury Addresses (if the protocol has treasury)
`0xA1148A1b94262540994Cf9Aa431A13ad39764228`

##### Chain:
Ethereum

##### Coingecko ID (so your TVL can appear on Coingecko, leave empty if not listed): (https://api.coingecko.com/api/v3/coins/list)
\- (ALTO token has not launched yet)

##### Coinmarketcap ID (so your TVL can appear on Coinmarketcap, leave empty if not listed): (https://api.coinmarketcap.com/data-api/v3/map/all?listing_status=active,inactive,untracked&start=1&limit=10000)
\- (ALTO token has not launched yet)

##### Short Description (to be shown on DefiLlama):
Alto is a decentralized credit protocol built around DUSD, a collateral-backed stablecoin. Users can borrow, lend, mint, and leverage DUSD using assets like ETH, LSTs, and tBTC as collateral across isolated markets.

##### Token address and ticker if any:
\- (ALTO token has not launched yet)

##### Category (full list at https://defillama.com/categories) \*Please choose only one:
CDP

##### Oracle Provider(s): Specify the oracle(s) used (e.g., Chainlink, Band, API3, TWAP, etc.):
Chainlink, combined with ratio-based pricing for LSTs and yield-bearing tokens

##### Implementation Details: Briefly describe how the oracle is integrated into your project:
Chainlink price feeds are used for base asset pricing. For LSTs and yield-bearing tokens, on-chain exchange rates are combined with Chainlink feeds to derive accurate collateral values.

##### Documentation/Proof: Provide links to documentation or any other resources that verify the oracle's usage:
[...]

PR #18559, opened 2026-03-27
add alto (refactor)

RohanNero:
resolves #18438 

## New Features

* Added Alto protocol integration with market and liquidity position tracking on Ethereum.
* Enabled Alto treasury data monitoring for Ethereum network.

### DIFF
--- a/projects/treasury/alto.js
+++ b/projects/treasury/alto.js
@@ -1,0 +1,5 @@
+const { treasuryExports } = require("../helper/treasury");
+
+module.exports = treasuryExports({
+    ethereum: { owners: ['0xA1148A1b94262540994Cf9Aa431A13ad39764228'] },
+  })


ghost:
The adapter at projects/alto exports TVL: 
        
 
 
    ethereum                  288.70 k
    
    total                    288.70 k

parth-15:
Hey @RohanNero , this is not entirely correct. We have two types of markets: Mint market and Borrow market.

The thing that you mentioned is correct for mint market. The DUSD minted from mint markets is backed by underlying collateral. However, borrow markets work similar to morpho markets where loan token is DUSD which is supplied by suppliers and borrowed by borrowers. The borrowers deposit collateral token so that they can borrow DUSD. So, we should count DUSD borrowed from borrow markets because they aren't backed by collateral.

Since this PR is merged, what's the best way to update it now?

RohanNero:
@parth-15 If the borrowed market allows users to deposit and borrow DUSD, then the collateral would still be double-counted from the mint market collateral since that is the DUSD backing. For future updates, you can open another PR to edit the existing alto adapter.

0xshippor:
@RohanNero 

Thanks for the review. I want to clarify how Alto's markets work, as there might be a misunderstanding about the double-counting concern.

Alto has two separate and independent market systems:

**Borrow markets** - traditional isolated-pair lending markets with borrowers and lenders. The contract holds both collateral (from borrowers) and currently available DUSD (deposited by lenders).

**Mint markets** - these also function as isolated-pair markets, **but there are no lenders**. When a user borrows, new DUSD is minted under the hood (and burned on repay). The contract **only holds collateral and has zero DUSD balance**, since DUSD is minted on-demand.

These are completely separate pools with separate collateral - depositing collateral into a mint market and depositing collateral into a borrow market are independent actions by different users.

Counting `borrowToken` balances via `balanceOf` doesn't double-count here because mint market contracts hold 0 DUSD, and the DUSD sitting in borrow markets was deposited by lenders who may have acquired it externally (e.g. via Uniswap or the PSM), not necessarily from mint markets.

Could we also add back the borrowed metric? It reports total DUSD debt outstanding across all markets and is independent of TVL. This is a standard metric that lending protocols like Aave and Morpho report, and it provides useful context for users.

RohanNero:
@0xshippor Thank you for the clarification, but the backing for DUSD comes from a mix of the USDC deposited into the PSM and collateral deposited into mint markets right? (both tracked as tvl currently) So even if a user acquired DUSD externally, e.g. uniswap, wouldn't it still be backed by one of these two sets of collateral?

0xshippor:
@RohanNero That's a fair point for DUSD minted through the permissionless PSM and mint markets. One additional nuance is that not all DUSD originates from sources currently tracked in TVL. We also have a permission**ed** PSM, only accessible by the protocol team, that minted DUSD 1:1 against USDC to seed the Uniswap USDC/DUSD pool. That DUSD is circulating, but its backing is not counted in TVL because the permission**ed** PSM is intentionally excluded from the adapter as it is not publicly accessible.

I just wanted to share that context so the decision on whether markets' borrow token balances should count toward TVL can be made with the full picture, because including USDC from the permission**ed** PSM in TVL would not feel right either.

0xshippor:
@RohanNero just a bump on this. If we are to create new PR we need to first understand what's the right way to implement the TVL for our protocol.

PR #18752, opened 2026-04-14
prism

waynebruce0x:
New OpenEden Product 

## New Features
* Added Ethereum TVL (Total Value Locked) calculation for OpenEden Prism that aggregates token balances and DeFi protocol positions from integrated data sources, along with off-chain reserve information for comprehensive TVL reporting.


g1nt0ki:
I have fixed it now

PR #18765, opened 2026-04-15
Fix Aquarius Stellar TVL adapter

wild-feather:
## Summary
Fix Aquarius Stellar TVL adapter.

## Changes
- Use UTC date keys
- Add fallback to nearest available row
- Make fetch cache retry-safe

## Validation
- Confirmed source API still returns daily TVL data
- Smoke-tested row selection for the current date

Please enable "Allow edits by maintainers".

## Summary by CodeRabbit

* **Bug Fixes**
  * TVL lookup now selects the closest available date when exact matches are missing.
  * Clearer errors when TVL data is missing or invalid.
  * TVL values normalized to consistent reported units.
  * Consistent UTC date handling for TVL records.

* **Performance**
  * Concurrent TVL fetches are consolidated to avoid redundant network requests and allow retries on failure.

### DIFF
--- a/projects/aqua-network/index.js
+++ b/projects/aqua-network/index.js
@@ -35,10 +35,15 @@ function pickClosestRow(rows, targetDate) {
   const exactMatch = sortedRows.find((row) => row.date === targetDate)
   if (exactMatch) return exactMatch
 
-  const priorRows = sortedRows.filter((row) => row.date < targetDate)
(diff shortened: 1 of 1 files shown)


RohanNero:
Sorry, closing because we prefer to throw errors if the current timestamp isn't available (fallback to older date would suppress failure)

PR #3162, opened 2022-07-04
Update DomFi token addresses with XYZDOM-JUN40 series on Boba

jaxrtech:
This updates the Domination Finance adapter to include newly deployed addresses on Boba.
The methodology and protocol information is the same as originally submitted.

PR #18796, opened 2026-04-17
update domination finance tvl adapter

bambotech:
this pull request updates the domination finance tvl adapter to reflect the latest deployment state on base mainnet

(updated from #3162)

## Summary by CodeRabbit

* **Refactor**
  * Removed TVL tracking support for Ethereum, Polygon, and Boba networks
  * TVL calculation now exclusively supports the Base network
  * Updated implementation approach from dynamic runtime logic to static configuration

### DIFF
--- a/projects/domfi/index.js
+++ b/projects/domfi/index.js
@@ -1,46 +1,19 @@
-const lsps = {
-  ethereum: [
-    // BTCDOM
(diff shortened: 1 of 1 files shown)


0xpeluche:
Thanks for the PR. I slightly updated the code so we can keep serving data for the other chains, since there still seem to be collateral assets locked in the contracts

PR #18797, opened 2026-04-17
updated travessia tvl with dynamic vault fetching

RedVeil:
Update for the Travessia Credit TVL adapter.
The adapter now fetches vault addresses dynamically from our API to no longer need manual PR updates to add / remove vaults.

## Summary by CodeRabbit

* **Refactor**
  * Vault addresses are now retrieved dynamically from a runtime configuration service instead of being hardcoded. This enables TVL calculations to use up-to-date vault lists fetched at runtime, improving accuracy and adaptability without changing public interfaces.


0xpeluche:
Hi, thanks for the PR. When the data is available on-chain, we try as much as possible to avoid using external APIs in order to make long-term maintenance easier. From a quick look at the Ethereum contracts, I found 6 vaults via a factory, so would it be possible to prioritize that approach instead? Also, out of those 6 vaults, only 2 are active according to your endpoint on ethereum. What about the others? There seems to be USDC on `0xb37cC5e891d87c5a3eF750e1bdf59C960575593c`

RedVeil:
> Hi, thanks for the PR. When the data is available on-chain, we try as much as possible to avoid using external APIs in order to make long-term maintenance easier. From a quick look at the Ethereum contracts, I found 6 vaults via a factory, so would it be possible to prioritize that approach instead? Also, out of those 6 vaults, only 2 are active according to your endpoint on ethereum. What about the others? There seems to be USDC on `0xb37cC5e891d87c5a3eF750e1bdf59C960575593c`

We are using Accountable for our vault tech. So the factory will show you plenty more vaults from other managers but us. Thats why i added the API since its not possible to distinguish them from the factory alone

PR #18811, opened 2026-04-19
fix "borrowed" to use the amount borrowed by borrowers

bunbun295:
Can you also replace all the existing audit links with

https://twyne.gitbook.io/twyne/resources/security

## Summary by CodeRabbit

* **Chores**
  * Updated borrowed amount calculation methodology to use collateral vault data sources, replacing the previous calculation approach.
  * Updated documentation to reflect the new data sourcing method.

### DIFF
--- a/projects/twyne/index.js
+++ b/projects/twyne/index.js
@@ -91,17 +91,27 @@ async function tvl(api) {
 }
 
 async function borrowed(api) {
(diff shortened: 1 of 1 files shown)


bunbun295:
@g1nt0ki is is possible to apply the new definition of borrowed retroactively? https://defillama.com/protocol/twyne applies the older "borrowed" definition till April 18.

PR #18859, opened 2026-04-22
Add Streamlock TVL adapter

0xNike:
**NOTE**

#### Please enable "Allow edits by maintainers" while putting up the PR.

---

##### Name (to be shown on DefiLlama):
Streamlock

##### Twitter Link:
https://twitter.com/0xHao_

##### List of audit links if any:
_None yet — internal security review only. Audit planned post-listing; will update via follow-up PR once available._

##### Website Link:
https://www.streamlock.fun/

##### Logo (High resolution, will be shown with rounded borders):
Attached in this PR at `projects/streamlock/streamlock.png` (1000×1000 square, white glyph on black circle — works on both light and dark backgrounds). A 240×240 JPG variant will be submitted to `defillama-server` as a follow-up per standard onboarding flow.

##### Current TVL:
~$190 USD (main `tvl` bucket, from `node test.js projects/streamlock/index.js` at submission time). Vesting bucket (reported separately under the `vesting` key, not rolled into main TVL): ~$7.3K USD of $LOCK across three team/treasury Streamflow streams. Small TVL reflects recent mainnet launch — protocol is in early-adoption phase.

##### Treasury Addresses (if the protocol has treasury)
- Protocol fee recipient (mainnet router authority): `8wk8NQc8hQ66R79Lx5GhKVgcPDpt2HXcUBeqH7xHCywt`
- Program upgrade authority: `HLWfdZTmHoABcEPy4bNjhZZmWTy4PGGd5CnrDSqn5CCr`

##### Chain:
Solana

##### Coingecko ID (so your TVL can appear on Coingecko, leave empty if not listed):
streamlock

##### Coinmarketcap ID (so your TVL can appear on Coinmarketcap, leave empty if not listed):
_Not listed on CoinMarketCap at the time of this PR._

##### Short Description (to be shown on DefiLlama):
Solana token launchpad with price-gated locks. Users buy launched memecoins via a bonding-curve AMM; purchases sit locked in on-chain escrows until each token hits its configured price milestones. Idle SOL reserves are staked as JitoSOL for yield that accrues to the protocol.

##### Token address and ticker if any:
`FNhcY1cwQvQqaM8CUjXSuoGKJniwC4maBRLqNRLipump` ($LOCK, Token-2022)

##### Category (full list at https://defillama.com/categories) *Please choose only one:
Launchpad

##### Oracle Provider(s):
[...]

PR #18874, opened 2026-04-23
add streamlock (refactor)

RohanNero:
resolves #18859

- `PoolConfig` size changed from 401 -> 409 (8 bytes added after original PR creation)

## Summary by CodeRabbit

* **New Features**
  * Added TVL tracking for Streamlock, enabling monitoring of the protocol's total value locked through native SOL and liquid staking token balances.

### DIFF
--- a/projects/streamlock/index.js
+++ b/projects/streamlock/index.js
@@ -1,0 +1,62 @@
+const { getConnection, sumTokens2 } = require("../helper/solana");
+const { PublicKey } = require("@solana/web3.js");
+
+// Streamlock Token Factory program — owns every pool_config and sol_vault PDA
+// created by the launchpad. https://www.streamlock.fun/
+const TOKEN_FACTORY_PROGRAM_ID = new PublicKey(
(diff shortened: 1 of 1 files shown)


0xNike:
Thanks for the quick turnaround @RohanNero! Confirmed the fix is correct:

1. The 8 new bytes are a single `first_milestone_hit_ts: i64` appended as the final field of `PoolConfig` — all prior offsets are preserved, so `SOL_VAULT_OFFSET = 40` and `dataSize: 409` are both right.
2. JitoSOL is held in an ATA owned by the `sol_vault` PDA (see the `lst_vault` field's docs in our IDL: *"ATA holding LST tokens, owned by sol_vault PDA"*), so `sumTokens2({ owners: solVaults, solOwners: solVaults })` captures it correctly. This also upgrades us from principal-SOL to mark-to-market valuation — a welcome improvement.
3. Dropping the Streamflow vesting bucket is fine — aligns better with DefiLlama's "don't count un-issued/vesting tokens" policy anyway, and Streamflow's own adapter covers it.

LGTM, happy for this to merge.

PR #18716, opened 2026-04-10
add monad vaults for VII Finance

kankodu:
## Summary by CodeRabbit

* **New Features**
  * Added support for the monad blockchain network. Users can now track Total Value Locked (TVL) metrics on monad alongside existing supported networks.

PR #18889, opened 2026-04-25
fix: VII Finance TVL for Monad

kankodu:
@waynebruce0x

We relied on events emitted by `GenericVaultFactory` from Euler and `UniswapV4WrapperFactory` from VII Finance to get the Euler vaults and UniswapV4Wrappers created by the VII Finance deployer. That was failing in the case of Monad. So, for Monad, we are using a static list of vaults and wrappers to get the TVL. This will fix the problem discussed [here](https://github.com/DefiLlama/DefiLlama-Adapters/pull/18716#issuecomment-4288952243).

## Summary by CodeRabbit

* **New Features**
  * Enhanced Monad support: improved Uniswap V4 NFT unwrap handling and default endpoint resolution for Monad chains.
  * Improved Monad TVL calculation: now aggregates ERC‑4626 vault assets and Uniswap V4 collateral in parallel for more accurate totals.


g1nt0ki:
thanks for the PR

kankodu:
Note:


> The adapter at projects/vii-finance exports TVL:
> 
> ```
> monad                     7.35 k
> unichain                  2.10 k
> ethereum                  1.66 k
> 
> total                    11.10 k 
> ```

Look at the CI output and look at the actual TVL after this PR was merged. The difference is because CI doesn't add uniswap v4 positions likely because subgraph is query fails in case of CI. The actual deployed version works correctly. @g1nt0ki @waynebruce0x 


image

PR #18894, opened 2026-04-27
aegis

walkermoz:
Can we update our Category from Liquidity Management to Lending, our Aegis V2 Hook is a lending based protocol.

## Summary by CodeRabbit

* **New Features**
  * Added total value locked (TVL) tracking for Aegis on the Unichain network, providing visibility into assets deposited in Aegis hooks on Uniswap v4.


RohanNero:
@walkermoz Hello, thanks for the PR. We prefer to only categorize a listing as "Lending" if users can borrow directly from the protocol.

PR #18896, opened 2026-04-27
Add ethereum chain support to goldenasset adapter

eric-volz:
## Description

This change extends the `goldenasset` adapter to support both `sei` and `ethereum` TVL calculation paths.

### What changed

- Added an `ethereum` chain export alongside the existing `sei` export.
- Refactored the duplicated TVL logic into a shared `tvl(api, tokenAddress)` helper.
- Introduced a `TOKEN_ADDRESSES` map to keep chain-specific token contract addresses in one place.

## Summary by CodeRabbit

* **New Features**
  * Added Ethereum TVL calculation support alongside existing SEI network configuration.

* **Refactor**
  * Updated TVL calculation implementation with historical price lookup and centralized token address configuration.

### DIFF
--- a/projects/goldenasset/index.js
+++ b/projects/goldenasset/index.js
@@ -2,45 +2,54 @@ const sdk = require('@defillama/sdk');
 const { fetchURL } = require('../helper/utils');
 
 const PRICE_DISCOUNT = 0.098;
+const TOKEN_ADDRESSES = {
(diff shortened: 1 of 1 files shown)


RohanNero:
Hello, thank you for the PR. We have recently changed how we list RWA tokens so this tvl adapter is no longer needed. The new listing for goldenasset can be seen on the RWA dashboard: https://defillama.com/rwa/asset/GEM

I will add the missing ethereum address to the listing.

PR #18918, opened 2026-04-28
Add sanity checks to PR validation workflow

DeFiDonkey:
## Why

The existing `Test_Change` workflow runs `node test.js` on every changed adapter and posts the TVL output as a PR comment via `commentResult.js`. It correctly catches adapters that *throw*, but it doesn't sanity-check the *value* the adapter computes. A few well-known shipping-failure modes therefore pass review silently:

- An adapter that runs cleanly but computes **`$0`** — almost always means the supported tokens have no price feed on `coins.llama.fi`. Very common when a new chain or bridged stablecoin needs a `tokenMapping.js` remap. (I hit this myself shipping a recent adapter on a fresh chain — the test passed, the value was zero, and there was no signal pointing at the fix.)
- An adapter that returns a **wildly inflated number** (>$50B) — usually a decimals bug, the same balance counted at two owners (vault + LP), reading `totalSupply` instead of `balanceOf`, or pricing a non-stable as USDC.
- A new listing without a `methodology` string export — this is a PR-template requirement and leaves the methodology blank on the UI.
- A probable **fetch adapter** (imports `axios` / `node-fetch` and reads a non-RPC HTTP URL) — explicitly banned for new projects per `pull_request_template.md`.

You can see the cost of these slipping through in the merge log: a steady cadence of follow-up PRs like *"paragon genuine spike"*, *"fix superform spike"*, *"rubic bridge aggr genuine spike"*, *"temp disable standx dusd fees"*. Each one is a day of maintainer time after the fact.

## What this PR does

Adds `.github/workflows/sanityCheck.js`. It consumes the same `output.txt` that `commentResult.js` already produces, plus the adapter source file path that `test.yml` already passes, and writes a follow-on markdown comment whenever any of the four signals fire.

**Warnings, not blockers** — the workflow's existing `ERROR` gate is untouched, the script never throws, and on any internal failure it exits silently. The maintainer always decides.

The integration is one line in `test.yml`:

```yaml
node ${{ github.workspace }}/test.js ...
node ${{ github.workspace }}/.github/workflows/commentResult.js ...
+ node ${{ github.workspace }}/.github/workflows/sanityCheck.js ${{ github.workspace }}/output.txt ${{ github.workspace }}/pr-comments ${i}
if grep -q "...ERROR..." output.txt; then exit 1; fi
```

The sanity comment file is named `z-sanity-<timestamp>-<pid>-<slug>.md` so it sorts after the existing TVL summary and reads as a follow-on note in the PR.

## Behaviour preserved

- The existing `ERROR` gate still fails the workflow on a thrown adapter.
- The existing `commentResult.js` output is unchanged.
- A healthy adapter produces zero warnings — verified locally against a real-world adapter (~$659K TVL, all conventions met).
- If the new script itself crashes for any reason (parsing edge case, file system hiccup, etc.) it logs to stderr and exits without affecting the rest of the pipeline.

## Smoke tests run locally

| Scenario | Expected | Result |
|---|---|---|
| Healthy adapter, $659K TVL, methodology present | no warning | ✅ no comment file produced |
| Adapter producing `$0` total | unpriced-tokens warning with hint about `tokenMapping.js` | ✅ |
| Adapter producing `$200B` total | high-TVL warning citing decimals / double-counting / pricing | ✅ |
| Adapter without `methodology` export | reminder to add a methodology string | ✅ |
| Adapter that errored out | suppressed (handled by `commentResult.js`) | ✅ no comment file produced |

## Files

- `.github/workflows/sanityCheck.js` — new (160 LOC)
- `.github/workflows/test.yml` — one-line addition

## Open questions / notes

- The `$50B` upper bound is conservative — well above the largest single-protocol TVL on DefiLlama at the time of writing — but we can dial it higher if a legitimate TVL approaches it.
- The fetch-adapter heuristic looks for `axios` / `node-fetch` + a non-RPC HTTP URL. It's intentionally a heuristic — false positives are possible (e.g. an adapter that imports axios for an RPC node behind a non-standard hostname). The phrasing of the warning makes that clear and asks the contributor to ignore if not applicable.
- The script reads the adapter source file to detect the `methodology` export and the fetch-adapter pattern. For directories, it looks at `index.js`. For single-file adapters, it reads the file directly.
- I could add a TVL spike check (current vs `-1d` vs `-7d`) as a follow-up — it'd add ~60s per adapter via additional `test.js` runs, which is meaningful but bounded. Happy to do that as a separate PR if you'd like.

## Summary by CodeRabbit

* **Chores**
  * Enhanced CI pipeline with additional automated validation checks for adapter quality, including TVL bounds validation and missing configuration field detection.

### DIFF
--- a/.github/workflows/sanityCheck.js
+++ b/.github/workflows/sanityCheck.js
@@ -1,0 +1,160 @@
+// Adds sanity warnings to a PR alongside the existing TVL output produced by
+// test.js / commentResult.js. Catches the common shipping-failure modes that
+// the existing pipeline does not flag, e.g. an adapter that runs cleanly but
+// computes $0 (unpriced tokens — almost always the cause when a new chain is
+// involved) or $50B+ (likely double-counting / wrong decimals).
+//
+// Usage:
+//   node .github/workflows/sanityCheck.js <output.txt> <pr-comments-dir> <adapter-path>
+//
+// Designed to be invoked from test.yml right after commentResult.js. Writes its
+// own markdown file into the pr-comments dir; comment.yml will pick it up and
+// post it. Never throws — on any error it just exits 0 silently so the rest of
+// the CI pipeline is unaffected.
+
+const path = require('path');
+const fs = require('fs');
+
+// $50B is well above the largest single-protocol TVL on DefiLlama at the time of
+// writing. Anything above that on a fresh adapter is almost always a bug
+// (decimals, double-counting, wrong owner).
+const HIGH_TVL_USD = 50_000_000_000;
+
+function parseTotalTvl(testOutput) {
+  // commentResult.js parses the same "------ TVL ------" section. We look for
+  // the "total" line: e.g. "total                    659.34 k"
+  const tvlSection = testOutput.split('------ TVL ------')[1];
+  if (!tvlSection) return null;
+
+  const totalMatch = tvlSection.match(/^\s*total\s+([\d.,]+)\s*([kKmMbB])?\s*$/m);
+  if (!totalMatch) return null;
+
+  let value = Number(totalMatch[1].replace(/,/g, ''));
+  if (Number.isNaN(value)) return null;
+  const suffix = (totalMatch[2] || '').toLowerCase();
+  if (suffix === 'k') value *= 1e3;
+  else if (suffix === 'm') value *= 1e6;
+  else if (suffix === 'b') value *= 1e9;
+  return value;
+}
+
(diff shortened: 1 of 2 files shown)


g1nt0ki:
sorry, not needed atm

g1nt0ki:
@DeFiDonkey to give bit more info, I dont want to complicate the script and the people who put up PR needs to test locally and put up the PR only when the adapter is ready

PR #18964, opened 2026-04-28
helper: add retry/timeout and fix correctness bugs

bajajanirudh:
Hardens the two helpers that mediate every off-chain HTTP call (`helper/http.js`, `helper/cache.js`) and fixes a silent LP under-counting bug in `helper/unwrapLPs.js`.

### Reliability
- Added `withRetry` (exp backoff + jitter, honors `Retry-After`, retries 429/5xx/network errors). Defaults: 3 retries, 60s timeout. Opt out via `{ retries: 0, timeout: 0 }`.
- Applied to `get`, `post`, `getWithMetadata`, `blockQuery`, and the GraphQL paths in `cachedGraphQuery` / `graphFetchById`.
- `cache.js` now routes through the resilient `httpGet` / `httpPost` instead of raw axios.

### Correctness fixes
- **`http.getCache` poisoned promise**: a transient rejection no longer poisons the in-memory cache for the rest of the process — the entry is evicted on failure.
- **`cache.js` key collision**: in-memory `configCache` is now keyed by `project + hash(endpoint, vars/body)`. Previously two `getConfig` sites sharing a project name silently shared one promise. Disk key is unchanged so existing fallback files still work.
- **`_setCache` validity check**: replaces the `length > 42` heuristic — small valid configs are now persisted, garbage strings >42 chars are no longer accepted.
- **`unwrapLPsAuto` LP detection**: `erc20:symbol` returning `null` on a transient RPC blip used to silently mis-classify the LP as a plain ERC20. One-shot retry resolves it before classification.

### Risk
Failure path is bounded but slower (up to ~3.7s extra per failed call before fallback). Failures stay strictly per-call-site — no new cross-adapter coupling. Adapters needing >60s for a single call should pass `{ timeout: 0 }`.

## Summary by CodeRabbit

* **New Features**
  * Automatic retries with exponential backoff, jitter, Retry-After respect, and per-request timeout controls.
  * Per-request retry/timeouts and improved request de-duplication to reduce redundant calls and speed responses.

* **Bug Fixes**
  * Stricter cache validation to avoid storing empty or error payloads while preserving small valid responses.
  * In-flight failures no longer leave stale cache entries; concurrent requests are de-duplicated and cleaned up on error.

### DIFF
--- a/projects/helper/cache.js
+++ b/projects/helper/cache.js
@@ -1,7 +1,21 @@
 const sdk = require('@defillama/sdk')
+const crypto = require('crypto')
 const Bucket = "tvl-adapter-cache";
-const axios = require('axios')
 const graphql = require('./utils/graphql')
+const { get: httpGet, post: httpPost, withRetry } = require('./http')
+
+// Hash of the request shape (endpoint + body/variables). Used to disambiguate the in-memory
+// promise cache when multiple call sites share a `project` name but hit different endpoints.
+// Disk cache keys remain project-only so existing fallback semantics are preserved.
+function _requestHash(...parts) {
+  const h = crypto.createHash('md5')
+  for (const p of parts) {
+    if (p == null) continue
+    h.update(typeof p === 'string' ? p : JSON.stringify(p))
+    h.update('\0')
+  }
+  return h.digest('hex').slice(0, 12)
+}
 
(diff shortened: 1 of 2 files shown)


RohanNero:
Sorry, closing as this is not needed at the moment

PR #19012, opened 2026-04-30
fix(ring-dex): unwrap Few pair reserves

### LINKED ISSUE #10741: Wrong TVL Data for Ring Swap
Looks like the Defillama website set a wrong TVL data source for Ring Swap (https://defillama.com/protocol/ring-swap). Right now, Defillama website shows $74m, while we have more than $88m in Ring Swap (https://info.ring.exchange/home)
Please help to take a look.

ardaerturk:
Closes #10741

## Context

Ring Swap pairs can hold Few-wrapped assets. The adapter was excluding those reserves, which left most pool liquidity out of TVL.

## Changes

- Convert Few-wrapped pair reserves into their proportional underlying asset balances using on-chain wrapper balances and total supply.
- Recursively unwrap nested Few wrappers before adding balances.
- Keep the existing factory and pair discovery flow.

## Tests

```bash
node -c projects/ring-dex/index.js
node test.js projects/ring-dex/index.js
```

### DIFF
--- a/projects/ring-dex/index.js
+++ b/projects/ring-dex/index.js
@@ -45,7 +45,8 @@ Object.keys(fewFactoryConfig).forEach(chain => {
  * Converts Few-wrapped balances in-place until final underlying assets remain.
  */
 async function unwrapFewBalances(api, balances, fewTokenSet) {
-  while (true) {
+  const maxDepth = Math.max(fewTokenSet.size, 1)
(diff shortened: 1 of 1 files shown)


g1nt0ki (changes requested):
@ardaerturk thanks for the PR but the issue is was wrong. We are intentionally excluding few tokens from the dex tvl as they are by the team and we are already counting that tvl under ring-few

PR #19060, opened 2026-05-01
Add LayerZero TVL helper and LayerZero RISE bridge adapter

slvDev:
# Add LayerZero TVL helper and LayerZero RISE bridge adapter

Adds a LayerZero TVL helper at `projects/helper/layerzero.js` and a new adapter at `projects/layerzero-rise/index.js` for the LayerZero side of the RISE bridge.

The RISE team originally proposed parallel bridge tracking in PR #18838: an OP Stack `OptimismPortal2` adapter at `projects/rise/index.js` plus a separate LayerZero adapter at `projects/layerzero-rise/index.js`, with stated methodology "Canonical bridge TVL + LayerZero-enabled bridge TVL". PR #18850 landed the OP Stack portal but dropped the LayerZero half because all LayerZero TVL was paused after the April 2026 KelpDAO incident. Issue #18926 has been open since then asking for a safer way to count it. This PR adds the LayerZero half back, behind a small safety helper that other LayerZero adapters can reuse.

The existing `projects/rise/index.js` (OP Stack `OptimismPortal2` at `0xad92...c4C`, around $108k of L1 ETH escrow) is unchanged. The two adapters are independent and do not overlap: the portal escrows ETH for canonical L1-to-L2 deposits, the LayerZero contracts escrow USDC for OFT mints on RISE.

The helper counts source-side escrow balances only. It never reads destination OFT `totalSupply`.

For each escrow it reads the LayerZero V2 send-side ULN config and computes verifier quorum as `requiredDVNCount + optionalDVNThreshold`. Paths below `minDvnQuorum` are excluded. The helper throws if `minDvnQuorum` is missing, so a future adapter cannot silently opt out. The RISE adapter uses `minDvnQuorum: 2`.

`assertDvnArrayValid` rejects malformed configs: a DVN array shorter than its declared count, an `address(0)` entry (which would let `[realDVN, address(0)]` pass a naive count check), or duplicate addresses (which would overstate independent signers).

Scope is LayerZero V2 only (EndpointV2 at `0x1a44...728c`). V1 endpoints are not supported.

RISE escrows, all USDC, dstEid 30401:

- Ethereum `0xf43a70B250a86003a952af7A7986CcC243B89D81`
- Base `0x82675d0553D802039e6776C006BEb1b820a69d55`
- Arbitrum `0x82675d0553D802039e6776C006BEb1b820a69d55`

[...]

PR #19065, opened 2026-05-01
add LayerZero safety gate to USDT0

slvDev:
# Add LayerZero safety gate to USDT0

This PR depends on #19060.

Wires the existing single-address USDT0 adapter through the LayerZero TVL helper introduced in #19060. The current `projects/usdt0/index.js` reads the Ethereum source escrow balance directly, without checking the LayerZero DVN configuration on the routes secured by that escrow.

Because this PR is stacked on the #19060 branch, the diff against `main` includes the helper file from that PR. The meaningful new code is the `dstEids` extension to the helper and the USDT0 adapter rewrite. Once #19060 merges, this branch will rebase onto main and only the USDT0-relevant diff will remain.

This PR makes two changes:

1. `projects/helper/layerzero.js`: a backward-compatible extension to `sumLayerZeroEscrows`. Escrow entries can now declare `dstEids: [...]` instead of a single `dstEid`; the helper includes an escrow only when every route meets the quorum. The input is type-checked (must be an array) and deduplicated. RISE's existing single-route consumer continues to work unchanged.

2. `projects/usdt0/index.js`: switched from a raw `balanceOf` to `sumLayerZeroEscrows`, gated by `minDvnQuorum: 3` over the 21 destination EIDs configured on Ethereum's OApp.

The quorum is 3, not 2, because Ethereum's USDT0 OApp currently runs `requiredDVNCount=3, optionalDVNCount=0, optionalDVNThreshold=0` on every route (Tether Operations DVN, LayerZero Labs DVN, Nethermind DVN). A drop to 2-of-N would be a real safety reduction, so the threshold matches the current posture rather than allowing silent degradation.

The adapter is marked `timetravel: false` because the gate depends on the live LayerZero route configuration. Historical route configs were weaker before the current 3-DVN posture, so replaying this gate at old blocks would zero out historical escrow balances rather than reproduce the old raw-balance adapter.

I verified the scope on-chain:

- Ethereum is the only non-zero USDT source escrow counted by this adapter. Most destination OApps are mint-side OFTs; Hedera (EID 30316) and Tempo (EID 30410) source-escrow OApps were observed but currently hold zero USDT0.
- All 21 send-side routes from Ethereum return `req=3 opt=0 thr=0` quorum 3, via SendUln302 `0xbB2Ea70C9E858123480642Cf96acbcCE1372dCe1`.
- Hedera and Tempo can be added in a follow-up if their source escrow balances become non-zero.

Test:

```
$ node test.js projects/usdt0/index.js

--- ethereum ---
USDT                      3.68 B
Total: 3.68 B

------ TVL ------
ethereum                  3.68 B

total                    3.68 B
```

The number is identical to the pre-change adapter. This PR adds the gate, not new TVL.

Lint is clean. This PR adds no npm dependencies.

## Summary by CodeRabbit

* **New Features**
  * Added LayerZero RISE TVL tracking on Ethereum, Base, and Arbitrum
  * Introduced a helper to aggregate LayerZero escrows with verifier-quorum validation and escrow route filtering

* **Updates**
  * Switched USDT0 Ethereum TVL to LayerZero-based aggregation with a 3-DVN quorum requirement
  * Disabled timetravel and added methodology text for affected adapters

### DIFF
--- a/projects/helper/layerzero.js
+++ b/projects/helper/layerzero.js
@@ -28,11 +28,7 @@ async function getSendUlnConfig(api, { oapp, dstEid, endpoint = ENDPOINT_V2 }) {
 
   assertDvnArrayValid(requiredDVNs, requiredDVNCount, 'required');
   assertDvnArrayValid(optionalDVNs, optionalDVNCount, 'optional');
-
-  const combined = [...requiredDVNs, ...optionalDVNs].map(a => a.toLowerCase());
-  if (new Set(combined).size !== combined.length) {
-    throw new Error('LayerZero DVN config reuses a verifier across required/optional sets; quorum would overstate independent signers');
-  }
+  assertNoDuplicateDvns([...requiredDVNs, ...optionalDVNs]);
 
   return {
     confirmations: Number(config.confirmations),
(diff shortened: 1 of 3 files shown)


g1nt0ki:
sorry, no again, when we compute bridge tvl, we care only about the amount of tokens locked in the source chain. And we might remove this listing entirely and tracking everything under lz

PR #19067, opened 2026-05-01
add easya-kickstart

fastparrotlol:
Adds a TVL adapter for EasyA Kickstart, a permissionless ideas launchpad on Solana built on top of Meteora Dynamic Bonding Curve (DBC).

How TVL is computed:
- Source: on-chain only, via `getProgramAccounts` against the Meteora DBC program (`dbcij3LWUppWqq96dh6gJWwBifmcGfLSB5D4DuSMaqN`).
- Filter: `dataSize=424` plus `memcmp` at offset 72 against each of EasyA's three PoolConfig accounts.
- For every active VirtualPool (anchor discriminator verified, `is_migrated == 0`), the adapter sums the `quote_reserve` field at offset 240 as WSOL.
- Migrated (graduated) pools are excluded since their liquidity has already moved to Meteora DAMM V2 and is tracked under Meteora.

Identification of EasyA pools:
- Three PoolConfig accounts: `FctVFHQvVaj3hTDHCSXZjTmmsRs5bX5ogUPGHFSgrJpU`, `NHT6MNushFNWpaFgQs5k49HHzsas9jQAVoRvqyXc5Qx`, `6iEekXhre85eDB1mxRuXbRDHbSG8HeSPYopp9e7fp4BJ`.
- All three share the same `fee_claimer` at PoolConfig offset 168 (`EfgbywXHbDnkbr4hqSv9dyo9hUJ62jmi2fm4e6gP2jLb`), confirming common ownership by EasyA.
- All three use WSOL as the quote asset.

Validation run:
- `node test.js projects/easya-kickstart/index.js` returned `solana 17.14k` TVL.
- Cross-checked against EasyA's own API liquidity sum ($16.9k), within 1.4 percent.

---
##### Name (to be shown on DefiLlama):
EasyA Kickstart

##### Twitter Link:
https://x.com/EasyA_Kickstart

##### List of audit links if any:
none

##### Website Link:
https://kickstart.easya.io/

##### Logo (High resolution, will be shown with rounded borders):
SVG to be attached via edit after PR creation.

##### Current TVL:
~$17,140 from local adapter test.

##### Treasury Addresses (if the protocol has treasury)
No separate treasury. Launchpad fee_claimer wallet: `EfgbywXHbDnkbr4hqSv9dyo9hUJ62jmi2fm4e6gP2jLb`.

##### Chain:
Solana

##### Coingecko ID:
none (no token)

##### Coinmarketcap ID:
none (no token)

[...]

PR #19086, opened 2026-05-02
add rise-protocol

fastparrotlol:
## Summary
Adds TVL + Borrowed adapter for [Rise](https://rise.rich) — a Solana launchpad with on-chain lending, built on top of the Mayflower program via CPI.

- **TVL**: sum of `MarketState.totalCashLiquidity` across all Rise markets in the underlying Mayflower `MarketLinear` accounts (in the market's main quote token: WSOL / USDC / etc.)
- **Borrowed**: sum of `MarketState.totalDebt` across the same markets

## Implementation notes

To stay friendly with the public Solana RPC's per-method 429 cooldown:

- One `getProgramAccounts` against Mayflower, filtered by Anchor discriminator + Rise's `MarketGroup` pubkey (`HA9pvTe8...`). `dataSlice` trims the response to the 32-byte `mintMain` field only — the corresponding `MarketLinear` PDA is derived locally.
- `MarketLinear` accounts are then read via `getMultipleAccountsInfo` in batches of 100 (different RPC rate-limit bucket from `getProgramAccounts`).
- Both `tvl` and `borrowed` share a single in-flight fetch (in-flight promise cache, 60 s TTL) so the adapter does not double-call.
- On a transient refresh failure (or an unexpectedly empty market scan), the last good snapshot is returned instead of zeroing out TVL/Borrowed.

## Local test
On the public `api.mainnet-beta.solana.com` RPC, current readings:
- TVL: ~$2.71M (SOL 2.69M + USDC 17.41k)
- Borrowed: ~$32.23M (SOL 32.02M + USDC 212.04k)

## Companion PRs
- DefiLlama/dimension-adapters#6765 — Volume + Fees + Revenue adapter
- DefiLlama/defillama-server#11912 — listing entry (id `7778`)
- DefiLlama/icons#2372 — protocol icon (`rise.jpg`)

### DIFF
--- a/projects/rise-protocol/index.js
+++ b/projects/rise-protocol/index.js
@@ -1,7 +1,7 @@
 const { getConnection } = require('../helper/solana')
 const { sliceIntoChunks } = require('../helper/utils')
 const { PublicKey } = require('@solana/web3.js')
-const { bs58 } = require('@project-serum/anchor/dist/cjs/utils/bytes')
+const bs58 = require('bs58').default
 
 // Rise (https://rise.rich) is a Solana launchpad with on-chain lending built in.
 // User-launched tokens live on Rise's program; the actual liquidity, debt and
@@ -68,6 +68,10 @@ async function getRiseMarkets() {
   _inflight = (async () => {
     try {
       const result = await fetchRiseMarkets()
+      // Don't overwrite a known-good snapshot if a refresh comes back empty
+      // (e.g., the RPC silently served an empty getProgramAccounts result):
+      // it would zero out TVL and Borrowed for no real reason.
+      if (!result.length && _cached?.length) return _cached
       _cached = result
       _cachedAt = Date.now()
       return result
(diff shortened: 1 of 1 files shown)


fastparrotlol:
Following up on the close of #19067, I want to flag that the analogous concern there (double-counting against Meteora) does not apply here.

EasyA is built on top of **third-party** Meteora infrastructure, which is independently listed on DefiLlama, hence the double-count.

Rise's case is structurally different: Mayflower is **Rise's own backend program**, not a third-party dependency. It was split out into a separate Anchor program for modularity, but it is operated by the same team (verified on-chain: the admin of the Rise MarketGroup `HA9pvTe8...` matches the Rise program admin `5scY2JG...`). Mayflower is not a standalone product, has no separate frontend, no separate brand, and is not listed on DefiLlama.

So this PR is not surfacing TVL that is already counted somewhere else, it is the only place this TVL would ever be counted. The marketGroup memcmp filter ensures we sum only Rise-owned markets, never any other tenant that might use Mayflower in the future.

Happy to add a short note to the methodology making this relationship explicit if useful.

RohanNero:
Thanks for the PR, I've added it in #19252

PR #19089, opened 2026-05-03
fix(contango-v2) update subgraph endpoint and pagination

### LINKED ISSUE #19022: fix contango v2 tvl
http://defillama.com/protocol/contango-v2

vsamavila:
Fixes #19022

The Contango V2 adapter was failing because it still used the old Satsuma subgraph URLs, like:
`https://subgraph.satsuma-prod.com/773bd6dfe1c6/egills-team/v2-arbitrum/api`

That host no longer resolves, so the adapter could not export TVL.

I found two things that needed changing:
- The subgraph URLs needed to move from the old Satsuma host to the public Goldsky endpoints.
- The old query asked the subgraph for 10,000 rows at once. Goldsky rejects that with `The first argument must be between 0 and 1000, but is 10000`, so I changed the positions/assets queries to page through results instead.

This keeps the calculation the same. The subgraph is still only used to discover positions/assets, and TVL/borrowed are still calculated from the Contango lens/vault calls.

Tested with:

```
node test.js projects/contango-v2/index.js
```

Local run:

```
total                    9.45 M
borrowed                 64.21 M
```

## Release Notes

* **Refactor**
  * Updated data fetching infrastructure to use shared request helper and Goldsky endpoints
  * Improved pagination approach for position and asset queries with reusable helpers
  * Enhanced data retrieval efficiency with updated pagination parameters

### DIFF
--- a/projects/contango-v2/index.js
+++ b/projects/contango-v2/index.js
@@ -1,8 +1,9 @@
-const axios = require('axios')
+const { request } = require('../helper/utils/graphql')
 
 const CONTANGO_PROXY = "0x6Cae28b3D09D8f8Fc74ccD496AC986FC84C0C24E";
 const CONTANGO_LENS_PROXY = "0xe03835Dfae2644F37049c1feF13E8ceD6b1Bb72a";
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR

PR #19092, opened 2026-05-03
fix(namada) use official infra for shielded pool tvl

### LINKED ISSUE #19023: Fix namada shielded pool tvl
https://defillama.com/protocol/namada-shielded-pools

vsamavila:
Fixes #19023

The Namada adapter was failing because it still used the old Tududes indexer/RPC.

The adapter now uses `https://indexer.namada.net` for token discovery and `https://rpc.namada.net` for the balance reads. Token discovery comes from the official Namada Indexer account endpoint (`/api/v1/account/<MASP_ADDRESS>`), while the actual MASP balances still come from Namada RPC through the existing `abci_query` storage path.

  Native NAM remains excluded, matching the existing non-native MASP balance scope.

Reference:
https://docs.namada.net/networks/services

Tested with:
```
node test.js projects/namada/index.js

local run:
namada                    526.18 k
total                     526.18 k
```

## Release Notes

* **Refactor**
  * Optimized Namada's Total Value Locked (TVL) computation by integrating with the official indexer for improved token discovery. The new implementation provides more reliable and efficient balance reporting through standardized RPC queries and simplified balance decoding, while maintaining accuracy and reducing code complexity.

### DIFF
--- a/projects/namada/index.js
+++ b/projects/namada/index.js
@@ -1,121 +1,55 @@
-const { fetchURL } = require('../helper/utils');
-const BigNumber = require("bignumber.js");
+const { get, post } = require('../helper/http');
 
-// Define the Borsh schema for deserializing the balance response
-// This should match the Amount struct in Namada which has 4 u64 limbs
-class Amount {
-    constructor({ amount }) {
-        this.amount = amount;
-    }
-}
-
-const schema = new Map([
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR

PR #19105, opened 2026-05-04
add stacks api depends

noateden:
## Summary by CodeRabbit

* **Chores**
  * Updated project dependencies to latest compatible versions to support enhanced capabilities.


g1nt0ki (changes requested):
no, we should instead edit hermetica so it doesnt have this dependency

PR #19106, opened 2026-05-04
scripts/audit: structural sanity checks for coreAssets and adapter sources

Himess:
## Summary

Adds five small Node scripts under `scripts/audit/` that complement DefiLlama's existing TVL spike detector (`defillama-server/defi/src/api2/scripts/tvlSpikeDetector.ts`).

The spike detector watches for **changes** in TVL — anomalies relative to a protocol's recent history. These scripts watch for **shapes** — always-on data bugs that don't manifest as a spike because they were already present at day-zero. Two real bugs in this repo surfaced while developing them; fix PRs are cross-linked at the bottom.

## Scripts

| Script | What it checks | Network calls |
|--------|----------------|---------------|
| `core-assets.js` | Static checks against `projects/helper/coreAssets.json`: EIP-55 checksum sanity, placeholder addresses, suspicious symbol names, cross-chain symbol/address mismatch | None |
| `onchain-symbols.js` | RPC `symbol()` for every EVM entry vs its key — finds renamed tokens, dead contracts, wrong-address bugs | RPC multicall per chain |
| `dex-token-universe.js` | Greps adapter source for risky token-universe sources (explorer APIs, unfiltered factory enumeration, balance-of-anything loops) | None |
| `uni-tvl-no-whitelist.js` | `getUniTVL({...})` calls that don't enable `useDefaultCoreAssets` and don't pass a custom `coreAssets` array — the DogePay class | None |

## Sample output

```sh
$ node scripts/audit/core-assets.js --severity=high
Scanned 357 top-level keys, 1464 unique EVM addresses
Findings: 1 (filtered to severity >= high)

── bad-checksum (1) ──
  [high] saga :: tBTC
     EIP-55 checksum is invalid (mixed case but does not validate)
     address: 0x7cF468a019C5bf734311D10C3a429bB504CAF3Ce
     expected: 0x7cF468a019C5bf734311D10C3a429bB504CAF3ce
```

```sh
$ node scripts/audit/uni-tvl-no-whitelist.js | head -5
Adapters with getUniTVL but no whitelist: 20

✗ archly-finance  →  archly-finance/index.js:40
   getUniTVL({ factory: ARCHLY_FACTORY_OTHER, seDefaultCoreAssets: true, hasStablePools: true, })
```

[...]

PR #19107, opened 2026-05-04
fix(coreAssets): correct saga.tBTC EIP-55 checksum

Himess:
## Fix

Single character — flips the penultimate `C` of `saga.tBTC`'s recorded address from upper- to lower-case so it validates as a proper EIP-55 checksum:

```diff
- "tBTC": "0x7cF468a019C5bf734311D10C3a429bB504CAF3Ce",
+ "tBTC": "0x7cF468a019C5bf734311D10C3a429bB504CAF3ce",
```

(The final `e` was already lowercase in both versions; only the `C` immediately before it changes.)

## Why it matters

The mis-cased version isn't a valid EIP-55 checksum. `ethers` v6 rejects it consistently across recent versions:

```js
> ethers.getAddress("0x7cF468a019C5bf734311D10C3a429bB504CAF3Ce")
TypeError: bad address checksum (argument="address", value="0x7cF468…CAF3Ce", code=INVALID_ARGUMENT, version=6.x)
```

`encodeFunctionData` runs the same validation internally, which means any `multiCall` path that takes the address as a `target` or as a `params: [address]` argument fails with the opaque `Bad encoding` error reported by [DefiLlama/defillama-sdk#217](https://github.com/DefiLlama/defillama-sdk/issues/217). Single-call `eth_call` happens to bypass the check (the address is forwarded to the JSON-RPC `to` field as a raw string), which is what hides the bug from quick smoke-tests.

## How it was found

[`scripts/audit/core-assets.js`](https://github.com/DefiLlama/DefiLlama-Adapters/pull/19106) — a small structural sanity pass over `coreAssets.json`. The bug doesn't show up to the spike detector (the address never produced a TVL number to begin with) but it shows up immediately in a static checksum scan.

```
$ node scripts/audit/core-assets.js --severity=high
── bad-checksum (1) ──
  [high] saga :: tBTC
     EIP-55 checksum is invalid (mixed case but does not validate)
     address: 0x7cF468a019C5bf734311D10C3a429bB504CAF3Ce
     expected: 0x7cF468a019C5bf734311D10C3a429bB504CAF3ce
```

Audit framework: #19106. Companion fix: `archly-finance` whitelist typo — #19108.


RohanNero:
Closing as duplicate of #19106, also saga is marked dead and tBTC is never referenced (fixed in #19286)

PR #19130, opened 2026-05-05
Fix Aquarius Stellar TVL adapter

wild-feather:
## Summary
Fix Aquarius Stellar TVL adapter to require an exact daily datapoint for the requested timestamp.

## Changes
- Keep UTC date formatting for daily key lookup
- Remove nearest-date fallback logic
- Throw an error when the current timestamp is missing from the source data
- Keep response-shape validation for Aquarius API payloads
- Keep TVL numeric validation before adding balances

## Validation
- Confirmed the adapter now uses exact-date matching only
- Confirmed missing daily data will throw instead of silently using an older/newer row
- Confirmed existing API response validation and finite TVL checks remain in place

Please enable "Allow edits by maintainers".

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TVL fetching reliability by caching in-flight requests and resetting on HTTP failure
  * Stricter response validation, clearer errors for missing or non-finite TVL values
  * Fixed TVL parsing and scaling so reported values use the correct numeric conversion

* **New Features**
  * Accepts multiple server response shapes (top-level array or common wrappers) when reading TVL records

* **Chores**
  * Updated date formatting to match API UTC ISO format

### DIFF
--- a/projects/aqua-network/index.js
+++ b/projects/aqua-network/index.js
@@ -35,10 +35,15 @@ function pickClosestRow(rows, targetDate) {
   const exactMatch = sortedRows.find((row) => row.date === targetDate)
   if (exactMatch) return exactMatch
 
-  const priorRows = sortedRows.filter((row) => row.date < targetDate)
-  if (priorRows.length) return priorRows[priorRows.length - 1]
+  return sortedRows.reduce((closestRow, currentRow) => {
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR, I have resolved in #19328

PR #19174, opened 2026-05-07
fix: disable timetravel for VisionBoard Vault

bergertolga:
VisionBoard Vault's ERC-4626 totalAssets() includes active HyperCore/HyperEVM vault accounting and needs to be read at latest state. The original custom adapter had timetravel disabled, but the registry refactor dropped that flag.\n\nCurrent behavior:\n- Latest local adapter run exports ~6.63k USDC TVL.\n- Historical/timestamped run at the API sample timestamp (2026-05-07 20:26:47 UTC / 1778185607) exports ~24.99 USDC, matching the stale DefiLlama API value.\n- On-chain latest totalAssets() currently returns ~6.63k USDC.\n\nThis adds timetravel: false back to the registry entry so DefiLlama uses the live/latest totalAssets() value rather than stale historical calls.\n\nTest:\n```bash\nrm -rf tvl-adapter-repo .cache cache\nnode test.js visionboard-vault\n# hyperliquid 6.63 k / total 6.63 k\n```\n

## Summary by CodeRabbit

* **Chores**
  * Updated TVL computation configuration for visionboard-vault protocol.

### DIFF
--- a/registries/erc4626.js
+++ b/registries/erc4626.js
@@ -269,6 +269,7 @@ const configs = {
     methodology: "TVL represents the total value of assets held within the vault. Each vault token is minted using USDC and appreciates in line with the performance of the underlying asset.",
   },
   'visionboard-vault': {
+    timetravel: false,
     hyperliquid: ['0x0a5e236425aca07fd087904F8863CAd554675E06'],
     methodology: 'TVL is calculated from VisionBoard Vault totalAssets() on HyperEVM. Deposits mint VBV vault shares backed by the vault asset, currently USDC.'
   },


bergertolga:
Clarifying the issue this fixes:

The current DefiLlama value (~$24) lines up with the vault's idle USDC sitting directly on HyperEVM, but that is not the full vault TVL.

VisionBoard Vault moves/converts deposited HyperEVM USDC into HyperCore account collateral for perp trading. Most vault value therefore lives in HyperCore state, not as ERC-20 cash sitting in the HyperEVM vault contract. That includes withdrawable/cash balance and deployed perp position assets/equity/PnL.

The vault's ERC-4626 `totalAssets()` is the source of truth because it reflects this full vault accounting/account equity. Reading only `USDC.balanceOf(vault)` on HyperEVM misses both the HyperCore cash/withdrawable balance and the active position equity.

Current live checks:
- `USDC.balanceOf(vault)` on HyperEVM = ~25 USDC idle in the vault contract
- `totalAssets()` = ~6.4k USDC
- The VisionBoard Vault UI reports the same ~6.4k TVL/accounting value, while HyperCore capital is used for perp trading

So if the adapter/indexer effectively tracks only the token balance held by the vault contract on HyperEVM, it undercounts TVL as ~$25. For this vault, DefiLlama should use latest `totalAssets()` as the source of truth. Adding `timetravel: false` restores the behavior from the original custom adapter and avoids misleading historical/timestamped reads for this live HyperCore-accounting vault.

bergertolga:
@RohanNero quick follow-up to the VisionBoard Vault registry refactor. The original adapter had `timetravel: false`, but the refactor dropped it, so DefiLlama is showing only idle HyperEVM USDC (~$25) instead of ERC-4626 `totalAssets()` (~$6.4k), which includes HyperCore account equity.

Tests are green and this is a 1-line fix. Could you merge when you get a chance?

RohanNero (changes requested):
Hello, could you please verify the vault's proxy and implementation contracts? I don't think adding `timetravel: false` will solve this issue, [altura](https://defillama.com/protocol/altura) is another hyperliquid vault inside the registry that works without it

bergertolga:
Thanks @RohanNero, that makes sense. I checked Altura as a comparison too.

VisionBoard is a bit different from Altura. Altura looks like the vault bytecode is directly at the registry address, while VisionBoard uses an EIP-1967/UUPS proxy.

VisionBoard Vault proxy:
`0x0a5e236425aca07fd087904F8863CAd554675E06`

Current implementation from the EIP-1967 implementation slot:
`0x2EbDef899C4BD3b1bF3CB1cFf0e9359fBC7E6989`

Latest proxy calls:
- `asset()` = `0xb88339CB7199b77E23DB6E890353E22632Ba630f` / HyperEVM USDC
- `totalAssets()` = `7237863108` = `7,237.863108` USDC

For comparison, Altura:
- vault: `0xd0Ee0CF300DFB598270cd7F4D0c6E0D8F6e13f29`
- EIP-1967 implementation slot is zero
- bytecode is directly at the vault address

So I think Altura may not be directly comparable to VisionBoard’s proxy setup.

The issue we are trying to fix is that DefiLlama seems to be showing only the idle HyperEVM USDC in the vault contract. VisionBoard’s `totalAssets()` includes the full vault accounting, including HyperCore-backed equity/collateral.

If `timetravel: false` is not the right fix, happy to adjust the PR to the adapter shape you prefer for reading latest `totalAssets()` through the proxy.

RohanNero:
@bergertolga Despite VisionBoard using a proxy, tvl is being counted as `totalAssets()` for both vaults so that shouldn't be causing an issue. Could you verify the implementation contract or share a github link to the source code? Just want to confirm that `totalAssets()` is behaving as expected.

bergertolga:
Thanks — verified on-chain.

VisionBoard Vault is a UUPS/ERC1967 proxy on HyperEVM:

- Proxy / vault: `0x0a5e236425aca07fd087904F8863CAd554675E06`
- Current implementation from the ERC1967 implementation slot: `0x2ebdef899c4bd3b1bf3cb1cff0e9359fbc7e6989`
- Chain: HyperEVM `999`
- Share decimals: `6`

`totalAssets()` is implemented as a live aggregate of the vault's underlying balances, not as share supply:

```solidity
/// @notice Total vault value: EVM USDC + HyperCore spot + perp equity
function totalAssets() external view returns (uint256) {
    return _totalAssets();
}

function _totalAssets() internal view returns (uint256) {
    uint256 evmBal = IERC20(asset).balanceOf(address(this));

    uint256 spotBal = 0;
    try this._spotBalance() returns (uint256 s) {
        spotBal = s;
    } catch {}

    uint256 perpVal = 0;
    try this._perpEquity() returns (uint256 p) {
        perpVal = p;
    } catch {}

    return evmBal + spotBal + perpVal;
}
```

I also checked the current live calls via HyperEVM RPC:

- `totalAssets()` → `8515025756` raw = `8,515.025756` USDC
- `lastKnownAssets()` → `5102188229` raw = `5,102.188229` USDC
- `totalSupply()` → `6483180378585` raw, 6 decimals

The source repo is currently private, so I can't provide a public GitHub source link in the PR thread yet. If useful, I can publish/verify the implementation source or make the relevant contract source public separately; the key bit for the adapter is that `totalAssets()` intentionally returns `EVM USDC + HyperCore spot USDC + xyz perp equity`.

RohanNero:
@bergertolga Thanks for the update, if you don't mind, I think it would be useful to see the currently deployed implementation verified: https://hyperevmscan.io/address/0x2ebdef899c4bd3b1bf3cb1cff0e9359fbc7e6989

If spotBal and perpVal are ever 0, or if `_perpEquity()` or  `_spotBalance()` ever failed, then it would explain why totalAssets() is equal to idle assets.

bergertolga:
Good catch. I checked the deployed proxy directly through HyperEVM RPC.

Current implementation is still:

- proxy/vault: `0x0a5e236425aca07fd087904F8863CAd554675E06`
- ERC1967 implementation: `0x2ebdef899c4bd3b1bf3cb1cff0e9359fbc7e6989`
- HyperEVMScan implementation page: https://hyperevmscan.io/address/0x2ebdef899c4bd3b1bf3cb1cff0e9359fbc7e6989

I re-ran the component read functions on the proxy. At the current block they are **not** failing and are **not** zero:

```text
vault.totalAssets()  = 8416919661 raw = 8,416.919661 USDC
vault._spotBalance() = 63269 raw      = 0.063269 USDC
vault._perpEquity()  = 8391856392 raw = 8,391.856392 USDC
```

So the deployed implementation is currently counting mostly `perpVal`, not just idle EVM USDC.

That said, your concern is valid: the deployed implementation intentionally wraps the HyperCore precompile reads in `try/catch`:

```solidity
uint256 spotBal = 0;
try this._spotBalance() returns (uint256 s) {
    spotBal = s;
} catch {}

uint256 perpVal = 0;
try this._perpEquity() returns (uint256 p) {
    perpVal = p;
} catch {}
```

So if either HyperEVM precompile call reverts/fails for a historical block, RPC/backend issue, or unsupported timetravel context, that component silently falls back to `0`. If both fail, `totalAssets()` can indeed collapse to only the vault's idle EVM USDC balance. That is likely the reason this protocol should not be timetraveled.

On source verification: I’m working on publishing the exact implementation source to HyperEVMScan, but I don’t want to mark it verified until the compiler output matches the deployed bytecode exactly. The local source I have is close but not an exact bytecode match yet, so I’m treating the live RPC behavior above as the source of truth for this adapter issue.

bergertolga:
@RohanNero quick follow-up — I re-checked live proxy reads today and `totalAssets()` is still behaving as expected:

```text
block                35393646
proxy/vault          0x0a5e236425aca07fd087904F8863CAd554675E06
implementation       0x2EBdeF899C4bd3b1BF3cB1cfF0e9359Fbc7E6989
vault.totalAssets()  10,984.258195 USDC
vault._spotBalance() 0.057025 USDC
vault._perpEquity()  10,962.660580 USDC
```

So the current implementation is clearly counting mostly HyperCore/perp equity, not just idle HyperEVM USDC. Since those HyperCore precompile reads can fall back to `0` in historical/timetravel contexts, I think restoring `timetravel: false` is the right fix here. Could we merge this?

RohanNero (approved):

RohanNero:
Thanks for the PR and apologies for the delay. Hopefully, this will fix it

PR #19181, opened 2026-05-08
Fix TONCO Router reference in TVL calculation

damnnou:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Tonco's total value locked (TVL) calculation to account for additional token sources, providing more accurate metrics.


RohanNero:
Thanks for the PR

PR #19249, opened 2026-05-12
Fix(Beefy): switch to /vaults/all endpoint and dedup CLM exposures

erenjaegar77:
**Summary**

- Switch the Beefy vault list source from `https://api.beefy.finance/vaults` to `https://api.beefy.finance/vaults/all`. The previous endpoint returns only `type: standard` vaults (~3,976), excluding `cowcentrated` (CLM) and `gov` (reward pool) entries entirely. `/vaults/all` exposes the full set (~5,482), letting the adapter see CLMs and their reward pools.

- Skip gov reward pools (`-rp`) whose `tokenAddress` is already referenced by a standard wrapper vault (`-vault`). Both reference the same CLM cow* share, and the existing `isConcLP` symbol-based branch resolves either reference to the CLM's full `balances()` — so when both exist they double-count the same underlying. The wrapper already covers the CLM TVL; dropping the duplicate RP entries removes ~$8M of double-counted TVL across 241 CLMs.

**Test plan**

- [ ] `node test.js projects/beefy/index.js` runs cleanly
- [ ] Total TVL: $134.92M (down from $143.15M pre-dedup; gap vs Beefy app driven by ~$9M of standard-vault underlying tokens DefiLlama can't price)

## Summary by CodeRabbit

* **Bug Fixes**
  * Vault data retrieval now includes vault type to enable more accurate filtering.
  * TVL calculations updated to correctly account for concentrated-style vault exposures by adding their underlying token amounts and excluding specific vault types from standard aggregation.
  * Non-staking token balance aggregation simplified to pair each vault’s primary token with its reported balance, reducing double-counting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/DefiLlama/DefiLlama-Adapters/pull/19249?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/projects/beefy/index.js
+++ b/projects/beefy/index.js
@@ -127,7 +127,7 @@ const beefyChainNameMapping = {
 };
 
 async function fetchVaultData() {
-  const vaultsResponse = await utils.fetchURL('https://api.beefy.finance/vaults');
+  const vaultsResponse = await utils.fetchURL('https://api.beefy.finance/vaults/all');
   const vaults = vaultsResponse.data;
 
   // sdk.log('Raw API response sample:', JSON.stringify(vaults[0], null, 2));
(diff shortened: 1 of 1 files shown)


erenjaegar77:
Adding the report of the vaults that can't be priced, 
[Unpriced Report](https://github.com/user-attachments/files/27655295/unpriced_report.txt)

RohanNero:
Thank you for the PR

PR #19041, opened 2026-04-30
feat(pool-party): switch TVL adapter to Send validator API (Phase 1)

SendPeter:
## Summary

Replaces the `projects/pool-party/index.js` adapter with a v2 implementation that reads from Pool Party's new public API on Send Foundation's validator. The previous adapter used the third-party Lighthouse explorer, which only exposes Canton Coin (CC) balances per party — it missed the stablecoin sides of every Pool Party pool, underreporting TVL by ~99.9%.

Live page result: **$458 → ~$1.31M** (per current mainnet snapshot).

## What changed

- **Data source:** `lighthouse.cantonloop.com` → `api-mainnet.cantonwallet.com/canton/pool-party/public/v1` (Send Foundation's own validator). Protocol-owned source instead of third-party explorer.
- - **Coverage:** CC + USDCx + CUSD (was CC-only). All three are read from the validator's `/tvl` endpoint.
- - **Pricing:** CC priced via existing `coingecko:canton-network`. USDCx (bridged USDC on Canton) and CUSD (Send's privacy stablecoin) priced as USDC (`coingecko:usd-coin`) — neither has a CoinGecko listing of its own. Methodology field is explicit about this.
## Companion PR

Send Foundation built the public API in `0xsend/canton-monorepo#3512` (Phase 1 — TVL, Volume, Fees `totalFees`). Tracking issue: `0xsend/canton-monorepo#3481`.

The API is live at `https://api-mainnet.cantonwallet.com/canton/pool-party/public/v1` (no auth, 60 RPM/IP, 5-min snapshot cadence, `Cache-Control: public, max-age=300, s-maxage=900, stale-while-revalidate=3600`).

## Verification

```bash
curl -s 'https://api-mainnet.cantonwallet.com/canton/pool-party/public/v1/tvl' | jq
```

Returns:

```json
{
  "asOf": "2026-04-30T21:46:27.032Z",
  "tvl": {
    "Amulet": "3167385.6643863726",
    "USDCx": "359442.1494694904",
    "481871d4-ca56-42a8-b2d3-4b7d28742946": "480511.519900348"
  },
  "pools": { ... }
}
```

Resolves to ~$475K CC + $359K USDCx + $480K CUSD ≈ **$1.31M total TVL**.

## Notes

[...]

PR #19355, opened 2026-05-21
fix(pool-party): source TVL from CCTools to include stablecoin pools

SendPeter:
Per @RohanNero's note on #19041 (option 2 — a neutral third-party indexer covering non-CC balances), this switches the Pool Party adapter to CCTools (`api.cctools.network`), an independent Canton DeFi aggregator not affiliated with Send.

CCTools indexes the stablecoin sides (USDCx, CUSD) that the Lighthouse API does not expose, so all three live pools are now counted: CC/USDCx, CC/CUSD, CUSD/USDCx.

TVL: ~$180k → ~$1.19M (current snapshot).

Note: CCTools serves pre-computed per-pool USD, so the adapter reports via `addUSDValue`. Happy to move to token-level balances if you'd prefer — the CCTools API would need a per-token-reserve endpoint for that.

## Summary by CodeRabbit

* **Updates**
  * TVL now aggregates reserves from all live pools via an external pools API, giving a more complete and accurate total value locked (now includes stablecoin sides).
  * Token pricing now maps pool tokens to market identifiers for correct USD valuation across assets.

* **Documentation**
  * Methodology updated to describe the live-pools raw-reserves data source and the new valuation approach.

### DIFF
--- a/projects/pool-party/index.js
+++ b/projects/pool-party/index.js
@@ -2,6 +2,9 @@ const { get } = require('../helper/http')
 
 const POOLS_URL = 'https://api.cctools.network/api/markets/send/pools'
 
+// CCTools rejects axios's default User-Agent; identify ourselves explicitly.
+const FETCH_OPTS = { headers: { 'User-Agent': 'DefiLlama-Adapter (pool-party)' } }
+
 // Canonical Canton token IDs → CoinGecko slugs.
 // USDCx (bridged USDC) and CUSD (Send privacy stable) priced as USDC; neither
 // has its own CoinGecko listing.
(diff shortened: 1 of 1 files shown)


projects/pool-party/index.js
```diff
@@ -1,28 +1,23 @@
...
+// (USDCx, CUSD) that the Lighthouse API does not expose.
+const POOLS_URL = 'https://api.cctools.network/api/markets/send/pools'
 
 async function tvl(api) {
-  for (const party of parties) {
-    const { balance } = await get(BASE_URL + party + '/balance')
-    api.addGasToken(Number(balance.total_coin_holdings) * 1e18)
+  const pools = await get(POOLS_URL)
+  for (const pool of pools) {
+    if (pool.status === 'live') api.addUSDValue(Number(pool.tvlUsd))
```
RohanNero: Hey, thanks for the PR. Could you please use token-level balances instead?

RohanNero (changes requested):

RohanNero (approved):

RohanNero:
Thanks again for the updates and PR

PR #19408, opened 2026-05-26
refactor(PumpSpace): optimize staking helper and exclude KRILL token

bluewhale-saito:
🛠️ Update Description (For Existing Project)
Excluded the KRILL token and its corresponding MasterChef contract from the staking calculation.

Refactored multiple staking helper calls into a single array-based staking function to match DefiLlama's standard coding pattern and improve code clarity.

Tested locally using node test.js projects/pumpspace/index.js and confirmed it successfully fetches the TVL and remaining staking assets (SHELL, SBWPM, SADOL, PEARL) without any errors.

## Summary by CodeRabbit

* **Refactor**
  * Streamlined Avalanche staking TVL calculation to use a grouped staking helper, improving efficiency and maintainability without changing overall TVL logic.

[Review Change Stack](https://app.coderabbit.ai/change-stack/DefiLlama/DefiLlama-Adapters/pull/19408?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


FelixBruguera:
Hi @bluewhale-saito, is there any particular reason for removing KRILL? if not, we would rather keep it in the adapter in case we need to refill historical data. Also, I think it's better to keep the previous staking calls with explicit masterchef-token pairs

bluewhale-saito:
> Hi @bluewhale-saito, is there any particular reason for removing KRILL? if not, we would rather keep it in the adapter in case we need to refill historical data. Also, I think it's better to keep the previous staking calls with explicit masterchef-token pairs


Thank you for the review! 

I agree with your point about keeping the historical data intact. I've updated the code to put KRILL back, and I also reverted the staking logic back to the previous style with explicit masterchef-token pairs as requested. 

Everything is pushed and ready for your check. Thanks again!

PR #19495, opened 2026-05-31
Add ClaimRush (Base) TVL adapter

claimrush:
#### Please enable "Allow edits by maintainers" while putting up the PR.

This PR adds a TVL adapter for **ClaimRush** under `projects/claimrush/index.js`. Adapter computes TVL purely from on-chain reads (no fetch / no third-party API), via the canonical `sumTokens2` + `staking` helpers in this repo. The companion listing-metadata PR for `DefiLlama/defillama-server` (`defi/src/protocols/data6.ts`, id `"7960"`) is queued for submission after this adapter PR merges.

---

##### Name (to be shown on DefiLlama):
ClaimRush

##### Twitter Link:
https://x.com/claimrushgame

##### List of audit links if any:
ClaimRush's public-facing security policy + disclosure pointer is at:
https://github.com/claimrush/claimrush/blob/main/SECURITY.md

The v1.0.0 internal security architecture (pre-implementation audit, global invariants, onchain threat map, CI security gates, verification + audit plan, security-test-matrix, third-party trust inventory, etc.) is maintained as an internal documentation set and shared directly with researchers / auditors / reviewers on request during coordinated disclosure — see the "Security documentation" section of SECURITY.md. Reviewer-access copies available on request to `security@claimru.sh`.

External third-party audit is scheduled post-soak (Phase 8 — Freeze-and-Burn). Will update `audits: "2"` and add the external audit URL via a follow-up `data6.ts` PR once it lands.

##### Website Link:
https://claimru.sh

##### Logo (High resolution, will be shown with rounded borders):
https://claimru.sh/brand/claimrush-defillama-512.jpg

512×512 JPG, canonical ClaimRush Ink background (`#0A0D14`) baked in (matches DefiLlama's `${baseIconsUrl}/<slug>.jpg` convention — JPG, no transparency, branded surface). 256×256 variant also available at `/brand/claimrush-defillama-256.jpg`. The canonical transparent-background SVG / PNG variants used by all other surfaces live at `/brand/claimrush-icon-color.svg` and `/brand/claimrush-icon-color-{64,200,256}.png`.

##### Current TVL:
[...]

PR #19496, opened 2026-06-01
claimrush: surface locked CLAIM as tvl (was staking)

claimrush:
Small follow-up to #19495 (the ClaimRush adapter, merged earlier today as `8401761`) — refines the bucket shape so the headline TVL number matches the productive ETH-yield-bearing position.

## Why

Current state on `main` after #19495 was merged:

| Bucket | Content | Live value |
|---|---|---|
| `tvl` | `() => ({})` (empty) | $0 |
| `pool2` | Aerodrome WETH/CLAIM LP positions | ~$218k |
| `staking` | locked CLAIM in `VeClaimNFT` | ~$19k |

This makes the headline TVL on `defillama.com/protocol/claimrush` read **$0**, with both productive buckets shown as side metrics. The locked CLAIM (`staking`) is actually the only ETH-yield-bearing position in the protocol — every Mine takeover routes 25% of its ETH to veCLAIM lockers pro-rata to ve-share — so it's a more accurate denominator for ETH-royalty APR display than either the empty `tvl` bucket or the `pool2` bucket would be.

ClaimRush has three positions with very different economic semantics:

| Position | Capital type | Yield earned | Right DefiLlama bucket |
|---|---|---|---|
| `VeClaimNFT` (locked CLAIM) | User capital, ≤1-year linear-decay lock | **ETH royalties** (25% of every Mine takeover) | `tvl` (productive ETH-yield position) |
| `LpStakingVault7D` (user LP) | User capital, 7-day rolling stake | CLAIM emissions (token issuance) | `pool2` (protocol-native LP) |
| `GenesisLPVault24M` (protocol seed) | Protocol-owned, 24-month time-lock | CLAIM emissions | `pool2` (protocol-native LP) |

Same pattern Convex / Olympus / Pendle / various LSDs use for protocols where the lock IS the yield-bearing position (as opposed to vanilla governance staking, which the `staking` bucket was designed for).

## What changed

```diff
- async function tvl(api) { ... }   // was renamed to pool2 at merge time
+ async function tvl(api) {
+   return sumTokens2({
+     api,
+     tokensAndOwners: [[CLAIM, VE_CLAIM]],
+   });
+ }
  ...
  module.exports = {
    methodology: '... (updated to cover both buckets) ...',
    base: {
-     tvl: () => ({}),
+     tvl,
      pool2,
-     staking: staking(VE_CLAIM, CLAIM),
    },
  };
```

- `tvl` now reads the CLAIM ERC-20 balance held by `VeClaimNFT` directly (raw token balance, no LP-detection trigger). This is the same value the `staking` helper was reading.
- `pool2` is unchanged (LP positions stay where they belong).
- `staking` export dropped (the value moved into `tvl`).
- `methodology` string rewritten to document the bucket distinction explicitly.

## Smoke output (local, 2026-06-01 ~00:48 UTC)

```
$ node test.js projects/claimrush/index.js

--- tvl ---
CLAIM                     19.17 k
Total: 19.17 k

--- pool2 ---
WETH                      111.40 k
CLAIM                     107.55 k
Total: 218.95 k

------ TVL ------
base                      19.18 k
base-pool2                218.95 k
total                    19.17 k
```

Live behavior cross-check:

- `cast call CLAIM "balanceOf(address)" VE_CLAIM` → `7172182607308538278169631 wei` = `7,172,182.61 CLAIM` (matches `tvl` bucket at ~$0.0027/CLAIM ≈ $19.17k)
- Pool reserves: `40,967,635.05 CLAIM + 54.36 WETH` (matches `pool2` underlying unwrap)

I have "Allow edits by maintainers" enabled on this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed TVL calculation to accurately report total value locked instead of displaying zero.

* **Documentation**
  * Updated methodology description to clarify ETH-royalty denomination and LP position reporting.

### DIFF
--- a/projects/claimrush/index.js
+++ b/projects/claimrush/index.js
@@ -1,11 +1,17 @@
 const { sumTokens2 } = require('../helper/unwrapLPs');
-const { staking } = require('../helper/staking');
 
 const CLAIM = '0x059D278233fEC14CB6D1A74E6FB482BC3f91ADbf';
 const VE_CLAIM = '0x876Da22a5bBEe4f8963b791631D2cAC5199389eE';
 const LP_STAKING_VAULT_7D = '0xafdbB422CF75D6f2C557BDD2EF955c518b086271';
 const GENESIS_LP_VAULT_24M = '0x1532F33e53680f89d083a0bf5baedcCCD2E7267a';
 const AERODROME_WETH_CLAIM_POOL = '0x7274599ec9DBf15D474A6FB18aA285aE001d87Aa';
+
+async function tvl(api) {
+  return sumTokens2({
+    api,
+    tokensAndOwners: [[CLAIM, VE_CLAIM]],
+  });
+}
 
 async function pool2(api) {
   return sumTokens2({
@@ -20,10 +26,9 @@ async function pool2(api) {
 
 module.exports = {
   methodology:
-    'Pool2 = Aerodrome v2 vAMM WETH/CLAIM LP tokens custodied by LpStakingVault7D (7-day rolling staking position for active stakers) plus GenesisLPVault24M (24-month time-locked genesis seed liquidity). LP tokens are unwrapped into their underlying WETH and CLAIM reserves. Staking bucket = CLAIM principal locked in VeClaimNFT (voting-escrow NFT, max 1-year linear-decay locks) to receive a pro-rata share of ETH royalties from every Mine takeover.',
+    'TVL = CLAIM principal locked in VeClaimNFT (voting-escrow ERC-721, max 1-year linear-decay locks). veCLAIM holders receive a pro-rata share of ETH royalties from every Mine takeover (25% of each takeover ETH flows to lockers, weighted by veCLAIM share). This is the productive long-term-locked, ETH-yield-bearing position and the correct denominator for the protocol\'s ETH-royalty APR. Pool2 = Aerodrome v2 vAMM WETH/CLAIM LP tokens custodied by LpStakingVault7D (7-day rolling stake for active LP stakers) and GenesisLPVault24M (24-month time-locked genesis seed liquidity). LP positions earn CLAIM emissions (token issuance) rather than ETH royalties, so they are reported under pool2 (protocol-native LP) per DefiLlama convention. LP tokens are unwrapped into their underlying WETH and CLAIM reserves.',
   base: {
-    tvl: () => ({}),
+    tvl,
     pool2,
-    staking: staking(VE_CLAIM, CLAIM),
   },
 };


claimrush:
Adding context on the bot's `tvl` vs `staking` flag, since the editorial call here matters:

CodeRabbit is correct that fee-earning ve-tokens conventionally live in `staking` on DefiLlama (vePENDLE, veCRV, vlCVX). The reason this PR keeps locked CLAIM in `tvl` is that **veCLAIM is the protocol** — not a side-of-house governance lock. The entire product concept is: lock CLAIM into `VeClaimNFT` (`0x876Da22…389eE`) to receive ETH royalties from every Mine takeover via `ShareholderRoyalties` (`0x74eDd39E…D2a91`). It is the only productive yield-bearing position the protocol offers. The LP is liquidity infrastructure for entry/exit, not the product, and it lives on Aerodrome's AMM anyway.

That makes the structural shape different from the protocols the convention was written for:

| protocol | productive position | ve-token role |
|---|---|---|
| Pendle | PT/YT deposits ($2B in `tvl`) | governance + boost (`staking`, $200M) |
| Curve | swap pools ($1.5B in `tvl`) | governance + admin fees (`staking`) |
| Convex | booster LPs ($1B in `tvl`) | governance + emissions (`staking`) |
| ClaimRush | **veCLAIM itself** | productive ETH-royalty position = the product |

For Pendle / Curve / Convex, putting the ve-token in `staking` is correct because the ve-token is auxiliary to the protocol's core deposit surface. For ClaimRush there IS no other deposit surface — moving veCLAIM out of `tvl` doesn't move productive capital into a more honest bucket, it just hides the protocol's only productive capital.

The closest functional precedent on Base, Aerodrome, sidesteps this by being categorized as `Dexs` and reporting AMM-pool TVL only — veAERO isn't tracked at all on its DL page. That works for a DEX with $137M of LP. For a yield protocol on day 3 with veCLAIM as the entire economic surface, the equivalent would be to not list at all, which defeats the purpose.

The [DefiLlama docs](https://docs.llama.fi/list-your-project/what-to-include-as-tvl) carve this out: "check if the locked assets provide utility beyond governance (such as **claiming protocol fees**) and categorize them according to their primary function." veCLAIM's primary function is claiming ETH protocol fees — and that's literally all it does.

@FelixBruguera — flagging this as the design rationale rather than litigating the bot. The shape in this PR (`tvl` = locked CLAIM = productive yield position, `pool2` = liquidity infrastructure) is what reflects the actual product.

FelixBruguera:
Hi, I understand your position and that showing 0 TVL is not ideal for your protocol, but our methodology is to count the protocol's own token under staking, it's not just governance or unproductive staking, but any position where the locked value is in the protocol's own token is counted under staking. You can read more about the methodology here: https://docs.llama.fi/faqs/frequently-asked-questions#what-is-staking-tvl

claimrush:
Thanks Felix — that FAQ entry is the clarification I needed. The hack-loss framing (core TVL = market trust in non-native deposits; staking TVL = riskless because the protocol controls its own token supply) is principled and consistent, and veCAKE is a clean structural analogue. Accepting the editorial call; the live #19495 shape (`tvl: () => ({})` / `pool2` = LP / `staking` = locked CLAIM) is correct as-is.

Following up shortly with a fees adapter for the ETH royalty stream emitted by `ShareholderRoyalties` — that will let the yields page surface the veCLAIM APR even with the $0 core TVL, which is the metric users actually care about for this position.

PR #19482, opened 2026-05-30
fix ssi protocol: compute TVL from on-chain basket balances via getBa…

BhariGowda:
Fixes #17720

## What was wrong
The adapter called `getBasket()` but never used the result to compute TVL, so TVL was always 0.

## What this fixes
- Fetches basket composition via `getBasket()` for MAG7.ssi, DEFI.ssi, MEME.ssi
- Maps basket symbols to coingecko IDs (underlying assets are native/cross-chain, `addr` is empty string)
- Divides by token decimals to get real amounts as required for coingecko-priced tokens
- Sets `start: 1733011200` (Dec 2024 launch) to avoid unnecessary historical backfill

## Test output
node test.js projects/ssi-protocol/index.js

base: $84.72M TVL
BTC   27.22M
ETH   10.91M
BNB   10.61M
XRP    9.55M
SOL    9.20M
DOGE   8.83M
ADA    8.40M

## Summary by CodeRabbit

* **Refactor**
  * SSI protocol adapter refocused to a single Base entrypoint instead of multi-chain TVL outputs.
  * TVL calculation now derives balances from on-chain basket composition queries and aggregates component token totals.
  * Module exports restructured with expanded methodology, a fixed start timestamp, and misrepresented-tokens flag.

[Review Change Stack](https://app.coderabbit.ai/change-stack/DefiLlama/DefiLlama-Adapters/pull/19482?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #19512, opened 2026-06-02
fix ssi protocol: compute TVL from totalSupply x market price (v2)

### LINKED ISSUE #17720: fix soso index tvl
compute it using on chain balances: https://github.com/DefiLlama/DefiLlama-Adapters/blob/main/projects/ssi-protocol/index.js

BhariGowda:
Fixes #17720

Following feedback from @FelixBruguera on #19482 — updated approach.

## What changed
Previous PR used `getBasket()` which is wrong. This version computes TVL from the actual `totalSupply` of each SSI index token, priced via their contract address on Base.

SSI tokens are custodied index tokens — underlying assets (BTC, ETH, SOL etc.) are held off-chain by Cobo/Ceffu. The correct on-chain TVL is the circulating supply of each token × its market price.

## Test output
node test.js projects/ssi-protocol/index.js

MAG7.ssi   82.82M
DEFI.ssi   10.27M
MEME.ssi    0.60M
Total:     93.69M

## Summary by CodeRabbit

* **Refactor**
  * SSI Protocol TVL is now reported only on the Base blockchain; previously tracked networks are no longer reported.
  * TVL calculation methodology updated to a direct per-token supply approach.
  * Exported metadata updated (including a start date and misrepresented-tokens flag) to reflect the new reporting shape.

### DIFF
--- a/projects/ssi-protocol/index.js
+++ b/projects/ssi-protocol/index.js
@@ -1,68 +1,47 @@
-const sdk = require('@defillama/sdk')
-const { getBlock } = require('../helper/http');
-
 const abi = {
-  getBasket: 'function getBasket() view returns ((string chain, string symbol, string addr, uint8 decimals, uint256 amount)[])'
+  getBasket: 'function getBasket() view returns ((string chain, string symbol, string addr, uint8 decimals, uint256 amount)[])',
 }
 
-const ssi_tokens = [
-  // MAG7.ssi
-  '0x9E6A46f294bB67c20F1D1E7AfB0bBEf614403B55',
-  // DEFI.ssi
-  '0x164ffdaE2fe3891714bc2968f1875ca4fA1079D0',
-  // MEME.ssi
-  '0xdd3acDBDc7b358Df453a6CB6bCA56C92aA5743aA'
+const SSI_TOKENS = [
+  '0x9E6A46f294bB67c20F1D1E7AfB0bBEf614403B55', // MAG7.ssi
+  '0x164ffdaE2fe3891714bc2968f1875ca4fA1079D0', // DEFI.ssi
+  '0xdd3acDBDc7b358Df453a6CB6bCA56C92aA5743aA', // MEME.ssi
 ]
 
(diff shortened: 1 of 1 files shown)


FelixBruguera:
Thanks for the PR, but this still doesn't resolve the issue. The goal of the issue is to track the balances of the underlying assets of each index onchain (BTC, ETH, BNB, XRP, SOL, DOGE and ADA for the MAG7 for example). If this isn't possible it's better to keep the current approach

BhariGowda:
Thanks for clarifying. So the original approach using getBasket() was correct  PR #19482 already did this and showed $84.7M TVL (BTC 27M, ETH 10M, BNB 10M, XRP 9M, SOL 9M, DOGE 8M, ADA 8M). The issue was only that the old adapter called getBasket() but never used the result. Would you like me to restore that approach in a new PR?

FelixBruguera:
> Thanks for clarifying. So the original approach using getBasket() was correct PR #19482 already did this and showed $84.7M TVL (BTC 27M, ETH 10M, BNB 10M, XRP 9M, SOL 9M, DOGE 8M, ADA 8M). The issue was only that the old adapter called getBasket() but never used the result. Would you like me to restore that approach in a new PR?

No, the current adapter works fine, but the issue is about trusting the getBasket() view. For example with dexes what we usually do is pull all the pools created by a factory and call balanceOf(token, pool) on each one, instead of trusting a view function we read the balances directly from the pools that hold them. That's what the SSI adapter should do if possible

BhariGowda:
Thanks for the clarification. After investigating further, I confirmed the underlying assets (BTC, ETH, SOL etc.) are held off-chain by Cobo/Ceffu custodians there are no on-chain pool contracts holding them directly. The assetLocking contract (0x935A4B1F) only holds dust/spam tokens with $0 value. So querying balanceOf(token, pool) isn't possible here. Given your note that "if this isn't possible it's better to keep the current approach," should we close this issue as won't fix?

PR #19538, opened 2026-06-03
feat: add ADA Harvest (Cardano PlutusV3 flash loan vault + yield aggregator)

upggr:
## ADA Harvest

**Website:** https://ada-harvest.com  
**Docs:** https://ada-harvest.com/flash-loan-docs  
**Chain:** Cardano mainnet  
**Category:** Yield Aggregator + Flash Loans

### What is ADA Harvest?

ADA Harvest is a non-custodial PlutusV3 vault on Cardano mainnet that earns yield through three sources:

1. **Flash loan fees** — borrowers pay 0.50% per transaction, fee stays in vault
2. **DeFi yield** — AI council deploys idle ADA to whitelisted protocols (Minswap, Liqwid)
3. **ADA staking** — vault always delegated to a stake pool (currently VIPER, 0% margin)

### Flash loans

First atomic flash loan implementation on Cardano. Borrowers can take uncollateralised ADA loans within a single transaction — repayment is enforced by the Plutus validator, not by trust.

### Adapter

Counts all ADA locked in the vault smart contract:

- Vault address: `addr1w89u29sdr6aazr6v9zhdk6vh4az75enlzph2372v3tq55usz9ale6`
- Script hash: `cbc5160d1ebbd10f4c28aedb6997af45ea667f106ea8f94c8ac14a72`

### Checklist

- [x] My code is error-free and follows the repository style
- [x] I have tested the adapter locally
- [x] The adapter uses `sumTokensExport` for Cardano (same pattern as Liqwid, Minswap)
- [x] Protocol is live on mainnet

## Summary by CodeRabbit

* **New Features**
  * Added ADA Harvest DeFi adapter for the Cardano blockchain with Total Value Locked (TVL) tracking capabilities and methodology documentation.

PR #19540, opened 2026-06-04
fix(ada-harvest): update to current vault address + lovelace-only TVL

upggr:
Follow-up to #19538 (merged).

The merged adapter pointed at an earlier ADA Harvest vault address that has since been migrated on mainnet — that old address is now drained (0 ADA), so the adapter reports 0 TVL.

This PR:
- Updates `VAULT_ADDRESS` to the current live base vault: `addr1zyn8hlf0wqkcvsdu89yvnt5kqnf69560xej36vrt4ue6uzcgucktr6zta3rtuqlqam4395sry3vl5c7ss08m50qtnj6sjernpa`
- Applies the CodeRabbit review from #19538: restricts `sumTokensExport` to `tokens: ['lovelace']` (no non-ADA double counting), adds `doublecounted: true`, and an accurate methodology string.

Verify on Cardanoscan: the vault holds ~90 ADA and is delegated to a stake pool.

## Summary by CodeRabbit

* **Bug Fixes**
  * ADA Harvest now reports TVL using ADA-only (lovelace) balances.
  * Vault updated to the new PlutusV3 deployment address.
  * Vault is marked as double-counted and methodology clarified to exclude non-ADA assets, preventing TVL inflation from downstream deployments.

### DIFF
--- a/projects/ada-harvest/index.js
+++ b/projects/ada-harvest/index.js
@@ -2,10 +2,10 @@ const { sumTokensExport } = require('../helper/chain/cardano');
 
 // ADA Harvest - Cardano PlutusV3 flash loan vault + yield aggregator
 // Vault address (base address, delegates to a stake pool):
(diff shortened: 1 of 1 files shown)


RohanNero:
Thanks for the update

PR #19544, opened 2026-06-04
Update Exponent TVL exposure accounting

sablevsky:
## Summary
This PR improves Exponent TVL accounting in two key areas:

1. Adds AUM(TVL) for Exponent strategy vaults.
2. Changes Core vaults exposure to reflect the actual base assets held by the vaults, instead of representing everything in quote asset terms.

## Core vault exposure
Previously, core vault exposure was normalized into the quote asset. While that made the output easier to compute, it was not a valid representation of the assets actually held by the vault.

For example:
- consider a vault for ONyc token
- its quote asset is USD
- its underlying/base asset is ONyc 

Previously, this vault’s TVL was effectively reported in USD(USDe) terms.
With this change, the adapter reports the vault’s exposure as ONyc , which is a more accurate representation of the assets backing the vault.

## Strategy vault AUM
This PR also adds AUM for strategy vaults, so managed strategy capital is included in total TVL.

## Result
This makes Exponent TVL more complete and economically more accurate.

## Summary by CodeRabbit

* **New Features**
  * Strategy-vault AUM aggregation added with support for many position types (CLMM, orderbook, Kamino farms/lending, Loopscale, reserves, yield positions), per-vault breakdowns, and managed-vault AUM reporting.
  * On-chain prices, PDAs, and account parsers added to support richer valuation and reward projection.

* **Refactor**
  * TVL calculation rewritten to use on-chain fixed‑point exchange rates and external payload mappings for SY→quote→base conversions.

* **Documentation**
  * Public methodology updated to describe the new conversion and AUM aggregation flow.


projects/exponent/index.js
```diff
@@ -1,62 +1,211 @@
...
+    if (!amount) return
+
+    api.add(payload.baseMint, amount)
+  })
+}
+
+async function addManagedVaultAum(api) {
+  // Managed strategies can deploy capital into external protocols that are not
+  // visible from core SY/PT/YT supply alone, so fetch that exposure separately.
+  const managedAum = await getConfig(
```
RohanNero: Sorry, we are no longer accepting tvl reported from APIs. Is it possible to derive the managed aum without it?

RohanNero (changes requested):

projects/exponent/index.js
sablevsky: Hey!
I managed to do it, but unfortunately, it ended up requiring a lot of code.
The Strategy Vaults program contains a lot of logic, including integrations with all Exponent programs, as well as external protocols such as Kamino, Loopscale, and others.

sablevsky: @RohanNero 
I also noticed that Defillama can't display the TVL in the rkuSOL token (rkubjTrZYioRSeXwDnhwGQzvW3qkcin72JSxUt3WMVp).
Because of this, we're missing approximately $1.97m in TVL.
Could you add support for this?

sablevsky:
Hey @RohanNero !
Sorry for the ping, I was not sure if my review request was missed. Let me know if there is anything else that needs to be done to resume the integration.
Thanks!

RohanNero:
@sablevsky Hey, apologies for the delay. I will finish reviewing this today

RohanNero:
Thank you for the PR, I've added exponent vault tvl in #19696 

Sorry, but we'd prefer to continue tracking quote mints since counting base mints would require an additional API dependency and token conversion. Also, it will become difficult to maintain in its current form and new external integrations/strategies would require updates to the adapter.

PR #19549, opened 2026-06-04
chutes: replace stale heroku export with self-contained RPC adapter

siteauditor:
DefiLlama currently shows $0 TVL for Chutes (id 7346). The adapter never computed anything itself: index.js delegated to the heroku-api elastic cache via getExports, while an unused implementation sat in api.js requiring @polkadot/api, which is not a repo dependency.

This PR makes the adapter self-contained: it reads the pool reserves (SubtensorModule.SubnetTAO / SubnetAlphaIn for netuid 64) over plain JSON-RPC using helper/http, both reads pinned to the same finalized block. JSON-RPC error bodies throw instead of being misread as an empty key, so an upstream failure never records a false $0. The now-dangling chutes entry in the sushi-analytics-v2 adapterMapping (pointing at the deleted api.js) is removed too.

`node test.js projects/chutes/index.js`: ~$84M (TAO leg ~$42M + alpha leg ~$42M) at the time of writing, consistent with taostats. No new dependencies.

Note: the Chutes listing in defillama-server has gecko_id null; if the site still shows $0 after this merges, that may need a one-line server-side fix as well. Happy to send that PR too.

## Summary by CodeRabbit

* **Refactor**
  * Reworked Chutes TVL calculation and updated methodology; legacy Chutes API module removed.

* **New Features**
  * Added support for Astar DApps Staking, Injective Orderbook, and Lisa adapters.

* **Chores**
  * Removed Chutes from scheduled adapter mappings.


siteauditor:
Anything I can clarify on this one? Latest harness run: ~$84M (TAO leg + alpha leg), consistent with taostats. If you'd prefer the sushi-analytics adapterMapping cleanup split out of this PR, happy to do that.

g1nt0ki:
@siteauditor thanks for the PR

siteauditor:
Thanks @g1nt0ki! If it's useful while you're in here, there are four more in the same Bittensor vertical, all green: #19550 (native dTAO subnet pools, ~$390M, currently untracked), #19577 (hatom-tao-bridge $0 fix), #19591 (wTAO on Ethereum, ~$24M, untracked), #19592 (canonical TAO on Solana). Happy to adjust any of them.

PR #19554, opened 2026-06-04
Remove LunarBase TVL adapter

NebulaNomadPixel:
## Summary

Remove the LunarBase TVL adapter from `DefiLlama-Adapters`.

LunarBase should be represented through DEX volume only using the dimension adapter flow.

## Why

The TVL adapter reports balances for LunarBase pools, but the intended DefiLlama integration is volume-only. Removing this adapter prevents LunarBase from being tracked with TVL data from this repository.

## Validation

- `git diff --check`

## Summary by CodeRabbit

* **Revert**
  * Removed Lunarbase TVL integration from the application.


RohanNero (changes requested):
Sorry, we don't remove adapters for reasons described here: https://github.com/DefiLlama/DefiLlama-Adapters/discussions/13020

PR #19563, opened 2026-06-05
Fusion (by IPOR) add option to properly calculate historical TVL

rav-ipor:
**_EXTRA REQUEST:_** Please recalculate Fusion (by IPOR) TVL value from 9th May 2026 till current date to fix TVL value.

There was an issue with vault list in `ipor-abi` repo which is the source for the vault list on DefiLlama.
Some vaults were duplicated. It was merged on 9th May 2026 and from this date TVL value was not counted correctly. We fixed this in repo and in process that is creating vault list. We have also added a uniqueness check for the vault address to avoid making this mistake in the future.

What was changed:
- Add `permitFailure: true` to properly calculate historical TVL. Some vaults will be not present in the past so to not throw an exception they will be skipped.
- Add some optional debug, turned off by default.
- Add deduplicate by PlasmaVault address

## Summary by CodeRabbit

* **Bug Fixes**
  * TVL calculation at historical blocks now skips missing or non-deployed vaults and avoids throwing errors.
  * Vaults are deduplicated by address (case-insensitive) to prevent double-counting.

* **Improvements**
  * Optional debug logging now reports per-vault and total USD contributions for clearer monitoring and troubleshooting.

### DIFF
--- a/projects/ipor-fusion/index.js
+++ b/projects/ipor-fusion/index.js
@@ -39,10 +39,28 @@ async function tvl(api) {
     return {};
   }
   
-  debugLog(`[Fusion (by IPOR)] Processing ${chainConfig.vaults.length} vaults on ${chain}:`);
-  
-  const calls = chainConfig.vaults.map((vault, index) => {
-    debugLog(`  Vault ${index + 1}/${chainConfig.vaults.length}: ${vault.PlasmaVault} (${vault.name || 'Unknown'})`);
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for letting us know and for the PR, I've refactored slightly and added the dedupe in #19569. We will refill the the chart from May 9th - now.

PR #19604, opened 2026-06-09
fix pondo tvl: include delegated stake

14MR:
Pondo's TVL on DefiLlama looked way off — we were showing around 648k ALEO when the protocol actually has roughly 24M staked. The adapter was only reading the liquid balance sitting in `pondo_protocol.aleo`, which is just the small buffer the protocol keeps on hand. But Pondo is a liquid-staking protocol, so almost all of the ALEO is delegated out to validators through five delegator programs (`delegator1-5.aleo`), and the adapter never looked there. So we were reporting maybe 3% of the real number.

I reworked the TVL calc to mirror how Pondo itself computes it — the same accounting behind the number on pondo.xyz. It starts with the liquid balance in the core program, adds each delegator's account + bonded + unbonding credits, and then subtracts the amounts set aside for pending withdrawals, since those are on their way out and no longer back pALEO.

While I was in there I also dropped the old `sumTokens` helper. It only ever read the `account` mapping, which is exactly what caused the undercount, and nothing else in the repo used it. I bumped the Aleo RPC timeout from 3s to 10s as well, since the adapter makes a handful more calls now.

Ran `node test.js projects/pondo/index.js` and it now comes out to about $743k (~24.34M ALEO), which matches pondo.xyz.

---

This is a fix to an already-listed protocol, not a new listing, so I've left the new-listing form out and just filled in the parts that apply:

Total ALEO backing pALEO: the liquid credits held by `pondo_protocol.aleo`, plus the account + bonded + unbonding credits of each of its five delegators (`delegator1-5.aleo`), minus the ALEO reserved for pending withdrawals. Read live from the Aleo RPC, so it's live-only (no timetravel).

No new npm dependencies, and `pnpm-lock.yaml` is untouched. "Allow edits by maintainers" is enabled.

### DIFF
--- a/projects/helper/chain/aleo.js
+++ b/projects/helper/chain/aleo.js
@@ -3,49 +3,22 @@ const axios = require('axios')
 const endpoint = 'https://api.provable.com/v1/mainnet'
 
 async function aQuery(path) {
-  const { data } = await axios.get(`${endpoint}${path}`, { timeout: 3000 })
+  const { data } = await axios.get(`${endpoint}${path}`, { timeout: 10000 })
   return data
 }
 
 /**
  * Fetch a value from a mapping within an Aleo program.
- * e.g., getProgramMappingValue('credits.aleo', 'account', 'aleo1...')
+ * Program ids are auto-resolved to their addresses by the API, so either a
+ * program id (e.g. 'delegator1.aleo') or an aleo1... address works as the key.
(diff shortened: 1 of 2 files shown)


RohanNero:
Apologies for the delay and thank you for the PR. I've updated pondo in #19770 to include delegated stake that has left `pondo_protocol.aleo`

PR #19614, opened 2026-06-09
realtoken-rmm-v3: price collateral via market oracle (fixes $0 TVL)

tarun-khatri:
### Problem
`realtoken-rmm-v3` (RealT RMM v3, Gnosis) reports **\$0 TVL** while the market is active with **~\$14.7M borrowed** — which is impossible for a healthy lending market.

The reserves are WXDAI, USDC and RealT "Wrapped USD" collateral tokens (RTW-USD-01, REUSD). WXDAI/USDC are ~100% utilized, so the aTokens hold ~0 underlying; the value actually locked is the RTW-USD-01 collateral (~31.2M, supplied by 2,238 addresses). DefiLlama has no market price for these RealT tokens (no DEX, not in the coins API), so they were valued at \$0 — and once WXDAI/USDC utilization reached ~100%, total TVL collapsed to \$0.

### Fix
The market exposes its own Aave price oracle (`0xb4AE809…`) that prices every reserve (RTW-USD-01 at \$1.00). This standalone adapter reads supply (aToken-held underlying) and debt (variable + stable debt-token supply) from the data provider and values both through that oracle. The sibling RealT market (`realtmarkets`) already uses the same oracle-based approach.

Moved out of the shared aave-fork registry into its own adapter because the registry path priced reserves through DefiLlama's token pricing (which can't price the RealT collateral), and its borrowed reader assumed a data-provider return layout this market does not use.

### Verification (Gnosis mainnet)
- Reserves — RTW-USD-01: supplied 31.19M / borrowed 0; USDC: supplied 8.84M / borrowed 8.84M; WXDAI: supplied 5.84M / borrowed 5.84M.
- Oracle `getAssetPrice`: RTW-USD-01 = \$1.00; WXDAI/USDC ≈ \$1.00.
- RTW-USD-01 is supplied by 2,238 distinct addresses and is not in the RealT community token API, so this does not double-count with the `realt-tokens` adapter.
- `node test.js projects/realtoken-rmm-v3`: TVL = \$31.25M, borrowed = \$14.67M.

## Summary by CodeRabbit

* **New Features**
  * Added support for RealT RMM v3 lending market on Gnosis, enabling TVL and borrowed-asset reporting in USD.

* **Chores**
  * Updated protocol registry by removing an outdated configuration entry related to the Aave integrations.

### DIFF
--- a/projects/realtoken-rmm-v3/index.js
+++ b/projects/realtoken-rmm-v3/index.js
@@ -27,9 +27,12 @@ async function getMarket(api) {
 }
 
 function addUsd(api, amounts, decimals, prices, baseUnit) {
+  const SCALE = 10n ** 18n; // keep fractional precision through integer math
+  const base = BigInt(baseUnit);
   amounts.forEach((amount, i) => {
-    const usd = (Number(amount) / 10 ** Number(decimals[i])) * (Number(prices[i]) / Number(baseUnit));
-    api.addUSDValue(usd);
+    const denom = 10n ** BigInt(decimals[i]) * base;
+    const usdScaled = (BigInt(amount) * BigInt(prices[i]) * SCALE) / denom;
+    api.addUSDValue(Number(usdScaled) / 1e18);
   });
 }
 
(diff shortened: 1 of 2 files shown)


tarun-khatri:
Addressed in dc72ed9b8: USD math now stays in BigInt through the arithmetic (amount × price scaled by 1e18 over the decimals/base denominator), converting to Number only on the final scaled value. Debt aggregation also sums as BigInt. Local run unchanged: TVL $31.25M, borrowed $14.68M.

@coderabbitai review

RohanNero:
Sorry for the delay and thank you for the PRs. I will add pricing for 0xd3dff217818b4f33eb38a243158fbed2bbb029d3 and 0x3390742ac0dce14ea6fcbd5ae02e2303c5d62ad9 on the server so it can stay in the registry

PR #19619, opened 2026-06-10
centrifuge: track JAAA on Solana, subtract from Ethereum to avoid dou…

0x4Graham:
…ble counting

Credit the JAAA SPL token's total supply to the Solana chain, valued at the EVM JAAA share price. Since the assets backing these Solana-issued shares already sit in the EVM vault and are counted in the Ethereum TVL, subtract the same value from Ethereum instead of using the blanket doublecounted flag.

**NOTE**

---

1. If you would like to add a `volume/fees/revenue` adapter please submit the PR [here](https://github.com/DefiLlama/dimension-adapters).

2. Once your adapter has been merged, it takes time to show on the UI. If more than 24 hours have passed, please let us know in Discord.
3. Sorry, We no longer accept fetch adapter for new projects, we prefer the tvl to computed from blockchain data, if you have trouble with creating a the adapter, please hop onto our discord, we are happy to assist you.
4. **For updating listing info** It is a different repo, you can find your listing in [this file](https://github.com/DefiLlama/defillama-server/blob/master/defi/src/protocols/data2.ts), you can edit it there and put up a PR
5. Please do not add new npm dependencies, do not edit/push `pnpm-lock.yaml` file as part of your changes

---
## Summary by CodeRabbit

* **New Features**
  * Added Solana TVL reporting for the JAAA share class, valuing Solana supply using EVM-derived price.

* **Bug Fixes**
  * Prevented cross-chain double-counting by subtracting the Solana-valued JAAA amount from Ethereum TVL in current (non-historical) runs.

* **Documentation**
  * Expanded TVL methodology to describe Solana valuation and the Ethereum subtraction approach.

### DIFF
--- a/projects/centrifuge/index.js
+++ b/projects/centrifuge/index.js
@@ -1,4 +1,5 @@
 const { getLogs2 } = require('../helper/cache/getLogs');
+const { getTokenSupplies } = require('../helper/solana');
 const ADDRESSES = require('../helper/coreAssets.json')
 
 const nullAddress = ADDRESSES.null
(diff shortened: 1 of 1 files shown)


0x4Graham:
Good catch — fixed in 44c01d280.

`getTokenSupplies` reads the mint's current supply and is not timestamp-aware (the Allium historical path in `helper/solana.js` only covers `sumTokens2` balances, not mint total supply), while the EVM vault reads are block-aware. During backfill that would subtract today's JAAA supply from past Ethereum balances.

Rather than disable timetravel for the whole adapter (which would lose backfill for all the EVM chains too), I gated the Solana credit **and** the matching Ethereum subtraction to live runs only (current UTC day, via `api.timestamp`). Historical points keep all JAAA backing on Ethereum — where it was before the Solana mint launched and where it's unmeasurable historically anyway — so EVM timetravel stays intact and the time series is consistent.

Verified: historical run makes zero `api.add` calls (skipped); live run credits JAAA to Solana and subtracts the equal value from Ethereum.

projects/centrifuge/index.js
```diff
@@ -153,7 +154,51 @@ const tvl = async (api) => {
...
+// Ethereum wraps the shared EVM tvl and removes the Solana-issued JAAA value,
+// whose backing is already reflected in the vault assets measured above.
+const ethereumTvl = async (api) => {
+  await tvl(api)
+  if (isHistorical(api)) return
+  const supply = await getJaaaSolanaSupply()
+  if (supply) api.add(JAAA_EVM, (-BigInt(supply)).toString())
+}
+
+module.exports.methodology = `TVL corresponds to the total USD value of tokens minted on Centrifuge. On EVM chains it is measured from the assets backing each share class vault. The JAAA share class is also issued on Solana; its SPL supply is valued at the EVM JAAA share price and shown on Solana, with the equivalent value subtracted from Ethereum to avoid double counting (its backing already sits in the EVM vault).`
```
RohanNero, line 198: Sorry, but if the Solana JAAA backing is already tracked on the ethereum side, then I don't think we should also track the SPL supply. We typically attribute tvl to the source chain where the collateral sits.

RohanNero (changes requested):

projects/centrifuge/index.js
0x4Graham: In the case of Solana, the acual fund shares / JAAA tokens are issued on Solana, the EVM tokens only represent accounting units.

Therefore the tracking is accurate to follow the SPL supply

RohanNero:
Apologies for the delay and the inconvenience, we are currently migrating RWA tokens to the new dashboard, so this adapter is essentially deprecated. The new page for centrifuge can be seen here: https://defillama.com/rwa/platform/centrifuge

The 200M is attributed to solana on the JAAA page: https://defillama.com/rwa/asset/JAAA

PR #19623, opened 2026-06-10
Rockawayx Add Morpho USDC Huma Cluster vault

Sithey:
## Chores
* Expanded RockawayX TVL aggregation to include an additional Ethereum Morpho vault.


RohanNero:
Thanks for the PR

PR #19624, opened 2026-06-10
sunswap-v2: widen API staleness window so daily-lagging data isn't dropped

tarun-khatri:
### Problem
`sunswap-v2` (TRON's main AMM) reports **\$0 TVL**, but the protocol is active — SunSwap's own liquidity API returns **~\$254M**, refreshed daily.

TVL for this protocol is produced by `api.js` (`httpTvl`). It fetches `getAllLiquidityVolume` (one snapshot per day) and only accepts the latest point if it is **less than 24h old**, otherwise it throws and falls back to on-chain. But that endpoint trails ~1 day, so the latest snapshot is consistently **24–48h old** (35h at the time of writing) → the 24h check always fails → it falls to the on-chain path, which currently returns \$0 (and takes ~196s) → \$0 is what gets stored.

### Fix
Widen the staleness window to 3 days, so the daily-lagging snapshot is accepted. The on-chain fallback still triggers if the source is genuinely frozen (> 3 days old).

### Verification
- `getAllLiquidityVolume` latest point is ~35h old; daily snapshots over the last 5 days are \$251M–\$256M.
- With the wider window, `node test.js projects/sunswap-v2/api.js` returns **\$254.07M** (previously it fell through to the on-chain path and returned \$0).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TVL retrieval by selecting the closest matching liquidity snapshot within a 24-hour publication-lag window, and falling back to the on-chain source when no suitable snapshot is found or data parsing fails—reducing unnecessary source switching.
* **Documentation**
  * Updated comments to clarify the daily snapshot cadence and the new data-freshness rules governing when the HTTP TVL source is used.

### DIFF
--- a/projects/sunswap-v2/api.js
+++ b/projects/sunswap-v2/api.js
@@ -3,36 +3,34 @@ const { getUniTVL } = require('../helper/unknownTokens.js')
 
 const onChainTvl = getUniTVL({ factory: 'TKWJdrQkqHisa1X8HUdHEfREvTzw4pMAaY', useDefaultCoreAssets: true, queryBatched: 11 })
 
-module.exports = {
-  timetravel: false,
-  misrepresentedTokens: true,
-  tron: {
-    // tvl: onChainTvl,
-    tvl: httpTvl,
-  }
-}
-
-
+const API_LAG_SECONDS = 24 * 3600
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR. Instead of widening the window, I've refactored so that the adapter only fetches data where the timestamp is within 24 hours of the input date

PR #19643, opened 2026-06-11
Add minimal balances health-check to test.js

### LINKED ISSUE #19601: Add warning balance health checks to the adapter test harness
## Description
I’ve been experimenting with a small check for adapter output and wanted to open this issue before turning it into a PR.

Rather than adding a new script, the idea is to integrate a couple of lightweight sanity checks into the existing adapter test flow in `test.js`.

The checks would look at the balances returned by adapters and warn when they contain:

- `NaN` values.
- Negative values.

This is meant to be a small first step toward better adapter validation. It would not change current behavior for valid adapters, and it would not fail the run yet, just surface warnings when something looks suspicious.

### Why this could be useful:

- It gives maintainers an early signal when an adapter returns clearly broken data.

- It fits naturally into the existing test harness, so there’s no extra CLI or separate tool to maintain.

- It keeps the tooling surface small while still improving visibility into adapter health.

Fils7:
Fixes:  #19601

### Description
This PR adds warning-only sanity checks to the existing adapter test flow in test.js. It inspects adapter balances for NaN and negative values, helping surface obviously broken output earlier without adding a new script or changing current behavior for valid adapters.

## Summary by CodeRabbit

* **Chores**
  * Enhanced TVL balance monitoring with automated validation checks that detect anomalies and emit diagnostic warnings.

### DIFF
--- a/test.js
+++ b/test.js
@@ -40,6 +40,23 @@ if (process.env.LLAMA_SANITIZE)
   })
 process.env.SKIP_RPC_CHECK = 'true'
 
+function analyzeBalances(balances) {
+  let hasNaN = false;
+  let hasNegative = false;
+
+  function visit(val) {
+    if (val == null) return;
+    if (typeof val === "number") {
+      if (Number.isNaN(val)) hasNaN = true;
+      if (val < 0) hasNegative = true;
+    } else if (typeof val === "object") {
+      for (const v of Object.values(val)) visit(v);
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR. Sorry, but closing for now as we have negative balance checks in the server: https://github.com/DefiLlama/defillama-server/blob/master/defi/src/storeTvlInterval/storeNewTvl2.ts

PR #19655, opened 2026-06-12
update tvl for berachain, hyperliquid on brownfi-v3

anhnq82:
This PR updates the Brownfi V3 tvl on berachain and hyperliquid network

## Summary by CodeRabbit

* **New Features**
  * Updated Brownfi protocol to V2 version with chain mappings preserved
  * Added Brownfi V3 protocol support for Berachain and Hyperliquid networks

### DIFF
--- a/registries/uniswapV2.js
+++ b/registries/uniswapV2.js
@@ -267,7 +267,7 @@ const uniV2Configs = {
   'BroSwap': {
     core: '0x8edF7B8411b2e5dB740dbbf949E011e59fc7980a',
   },
-  'brownfi': {
+  'brownfi-v2': {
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR, I've added v3 in #19678 to keep the original entry as `brownfi`. If you would like, we can update the page to show "BrownFi-v2" instead of "BrownFi"

anhnq82:
> Thank you for the PR, I've added v3 in #19678 to keep the original entry as `brownfi`. If you would like, we can update the page to show "BrownFi-v2" instead of "BrownFi"

Thank you so much @RohanNero 
Please help me separate the UI into 3 pages: BrownFi (a combination of V2 and V3), BrownFi V2, and BrownFi V3.

PR #19669, opened 2026-06-15
Fix hinkal tvl

delibaaa:
Adds Tron and Solana to the Hinkal TVL adapter (Solana funds are read from the program's `hinkal_vault` PDA, not the program id). Also folds in aggregated public balances, so the protocol's Total TVL reflects all the current funds instead of only shielded balances.


projects/hinkal/solanaTokens.js
```diff
@@ -0,0 +1,73 @@
+// SPL + Token-2022 mints Hinkal supports on Solana. Checking this fixed list lets
+// the vault's balances be read in one batched call, instead of scanning every
+// account it owns (heavy on public RPCs); native SOL is added via solOwners.
+module.exports = [
```
RohanNero, line 4: Could you please remove this token list? We don't use the public RPC in production so the `Error: Failed to post https://api.mainnet-beta.solana.com` message can be ignored

projects/hinkal/publicTvl.js
```diff
@@ -0,0 +1,34 @@
+const { get } = require("../helper/http");
+
+const PUBLIC_TVL_URL =
+  "https://wallet-staging-v2.hinkal.io/data-server/fetch-proxy-tvl";
```
RohanNero, line 4: Apologies for the inconvenience, but we are no longer accepting tvl reported by APIs for reasons described here: https://github.com/DefiLlama/DefiLlama-Adapters/discussions/432

Could you track this on chain instead?

RohanNero (changes requested):

RohanNero:
closing as stale

PR #19695, opened 2026-06-17
Monetrix: derive TVL from USDM supply for historical backfill

MonetrixLab:
Updates the already-merged Monetrix TVL adapter so its history backfills correctly.

The previous version read the Core backing via MonetrixAccountant.totalBackingSigned(), which uses Hyperliquid precompiles that only expose live L1 state and cannot be queried at historical blocks — so the TVL chart could only fill forward from the listing date.

USDM is minted/redeemed 1:1 against USDC, so circulating USDM supply equals the USDC collateral backing the token. Reading USDM.totalSupply() (plain ERC20 storage, archival-readable) lets DefiLlama backfill the full TVL history. Added `start: '2026-06-10'` (mainnet launch). The live value is unchanged (~0.06% vs the backing read). Genesis Vault USDC is still counted.

## Summary by CodeRabbit

* **TVL Calculation Updates**
  * Monetrix TVL calculation enhanced to include Genesis Vault USDC contributions, providing comprehensive view of total protocol collateral and liquidity
  * TVL now directly derives from USDM token total supply, enabling transparent on-chain verification of the 1:1 USDC collateral backing model
  * Updated protocol metadata and start date information to June 10, 2026

### DIFF
--- a/projects/monetrix/index.js
+++ b/projects/monetrix/index.js
@@ -1,7 +1,7 @@
 const ADDRESSES = require('../helper/coreAssets.json')
 
 const MONETRIX_GENESIS_VAULT = '0xc50A1dd2866A822c81bd0aA00B638c4BdDc9cd63'
-const MONETRIX_ACCOUNTANT = '0x8950A5136f3994f82b998e37e1183b8A37c12705'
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR, I will merge and backfill then revert to using the accountant contract

PR #19698, opened 2026-06-17
Add JustLend V2 (Tron) TVL adapter

BlackChar92:
## JustLend V2 (Tron)

Adds a TVL adapter for **JustLend V2**, an isolated-market lending protocol on Tron.

- **TVL** sums each market's supplied loan-token balance and posted collateral.
- **Borrowed** sums each market's outstanding debt.
- Balances are fetched from the JustLend V2 vault-list API and valued by their **real on-chain token addresses** — DefiLlama prices them itself (no CoinGecko-id mapping, no USD passthrough). Human-readable API amounts are converted to raw integer units via each market's decimals.

Local test (`node test.js projects/justlend-v2`) returns ~$700k mainnet TVL across the USDT / USDD / WTRX / sTRX / BTC markets.

JustLend V1 is already covered by the existing `projects/justlend.js`, so this PR adds only the V2 adapter.

### DIFF
--- a/projects/justlend-v2/config.js
+++ b/projects/justlend-v2/config.js
@@ -1,0 +1,13 @@
+// JustLend V2 isolated-market lending — vault-list API (TRON mainnet).
+// Amounts are returned human-readable; index.js converts them to raw units via each market's
+// decimals, then values them by their real on-chain token addresses.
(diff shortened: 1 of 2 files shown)


g1nt0ki (commented):
@BlackChar92 thanks for the PR

PR #19704, opened 2026-06-17
spl-governance (Realms): compute TVL on-chain instead of a frozen data feed

tarun-khatri:
The Realms / `spl-governance` adapter pulled TVL from `realms-tvl.vercel.app`, which has served a frozen snapshot since 2024-11-15 (its prices are stuck at Nov-2024 levels, e.g. SOL ~$210), so the reported TVL and staking have been flat ever since. The source it used before that, `api.realms.today`, no longer resolves.

This replaces the dead feed with an on-chain computation:
- enumerate the governance accounts across the SPL Governance program instances (`getProgramAccounts`),
- derive each governance's native treasury PDA (`['native-treasury', governance]`),
- sum the treasuries' SOL + SPL holdings from Allium's `solana.assets.balances_daily` (the same balances source the `pump-swap` adapter uses).

SOL, stables, BTC and ETH are counted under TVL; governance tokens under staking — the same split the adapter used before.

Verified the treasury balances against on-chain state, e.g. the Jito DAO treasury currently holds ~209M JTO, which the adapter picks up. The protocol itself is active (governance transactions land continuously); only the old data feed was stale.

A few legacy program instances have an account index too large for a plain `getProgramAccounts` (the RPC asks for paginated `getProgramAccountsV2`); they hold negligible treasury value and are skipped with a logged reason.

## Summary by CodeRabbit

* **New Features**
  * Separated Total Value Locked (TVL) and staking into independent functions for more granular reporting.
* **Refactor**
  * Updated TVL and staking calculations to use on-chain SPL Governance treasury enumeration and balance aggregation.
  * Refreshed the reported methodology description to match the new on-chain-backed approach.

### DIFF
--- a/projects/spl-governance/index.js
+++ b/projects/spl-governance/index.js
@@ -2,7 +2,7 @@ const { PublicKey } = require('@solana/web3.js')
 let bs58 = require('bs58'); if (bs58.default) bs58 = bs58.default
 const { getConnection } = require('../helper/solana')
 const { queryAllium } = require('../helper/allium')
-const { sliceIntoChunks } = require('../helper/utils')
+const { sliceIntoChunks, sleep } = require('../helper/utils')
 
 // SPL Governance program instances that hold Realms DAO treasuries.
 const GOV_PROGRAMS = [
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR

PR #19708, opened 2026-06-18
Update AFX methodology

afx-exchange:
## Release Notes

* **Documentation**
  * Enhanced methodology documentation with detailed explanations of USDC deposit and AFX fee accounting, including clarification of exclusions and incentive program allocations.


RohanNero:
Thanks for the PR. Closing since we use the methodology string to describe how TVL is calculated, but I will update the AFX description to include the added text. Also, if you'd like to track fees/volume, you can open a PR here: https://github.com/DefiLlama/dimension-adapters

PR #19728, opened 2026-06-19
Rockawayx Removing old midas vaults

Sithey:
Rockawayx Removing old midas vaults

## Summary by CodeRabbit

* **Updates**
  * Updated MIDAS TVL aggregation to reflect the current set of supported Ethereum vaults.
  * MIDAS Ethereum now only includes the remaining `mROX` vault for TVL metrics, removing the previously included additional Ethereum vault addresses.
  * TVL data for EMBER, LISTA, and ACCOUNTABLE vault groups remains unchanged.


RohanNero (changes requested):
Thank you for the PRs. On the Midas frontend Rockawayx is still the curator, despite being deprecated, we'd prefer to continue tracking them.

However, we can remove them if midasTvl() is adding stale balances that are no longer accurate. If this is the case, could you export plume_mainnet and etlk as `() => ({})` please

Sithey:
I will cancel the PR for the moment, will remake it when vaults will be removed from midas

PR #19750, opened 2026-06-23
beraborrow: fix no-op blacklist

DaniiRix:
Fixes blacklist creation in `projects/beraborrow/index.js`.

Previously:

```js
const beraTokens = bbInfraWrappers.push(beraBorrowWberaToken)
return api.sumTokens({ tokensAndOwners, blacklistedTokens: beraTokens })
```

### Fix

```js
const blacklistedTokens = [...bbInfraWrappers, beraBorrowWberaToken]
return api.sumTokens({ tokensAndOwners, blacklistedTokens })
```

This correctly passes the intended blacklist and avoids mutating `bbInfraWrappers`.

### Impact

No TVL change observed ($317.56k before/after).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed token blacklisting logic in TVL calculations to ensure the correct set of blacklisted tokens are properly excluded from the total value locked computation.


RohanNero:
Thanks for the PR

PR #19779, opened 2026-06-24
bouncebit: replace subgraph tvl with onchain reads

DaniiRix:
This PR removes subgraph-based TVL calculations and replaces them with block-accurate on-chain reads for both V2 and V3.

The previous implementation queried subgraphs without block filters, which broke historical (`timetravel`) queries and led to several inaccuracies:

* Ethereum V3 could report negative USDT TVL due to event-based accounting.
* The configured BSC V3 subgraph points to a retired deployment.
* The BounceBit `boyya` subgraph reported ~76M of phantom stBBTC TVL that does not exist on-chain.

### V3 (`cedefiV3.js`)

* Replaced subgraph queries with direct reads from the live V3 gateway.
* Fetches supported tokens via `getAllSupportedTokens()`.
* Uses `api.sumTokens()` to measure actual assets held by the gateway.
* Handles native token balances correctly via `ADDRESSES.null`.
* Removed BSC V3 support since there is no active V3 gateway.

### V2 (`index.js`)

* Replaced subgraph TVL reconstruction with on-chain vault reads.
* For each strategy:

  * `getStrategyIds()`
  * `vaultRegistries()`
  * `getTotalInfo()`
* TVL is calculated as `principal + rewards`.

Covers all Ethereum, BSC, and BounceBit V2 gateways.

## Verification

* Verified V2 and V3 ABIs against deployed contracts.
* Confirmed historical block accuracy by comparing outputs across different Ethereum blocks.
* Investigated the BounceBit `boyya` discrepancy and confirmed the reported ~76M stBBTC TVL was phantom subgraph state rather than real on-chain assets.

## Summary by CodeRabbit

* **New Features**
  * Updated CeDeFi TVL to read positions on-chain (Gateway/Vault strategy aggregation) for supported networks, improving accuracy.
  * CeDeFi V2/V3 TVL is now combined with existing EasyBTC, Premium, Promo, and Solana vault staking sources.

* **Bug Fixes**
  * TVL aggregation now includes both principal and rewards when summing vault/strategy balances.
  * Added a fallback to standard balance retrieval for unsupported chains.

### DIFF
--- a/projects/bouncebit-cedefi/index.js
+++ b/projects/bouncebit-cedefi/index.js
@@ -1,38 +1,37 @@
-const { cachedGraphQuery } = require('../helper/cache')
 const { sumTokens, sumTokensExport } = require('../helper/sumTokens');
-const ADDRESSES = require('../helper/coreAssets.json')
-const { getProvider, } = require('../helper/solana')
+const ADDRESSES = require('../helper/coreAssets.json');
+const { getProvider, } = require('../helper/solana');
 const { Program } = require("@project-serum/anchor");
-const cedefiFromSolana = require('./cedefiFromSolana')
-const cedefiV3 = require('./cedefiV3')
+const cedefiFromSolana = require('./cedefiFromSolana');
+const cedefiV3 = require('./cedefiV3');
 
 // ===== inlined from ./easyBTC =====
 const easyBTC = (() => {
-  const ETH_EasyBTC_USD_Contract = '0xFe32277D00e57D864B8BC687d0a442D663Aa1dF6'
+  const ETH_EasyBTC_USD_Contract = '0xFe32277D00e57D864B8BC687d0a442D663Aa1dF6';
 
-  const BNB_EasyBTC_USD_Contract = '0xF3FB36F32Dad91627f688e7332472d69F6C985c6'
-  const BNB_EasyBTC_BTC_Contract = '0x38D239a8D33BF7424A1Df6d39cb8523cCc25DE0e'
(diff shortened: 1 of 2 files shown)


projects/bouncebit-cedefi/index.js
```diff
@@ -224,117 +223,138 @@ const promoFromSolana = (() => {
...
+      '0x8d0d000ee44948fc98c9b98a4fa4921476f08b0d', // USD1
+      NATIVE,
+    ],
   },
   bouncebit: {
-    main: { url: 'https://bitswap-subgraph.bouncebit.io/subgraphs/name/bb-defillama-bb' },
-    boyya: { url: 'https://bitswap-subgraph.bouncebit.io/subgraphs/name/bb-defillama-boyya-bb' }
+    gateways: [
+      '0xc4F65Bbdd0B9eCFeaA253a65DC0601C97061a02C', // main
+      '0x4d4Bd4565d3D39940886b1311f54fd3A766013be', // "boyya"
```
RohanNero, line 282: It looks like the missing ~$76M is sitting in 0x7F150c293c97172C75983BD8ac084c187107eA19 as stBBTC, should this be included to match the tvl reported by the subgraphs?

RohanNero (changes requested):

projects/bouncebit-cedefi/index.js
DaniiRix: the address seems like an EOA and not like protocol locked funds

RohanNero (approved):

RohanNero:
Apologies for the delay and thank you for the PR

PR #19790, opened 2026-06-25
fix: pulsex-v2 adapter to use onchain data

DaniiRix:
This PR fixes the PulseX V2 TVL calculation by replacing a stale subgraph value with on-chain liquidity data. This restores accurate historical TVL (timetravel) support.

## What was wrong?

The adapter was reading a single subgraph field (`totalLiquidityUSD`), which always returned the latest TVL regardless of the requested timestamp.

This caused:
- TVL to remain around **$15M** across all dates
- Historical queries to return current values
- TVL to be derived from subgraph data instead of on-chain reserves

## Results

| Date | Pair Count | TVL |
|--------|----------:|--------:|
| Current | 181,356 | **$19.19M** |
| 2024-06-23 | 77,372 | **$26.92M** |

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated PulseX V2 TVL tracking to use on-chain pool enumeration, improving the accuracy and reliability of displayed liquidity data.
  * Core asset reserves are now counted directly in the valuation, helping keep TVL totals aligned with live protocol state.

* **Documentation**
  * Added clearer methodology information for how PulseX V2 TVL is calculated.

### DIFF
--- a/projects/pulsex-v2/index.js
+++ b/projects/pulsex-v2/index.js
@@ -1,19 +1,14 @@
-const { graphQuery } = require('../helper/http')
+const { getUniTVL } = require('../helper/unknownTokens');
 
-const SUBGRAPH = 'https://graph.pulsechain.com/subgraphs/name/pulsechain/pulsexv2'
-
-async function tvl(api) {
-  const { pulseXFactories } = await graphQuery(SUBGRAPH, `{
-    pulseXFactories {
-      totalLiquidityUSD
-    }
-  }`)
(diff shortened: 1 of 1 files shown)


RohanNero:
Sorry, usually we would prefer to read balances onchain like this, but due to the number of pairs, the adapter was breaking so we updated it to use the subgraph recently: https://github.com/DefiLlama/DefiLlama-Adapters/commit/d484762937c6ebd771b18cebe26a0b761eaa4c90

Closing for now, but in the future we may revert to using the factory and uni helper

PR #19791, opened 2026-06-25
fix(increment-lending): replace web API with on-chain Cadence reads

DaniiRix:
This PR replaces Increment Lending’s previous API-based data source with an on-chain Flow query.

To remove that dependency, the adapter now reads protocol data directly from Flow using a Cadence script executed through the existing `callCadence` helper.

## Implementation Details

The adapter executes a single Cadence view script that gathers both TVL and borrow data in one request.

The script:

- Borrows the `LendingComptroller` capability at `0xf80cb737bfe7c792`.
- Iterates through every market returned by `getAllMarkets()`.
- For each market, borrows its `PoolPublic` capability and retrieves:
  - `getPoolCash()` — assets currently held in the pool vault.
  - `getPoolTotalBorrowsScaled()` — total outstanding borrows.
- Reads market prices from the `LendingOracle` at `0x72d3a05910b6ffa3`.
- Converts scaled `UInt256` values into `UFix64` using `LendingConfig.ScaledUInt256ToUFix64`.
- Multiplies balances by their corresponding USD prices and accumulates:
  - `totalCashUSD`
  - `totalBorrowUSD`
- Automatically skips markets that do not have an oracle price available (for example, deprecated markets such as USDC).

## Test Output

| Metric | Token | Value |
|----------|----------|----------|
| flow | USDT | 407.31k |
| flow-borrowed | USDT | 46.94k |
| borrowed | USDT | 46.94k |
| total | — | 407.31k |

## Summary by CodeRabbit

* **Bug Fixes**
  * Switched lending TVL and borrow calculations to use on-chain data, improving accuracy and reducing reliance on external snapshots.
  * Added a shared cache so repeated balance and borrow lookups reuse the same result and recover cleanly if a fetch fails.

* **New Features**
  * Exposed additional USD-based metrics for cash and borrows in the lending dashboard.
  * Marked the project as not supporting timetravel queries.

### DIFF
--- a/projects/increment-lending/index.js
+++ b/projects/increment-lending/index.js
@@ -1,17 +1,67 @@
-const { fetchURL } = require("../helper/utils");
+const { callCadence } = require("../helper/chain/flow");
 
-// lending info: https://app.increment.fi/markets
-async function borrowed(api) {
-  const { data: tvls } = await fetchURL("https://app.increment.fi/info/tvl");
-  return api.addUSDValue(Math.round(tvls.LendingBorrow))
+// Queries each lending pool's cash + borrows via on-chain Cadence script.
+const script = `
+import LendingInterfaces from 0x2df970b6cdee5735
+import LendingConfig from 0x2df970b6cdee5735
+
+access(all) view fun main(): [UFix64] {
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR

PR #19800, opened 2026-06-25
fix(baton): price fractional nft farms and correct filter assignment

tarun-khatri:
This PR fixes the zero TVL bug for Baton.

- Fixes a type strict equality bug (`i.farmType === 2`) on getLogs decoded arguments. Since `farmType` is a uint8, the event decoder returned it as a `BigInt` (e.g., `2n`). This failed the strict comparison `2n === 2`, which filtered out all farms and resulted in 0 TVL. We resolved this by wrapping with `Number(i.farmType || i[6]) === 2` for a robust, type-safe comparison.
- Fixes an assignment bug `baseTokens[idx] = nullAddress` which erroneously assigned the `nullAddress` instead of evaluating with `=== nullAddress`.
- Verified locally via `node test.js projects/baton/index.js` which successfully compiles and recovers ~$8.14k TVL in ETH instead of returning 0.00.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved farm filtering so results work correctly across different data shapes.
  * Fixed an issue that could cause incorrect filtering due to a mistaken comparison.

### DIFF
--- a/projects/baton/index.js
+++ b/projects/baton/index.js
@@ -16,9 +16,9 @@ async function tvl(api) {
   });
 
   // filter any farms where the reward token is not fractional nfts or the underlying pair is not paired with eth
-  let filteredLogs = logs.filter(i => i.farmType === 2)
+  let filteredLogs = logs.filter(i => Number(i.farmType || i[6]) === 2)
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PRs

PR #19817, opened 2026-06-27
fix(predmart): remove inactive Base entry (Polygon-only)

sev-kach:
PredMart operates only on Polygon. The `base` export returns no TVL (`tvl: () => ({})`) and only causes Base to appear as a supported chain on the protocol page. This removes it; the Polygon `tvl` and `borrowed` logic is unchanged.

- TVL: liquid USDC in the lending pool (Polygon)
- Borrowed: `totalBorrowAssets()` on the pool

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved PredMart TVL reporting by removing an empty chain entry and keeping only the supported network data.
  - Liquid USDC balances and borrowed USDC are now reflected more accurately for the lending pool.


RohanNero:
Sorry, but we prefer to keep deprecated chains in adapters to preserve historical data: https://github.com/DefiLlama/DefiLlama-Adapters/discussions/13020

PR #19825, opened 2026-06-28
Update Cookie Chain TVL to price both sides of liquidity pools

fibanachos:
## Problem

On Cookie Chain, only **COOK** (the native gas token, mapped to coingecko `cookie-2`) has a price feed. Every other token on the chain has no price source, so DeFiLlama silently drops it from TVL — leaving each liquidity pool counted at **only its COOK side**.

## Solution

New shared helper `projects/helper/cookiechain.js` that derives prices for the chain's tokens directly from the on-chain DEX pools and re-expresses them as their COOK-equivalent (the SVM analog of the EVM-only `helper/cache/sumUnknownTokens`).

It reads every pool over RPC and:

1. **Phase A — direct COOK pricing (authoritative):** each token is priced from the COOK pool holding the most COOK; these prices are never overridden.
2. **Phase B — indirect pricing:** a token with no direct COOK pool is priced one hop further through an already-priced neighbour, but only along the deepest-backed path and only when the hop clears a minimum COOK-liquidity floor. (Unbounded transitive pricing was tested and rejected — it corrupts good prices and blows up through thin pools.)
3. **Per-token cap:** a token's total contribution is capped at `RESTRICT_TOKEN_RATIO` (100×) the COOK liquidity backing its price, so a thin or manipulated pool cannot inflate TVL.
4. **Bonding curves (DBC):** counted by their **quote (deposited COOK) side only** — the base side is unsold mint supply, not real locked value.

AMM pools (DAMM, CLMM, BAMM, CPAMM) contribute both sides.

## Implementation notes

- Pure on-chain reads via the configured Cookie Chain RPC — `getProgramAccounts` (pool discovery, with `dataSize` filter + `dataSlice`) and `getMultipleAccountsInfo` (vault mint + balance, sliced to 72 bytes). No external/price API is called inside the adapter.
- All pool account layouts (vault offsets, discriminators) and the `sqrt_price → price` direction were validated byte-for-byte against live accounts on `rpc.cookiescan.io`.
- COOK is summed exactly (`BigInt`); derived values are float-rounded (≪0.01% error).
- Both adapters set `misrepresentedTokens: true`.

## Summary by CodeRabbit

* **New Features**
  * Introduced a shared Cookie Chain pool valuation helper and enabled Cookie Chain + CookieSwap TVL tracking through it for consistent, COOK-equivalent pool valuation.
  * Added pricing support that can derive token value via multi-hop routes when direct COOK pricing is unavailable.
* **Bug Fixes**
  * Improved TVL methodology and token handling for launch/bonding-curve style cases, with safeguards to reduce overcounting.
  * Updated Cookie Chain/CookieSwap metadata (including methodology text and token representation rules).

### DIFF
--- a/projects/cookiechain/index.js
+++ b/projects/cookiechain/index.js
@@ -1,66 +1,12 @@
-const { PublicKey } = require('@solana/web3.js')
-const { getConnection, getTokenAccountBalances } = require('../helper/solana')
-
-const TOKEN_PROGRAMS = [
-  new PublicKey('TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA'),
-  new PublicKey('TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb'),
-]
-
-// CookieBox DAMM (Meteora Dynamic AMM v2 / cp_amm fork) and CookieBox DBC (Meteora
-// Dynamic Bonding Curve fork) both hold every pool's token vaults under a single
-// `pool_authority` PDA, so we enumerate the vault token accounts owned by that PDA.
-async function sumPoolAuthorityVaults(connection, api) {
-  const programs = [
-    'DAMMjDCEFTDkt7ywazZS8GoaLtjb3HaJo3pLbf64xrPY', // CookieBox DAMM
-    'DBCg4ugDEztk6MbqHEJvx5a5YGJTj45Jb5NvtQ48Rvsf', // CookieBox DBC
-  ]
-  for (const programId of programs) {
-    const [auth] = PublicKey.findProgramAddressSync(
-      [Buffer.from('pool_authority')],
-      new PublicKey(programId),
-    )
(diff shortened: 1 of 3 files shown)


RohanNero:
Apologies for the delay, I am currently adding an adapter on the server for cookiechain token prices so that these adapters can remain unchanged

fibanachos:
Hi @RohanNero, thanks for the update! Just wondering if you have an ETA for when the adapter might be ready?

RohanNero:
@fibanachos Hey, I will have it finished today. Once it is live I will close this PR with a note

fibanachos:
Thank you very much @RohanNero!

RohanNero:
I've added a cookiechain pricing adapter using the on-chain pool liquidity, but since we use a minimum liquidity threshold, no tokens are currently being priced. Apologies for the inconvenience, but once there is at least $10k worth of COOK in a pool, the token paired with COOK will be priced

PR #19826, opened 2026-06-28
Add Neverland AUSD/PT-AUSD Isolated Pool

0xCatalyst:
Adds the `AaveProtocolDataProvider` for the new isolated AUSD/PT-AUSD market alongside the canonical (cross) one.
`aaveV3Export` already aggregates multiple providers, so TVL/Borrowed pick it up with no new code.

There is no need of any special handling. The market is isolated in name only; its behavior is identical to any Aave V3 pool, and the isolation derives purely from the operational listing of correlated assets only, here a Pendle strategy with the AUSD and it's PTs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the Monad market configuration to use multiple pool data sources, improving the accuracy of reported Aave V3 market data.
  * Adjusted the calculated TVL and borrowed values for Monad to reflect the expanded set of markets.

### DIFF
--- a/projects/neverland/index.js
+++ b/projects/neverland/index.js
@@ -1,19 +1,23 @@
 const { aaveV3Export } = require("../helper/aave");
 const { sumTokens2 } = require("../helper/unwrapLPs");
 
-// Neverland Money - Aave V3-based lending protocol on Monad with vote-escrow tokenomics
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR

PR #19831, opened 2026-06-29
special handling for coins that are not priced in defillama yet

Zorag44:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved how certain vault shares are counted in market views, so totals now reflect the underlying asset instead of the wrapper token when applicable.
  * Updated TVL and borrowed figures for affected markets to show more accurate values across supported assets.


RohanNero:
Thank you for flagging this, I've fixed the ember price adapter on the server: https://coins.llama.fi/prices/current/sui:0x89b0d4407f17cc1b1294464f28e176e29816a40612f7a553313ea0a797a5f803::ethird::ETHIRD

PR #19844, opened 2026-06-29
Update P0 adapter to new API endpoint

princedotrs:
# Update P0 adapter to new API endpoint

**Protocol:** Project 0 (`projects/p0`)
**Base PR target:** `main`

## What changed
Repoints the P0 TVL adapter from the legacy `base-api-public.mrgn.app/v1/bank/metrics`
service to Project 0's `https://api.0.xyz/v0/bankMetrics` endpoint.

## Methodology
- **TVL = sum of all deposited funds** across every P0 bank (deposit amount × oracle price).
  *(Previously the adapter reported available liquidity = deposits − borrows; it now reports
  gross deposits, which is the figure Project 0 wants surfaced as TVL.)*
- **Active Loans (`borrowed`) = sum of all outstanding borrows.**
- Standard banks are added per underlying mint (`api.add`) so DefiLlama prices them and can
  track inflows.
- Kamino / Drift / JupLend **integration banks** hold receipt-token mints that DefiLlama can't
  price independently, so they're reported as the oracle-computed USD the API already returns
  (`api.addUSDValue`).
- `doublecounted: true` retained (borrowed funds originate as deposits, and integration-bank
  underlyings are also counted by the venue they're deposited into).

The host can be overridden with the `P0_PUBLIC_API_BASE` env var if our hostname ever changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated TVL and borrowed calculations to use the latest public **v0** bank metrics feed (current state only).
  * Removed the prior time-range aggregation and deposits-minus-borrows TVL logic in favor of native totals reporting.
* **New Features**
  * Added support for overriding the public API base URL via configuration.
* **Documentation**
  * Refreshed the exported methodology text to describe the new v0 source and “available/active” behavior based on current-state data.

### DIFF
--- a/projects/p0/index.js
+++ b/projects/p0/index.js
@@ -1,108 +1,55 @@
 const axios = require('axios');
-const { getEnv } = require('../helper/env');
 
-const BASE_URL = 'https://base-api-public.mrgn.app';
-const API_ENDPOINT = `${BASE_URL}/v1/bank/metrics`;
-const startTime = new Date().toISOString();
-const END_TIME = '9999-12-31T23:59:59.999Z';
-const GROUP_ADDRESS = '4qp6Fx6tnZkY5Wropq9wUYgtFxXKwE6viZxFHg3rdAG8';
+// P0 public API (p0-monitor). Override the host via P0_PUBLIC_API_BASE if needed.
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR, all integration bank mints appeared to be priced so I dropped the oracle from addSide

PR #19852, opened 2026-06-30
Fix hinkal total tvl

delibaaa:
Adds Tron and Solana to the Hinkal TVL adapter and folds the protocol's largest public proxy wallets into each chain's TVL, so the reported total reflects both shielded-pool balances and live public funds. Public balances are now read entirely on-chain from a fixed set of top proxy addresses (owners.js) instead of the previous API endpoint, per the maintainer's request in #19669. Solana balances use solOwners auto-discovery, removing the hardcoded mint list that triggered the public-RPC error. Tron support reads the shielded vault's USDT/USDC plus native TRX on-chain.

## Summary by CodeRabbit

* **New Features**
  * Added support for tracking assets on two additional chains.
  * Expanded balance coverage to include more token holdings and native balances in the total value calculation.

* **Bug Fixes**
  * Improved balance detection by calculating holdings directly and tolerating partial query failures.
  * Updated token tracking so underlying assets are counted correctly when available.


RohanNero:
Thank you for the PR. Apologies for the inconvenience, but I've refactored slightly and removed the new owners addresses as we are no longer accepting tvl in EOAs or safe proxies

PR #19873, opened 2026-07-02
fix debank helper

RohanNero:
update to fail loudly when out of credits

## Summary by CodeRabbit

* **Bug Fixes**
  * Requests now fail fast when the required Debank API key is missing, instead of returning empty results.
  * When a Debank API call fails, the app now surfaces a clearer error with endpoint and response details instead of silently falling back to empty data.


albertoblue87-netizen (commented):
Listo

PR #19880, opened 2026-07-02
fix(t3tris-finance): read grossTVL so TVL is not reported as 0

0xshubhs:
## Summary

T3tris Finance reports **$0 TVL on DefiLlama** ([live API](https://api.llama.fi/protocol/t3tris-finance)) despite its vaults holding real assets. Two problems:

**1. Field-name casing typo.** `getGrossTVL()` returns a field named **`grossTVL`**:
```js
"function getGrossTVL() external view returns (uint256 totalManagedAssets, uint256 pendingDeposits, uint256 claimableRedeems, uint256 grossTVL)"
```
but the code read **`g.grossTvl`** (lowercase `v`) → always `undefined` → `api.add(token, undefined)` counted nothing.

**2. Duplicate adapter for the same slug.** Both `projects/t3tris-finance.js` and `projects/t3tris-finance/index.js` existed (the flat file even worked via a `g[3]` positional fallback, but the directory version, the one effectively loaded, did not). This dual-file setup is fragile. Removed the redundant flat file and kept the directory adapter as the single source of truth.

## Testing

```
node test.js projects/t3tris-finance/index.js
  before -> arbitrum $0
  after  -> arbitrum ~$6.0M   (USDC)
```

No API keys required (public vaults API + public Arbitrum RPC). No methodology change; this makes the adapter read the value its own `methodology` already describes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected a TVL calculation field so reported totals now match the returned vault data more reliably.

* **Chores**
  * Removed an outdated T3tris Finance vault TVL integration from the app.


RohanNero:
Thank you for the PR

PR #19900, opened 2026-07-04
fix(k3): remove double-counted Euler owner on monad

mkzung:
## Problem
k3 double-counts the monad Euler vault owner `0x5D42F8aCd567810D57D60f90bB9C6d194207a6e1`. It is summed in **both** the main `configs` (no `start`) **and** `eulerSunsetConfigs` (`start: "2026-05-06"`), which are combined with `mergeExports` - and `mergeExports` is additive (`addToExports` pushes each tvl fn into an array and sums them). So:

- post-sunset (now) the owner is counted **twice**, inflating k3 TVL by **~$62.74M**; and
- pre-sunset it is wrongly attributed to K3 instead of euler-dao.

Verified on-chain (`node test.js`, current timestamp):
- owner via a plain config: **$62.74M**
- owner via the sunset config (`start: "2026-05-06"`): **$62.74M** (so `start` still counts it now)
- `mergeExports([both])`: **$125.48M = exactly 2x** -> the double-count

## Fix
Remove the duplicated `eulerVaultOwners` line from `configs.blockchains.monad` (keep its `erc4626` vaults). `eulerSunsetConfigs` already attributes this owner correctly from the 2026-05-06 sunset, so current TVL is preserved (counted once).

Closes #19894.

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed an outdated owner entry from one configuration, keeping the setup aligned with the current state.

PR #19904, opened 2026-07-05
fix(k3): stop double-counting monad Euler vault owner

### LINKED ISSUE #19894: k3: monad Euler owner double-counted via eulerSunsetConfigs overlap
`projects/k3/index.js` sums `configs.blockchains.monad.eulerVaultOwners` and `eulerSunsetConfigs.blockchains.monad.eulerVaultOwners` independently via `mergeExports`, both listing the same owner (`0x5D42F8aCd567810D57D60f90bB9C6d194207a6e1`).

`eulerSunsetConfigs` has `start: "2026-05-06"` but there's no corresponding cutoff on the `configs` side, so from that date onward both configs are active simultaneously and their TVL gets summed together (via `sdk.util.sumChainTvls` in `mergeExports`, which sums multiple TVL functions registered under the same chain+key).

Verified with `test.js`: removing `eulerSunsetConfigs` drops monad TVL from $195.50M to $133.51M, a difference of approximately $62M currently being double-counted.

Relevant code: https://github.com/DefiLlama/DefiLlama-Adapters/blob/main/projects/k3/index.js

Likely fix: remove the owner from `configs.blockchains.monad.eulerVaultOwners` (if fully migrated to K3), or add an end-boundary so the two configs don't overlap in time.

0xshubhs:
Fixes #19894.

The monad Euler vault owner `0x5D42...a6e1` was listed in both the always-on `configs.blockchains.monad` and `eulerSunsetConfigs` (which counts it from the 2026-05-06 EulerDAO sunset). `mergeExports` sums both, so from the sunset date on that owner's TVL was counted twice.

The sunset config is the right place for it, it models the euler-dao → K3 handover with the correct start date, so I dropped the duplicate `eulerVaultOwners` entry from the always-on monad config and left the `erc4626` vaults there untouched.

Monad goes from 203.86M → 139.13M (total 353.30M → 288.57M); every other chain is unchanged. The ~64.7M removed is the double-counted owner.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected Monad TVL reporting to avoid double-counting a vault owner that is tracked separately.
  * Improved curation data handling so totals are more accurate and consistent.

### DIFF
--- a/projects/k3/index.js
+++ b/projects/k3/index.js
@@ -77,7 +77,9 @@ const configs = {
       eulerVaultOwners: ["0xAeE4e2E8024C1B58f4686d1CB1646a6d5755F05C"],
     },
     monad: {
-      eulerVaultOwners: ["0x5D42F8aCd567810D57D60f90bB9C6d194207a6e1"],
+      // The Euler vault owner 0x5D42...a6e1 was handed to K3 in the EulerDAO sunset
+      // and is counted from that date via eulerSunsetConfigs below; listing it here as
+      // well summed the same owner twice, double-counting monad TVL. Keep it only there.
       erc4626: [
         "0x1E4D67c666c2Ccf27A0aF980fE6c8e0f05aC8949",
         "0x502e4a0B61dEBA3015Eb9E51116B832003B22c2C",


RohanNero:
Thank you for the PR, closing in favor of #19900

Issue #19689, opened 2026-06-16
Incorrect Felix TVL reporting on HyperEVM


0x0Glitch:
<img width="2048" height="1174" alt="Image" src="https://github.com/user-attachments/assets/070065e9-bbc8-4181-ac52-6469e65e370e" />

<img width="2048" height="409" alt="Image" src="https://github.com/user-attachments/assets/ed9b10bb-2b38-4b02-aca5-e58806b6e60f" />

Hey @RohanNero @g1nt0ki, I was checking the Felix page on DefiLlama(https://defillama.com/protocol/felix), and I found an issue with the TVL shown for Felix(https://www.usefelix.xyz/vanilla/stats).

DefiLlama is showing Felix Combined TVL around $137M, but the Felix Vanilla alone has much higher TVL. In the attached screenshot, Felix Vanilla shows around $336.22M in Total Collateral Deposited, around $91.47M in Total Borrowed, and around $113.19M in Lending Vault Deposits. Using the Felix Vanilla frontend numbers, TVL should be calculated as Total Collateral Deposited + Lending Vault Deposits - Total Borrowed, which gives around $357.94M for Felix Vanilla alone.

I wanted to know the reasoning behind this difference between the Felix frontend numbers and the Felix Combined TVL shown on DefiLlama. Could you please clarify whether Felix Vanilla deposits are treated differently in DefiLlama’s TVL methodology, or if there is another reason why they are not fully reflected in the combined Felix TVL?

0xshubhs:
Dug into this, there are two separate parts.

Two Felix Metamorpho vaults (Felix USDC and Felix USDH "Frontier") were missing from the vault list, so Felix Vaults was under-reporting by ~$38M. Fixed in #19905, both verified same `owner()` as the already-listed vaults, so felix-vaults goes 63.8M → 102.4M.

The bigger gap vs the frontend is a different thing: the `felix-vanilla` adapter (~$207M of direct Morpho lending markets) doesn't seem to be listed as a protocol on the server side, nothing appears to consume that module (`/tvl/felix-vanilla` 404s). So that piece needs a listing on your end rather than an adapter change here. @g1nt0ki @RohanNero

RohanNero:
Thank you for flagging this, we removed felix-vanilla since it was tracking the same assets as felix-vaults. The value reported on Felix's stats page may include non-felix collateral but we will look into this to confirm.

PR #19905, opened 2026-07-05
fix(felix): add two missing Felix Metamorpho vaults

0xshubhs:
Refs #19689.

Two Felix Metamorpho vaults on HyperEVM were missing from the hardcoded list, so Felix Vaults TVL was under-reported:

- `0x808F72b6Ff632fba005C88b49C2a76AB01CAB545`, Felix USDC (Frontier), ~$38.5M
- `0x274f854b2042DB1aA4d6C6E45af73588BEd4Fc9D`, Felix USDH (Frontier), ~$60k

Both are genuine Felix vaults: `owner()` on each is `0x4A827418D632C415E19825fd011283A4ba020B3A`, the same owner as the already-listed vaults, and their names read "Felix USDC (Frontier)" / "Felix USDH (Frontier)". I added them to `felix-vaults` and to the subtraction list in `felix-vanilla` so the two siblings stay consistent and don't double-count.

felix-vaults: 63.81M → 102.39M. felix-vanilla still runs clean (206.52M).

Note: the larger gap vs the frontend in #19689 is separate, the `felix-vanilla` adapter (~$207M of direct Morpho lending markets) doesn't appear to be listed as a protocol on the server side, so that piece is a listing question for your team rather than an adapter change here.

## Summary by CodeRabbit

* **New Features**
  * Added support for two additional Felix Frontier vaults: USDC and USDH.
  * TVL calculations now include these newly tracked vaults, improving coverage of Felix vault balances.
* **Bug Fixes**
  * Updated the vault list so displayed totals reflect the current set of supported Felix vaults.


RohanNero:
Thank you for the PR

PR #19909, opened 2026-07-06
Add BitFi bfBTC TVL on Pharos

PeratX:
## Summary

This PR fixes BitFi CeDeFi bfBTC TVL pricing and add `Pharos bfBTC`.

Previously the adapter added each chain's bfBTC token address directly. Some chains had no DefiLlama price key (`btr`, `core`, `pharos`), so their TVL showed as `0`. It also priced bfBTC 1:1 with BTC on chains where a price key existed.

The adapter now:
- reads bfBTC `totalSupply()` on each supported chain
- reads `currentRatio()` from the Ethereum bfBTC contract
- converts bfBTC supply into BTC-equivalent amount
- prices the result directly with `coingecko:bitcoin`

## Methodology

TVL is calculated as the BTC-equivalent value of all bfBTC supply across supported chains.

Formula:

```text
btcAmount = bfbtcTotalSupply * 1e8 / currentRatio / 1e8
```

## Summary by CodeRabbit

* **New Features**
  * Added bfBTC support for the Pharos chain so holdings can be included in TVL reporting.
* **Chores**
  * Tidied minor comment formatting near the end of the source file.

### DIFF
--- a/projects/bitfi-cedefi/index.js
+++ b/projects/bitfi-cedefi/index.js
@@ -16,6 +16,9 @@ const mapping = {
   },
   core: {
     bfBTC: '0xCdFb58c8C859Cb3F62ebe9Cf2767F9e036C7fb15',
+  },
+  pharos: {
+    bfBTC: '0x623F2774d9f27B59bc6b954544487532CE79d9DF',
   }
 }
 const exportObject = {}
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR, as you said, all chains are 1:1 with btc currently but I will update their prices on the server to reference `currentRatio`

PeratX:
Thanks, please let me know when the prices are updated.
bfBTC price should be `BTC price / bfBTC.currentRatio()`

PR #19910, opened 2026-07-06
test.js: add NaN and negative balance health warnings

Ramprasad4121:
Implements #19601

Adds `checkBalanceHealth()` to the test harness. Called in `getTvl()` after raw balances are resolved, before `computeTVL()`. Warns via `console.warn` when any token balance is NaN or negative, so developers catch broken adapters during local testing instead of silently submitting bad data.

- Non-breaking: test run continues normally, no exceptions thrown
- Only `test.js` modified
- No new dependencies

## Summary by CodeRabbit

* **Tests**
  * Added balance health checks during TVL calculations to flag invalid numeric values and negative token amounts.
  * Improved warning output with token-level details when balance issues are detected.
  * Balance checks now run immediately after balances are retrieved, before further TVL computation.

### DIFF
--- a/test.js
+++ b/test.js
@@ -31,7 +31,7 @@ function checkBalanceHealth(balances, storedKey) {
   for (const [token, amount] of Object.entries(balances)) {
     const value = Number(amount)
     if (Number.isNaN(value)) {
-      nanTokens.push(token)
+      nanTokens.push(`${token} = ${amount}`)
     } else if (value < 0) {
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR, sorry but closing for the same reason as https://github.com/DefiLlama/DefiLlama-Adapters/pull/19643#issuecomment-4682836621

PR #19916, opened 2026-07-06
Add Liquidium ETH pool to TVL adapter

robin-liquidium:
Adds the live Liquidium ETH pool to the existing Liquidium TVL adapter. This follows the same pattern as the previous Liquidium USDC and ICP pool additions.

## Pool

ETH pool canister from the live lending canister `list_pools` response:

- `qcg7y-syaaa-aaaar-qb75q-cai`

## Validation

```bash
node test.js projects/liquidium
```

Result includes ETH in TVL:

```text
ETH                       1.21 k
```

Borrowed remains unchanged for ETH because the pool currently has zero outstanding ETH debt.


RohanNero:
Thank you for the PR

PR #19948, opened 2026-07-08
Add Curvance vault adapter and configure curvance adaptor to not pull in vault markets

ZeroXMai:
Adds a new `curvance-vaults` adapter for Curvance lending optimizer vault TVL and updates the existing `curvance` adapter to exclude lending optimizer share cToken markets from the parent adapter.

TVL is computed from onchain data. The adapters discover listed Curvance cToken markets from the central registry via `marketManagers()` and `queryTokensListed()`.

For the parent `curvance` adapter:
- listed cToken markets are kept as before
- lending optimizer share cToken markets are excluded programmatically
- detection uses the cToken market name suffix plus an optimizer guard on the cToken asset via `numApprovedMarkets() > 0`

For the new `curvance-vaults` adapter:
- optimizer vaults are discovered from the same listed market set
- TVL is reported from the vault token’s underlying `asset()`
- balances are computed with `totalSupply()` and `convertToAssets(totalSupply)`
- `doublecounted` is set to `true`

Tested with:
- `node test.js projects/curvance/index.js`
- `node test.js projects/curvance-vaults/index.js`

## Summary by CodeRabbit

* **New Features**
  * Added TVL tracking for Curvance vaults, including support for vault-based market discovery and asset conversion.

* **Bug Fixes**
  * Improved market filtering so certain vault-related markets are excluded from TVL and borrowed calculations.
  * Added safeguards to avoid counting blocked or duplicated markets in totals.

### DIFF
--- a/projects/curvance-vaults/index.js
+++ b/projects/curvance-vaults/index.js
@@ -37,24 +37,26 @@ Object.keys(config).forEach(chain => {
     return uniqueMarkets.filter(m => !blcklistedMarkets.includes(String(m).toLowerCase()))
   }
 
-  async function getMarkets(api) {
+  async function getVaults(api) {
     const markets = await getListedMarkets(api)
-    const optimizerVaultMarkets = new Set((await getOptimizerVaultMarkets(api, markets)).map(m => String(m).toLowerCase()))
-    return markets.filter(m => !optimizerVaultMarkets.has(String(m).toLowerCase()))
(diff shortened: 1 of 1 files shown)


RohanNero:
Sorry, for now we will continue to track tvl only in `curvance/index.js` since all vault collateral is deposited into curvance markets

PR #19887, opened 2026-07-03
AEON Protocol: migrate adapter from Avalanche to Robinhood Chain

easytigar13:
AEON Protocol (already listed, adapter merged in #19845) has migrated from Avalanche to Robinhood Chain (chainId 4663). The old Avalanche factory address is defunct.

This PR:
- Removes the stale `aeon-protocol` entry from `registries/uniswapV2.js` (was still pointing at the old Avalanche AeonFactoryRH address)
- Adds a consolidated `projects/aeon-protocol/index.js` covering all three of the protocol's pool types on Robinhood Chain:
  - vAMM (constant-product) pools, enumerated live from AeonFactoryRH's `allPools`/`allPoolsLength`
  - Algebra Integral concentrated-liquidity (CL) pools
  - Trader Joe / LFJ Liquidity Book (DLMM) pools, enumerated via the existing `joeV2Export` helper

##### Website Link:
https://aeonprotocol.net

##### Chain:
Robinhood Chain

##### forkedFrom (Does your project originate from another project):
No — vAMM/veNFT/voting contracts are original. The CL and DLMM pool types fork the open-source Algebra Integral and Trader Joe/LFJ Liquidity Book engines respectively (both MIT/GPL licensed), same as most DEXes on DefiLlama do for their concentrated-liquidity or bin-liquidity products.

##### methodology (what is being counted as tvl, how is tvl being calculated):
TVL counts tokens held directly by AEON Protocol's pool contracts on Robinhood Chain, summed across all three pool types.

Please enable "Allow edits by maintainers" — done.

## Summary by CodeRabbit

* **New Features**
  * Added TVL support for Aeon Protocol on the Robinhood network, aggregating balances across multiple pool types (including vAMM, concentrated-liquidity, and DLMM) for more complete reporting.
  * Included a methodology description to clarify how TVL is calculated.
* **Chores**
  * Adjusted minor formatting in a Uniswap V2 registry configuration entry for consistency.

PR #19972, opened 2026-07-10
AEON Protocol: fix DLMM ABI, guard pool failures, cover pools added 2026-07-09

easytigar13:
Follow-up to #19887 (AEON Protocol Robinhood Chain migration, already merged). Since then a few real issues surfaced:

- **DLMM was reporting $0**: `joeV2Export` was called with `isLb: true`, which selects the `allLBPairs`/`tokenX` ABI variant. This factory actually exposes the v1-style `getLBPairAtIndex`/`getNumberOfLBPairs`/`getTokenX`/`getTokenY` (verified on-chain -- `allLBPairs()`/`tokenX()` both revert, the v1 getters do not).
- **One bad pool could zero the whole adapter**: added `permitFailure: true` so a single reverting/stale pool or token call (e.g. pre-migration vAMM dust) only excludes that pool instead of failing the whole TVL calc.
- **8 vAMM pools + 7 CL pools were missing from TVL entirely**: the protocol deployed a second vAMM factory (`AeonFactoryV2`) on 2026-07-09 that all new pool creation now targets exclusively, but the adapter only ever queried the original factory. Now enumerates both dynamically (so future pools on the current factory need no adapter change), and the 7 new Algebra CL pools deployed the same day for the same pairs are added to the static CL list.

##### Website Link:
https://aeonprotocol.net

##### Chain:
Robinhood Chain

##### methodology (what changed):
Same methodology as #19887 (sum of tokens held directly by pool contracts across vAMM/CL/DLMM). This PR only fixes coverage/correctness bugs in that same calculation -- no new pool types, no methodology change.

Please enable "Allow edits by maintainers" -- done.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved AEON TVL reporting by aggregating balances across both vAMM factories.
  * Expanded Algebra CL pool and token coverage, including additional migrated pools, for more complete and accurate TVL figures.
* **Documentation**
  * Updated the TVL methodology description to match the new multi-factory enumeration and explicitly listed pools.

### DIFF
--- a/projects/aeon-protocol/index.js
+++ b/projects/aeon-protocol/index.js
@@ -1,36 +1,20 @@
-const { sumTokens2 } = require('../helper/unwrapLPs')
+const sdk = require('@defillama/sdk')
+const { sumTokensExport } = require('../helper/unwrapLPs')
+const { getUniTVL } = require('../helper/unknownTokens')
 const { joeV2Export } = require('../helper/traderJoeV2')
 
-// AeonFactoryRH — the original constant-product (vAMM) pool factory. Not a
-// standard UniswapV2Factory: pools are exposed via a public `allPools` array
-// + `allPoolsLength()` instead of `allPairs`/`allPairsLength`. Kept for TVL
-// purposes since some already-created pools (now dust) still live here.
 const AEON_FACTORY = '0xD8495E398Fd7F0293Ccfca4a16181216CfDa6ED6'
-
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR, the JoeExports helper is returning non-zero tvl that matches your refactored version's output when I test it so I left it untouched

PR #20001, opened 2026-07-13
Add Robinhood Chain to T3tris Finance TVL adapter

t3trominos:
T3tris Finance is now deployed on Robinhood Chain (4663). This adds `robinhood` to the TVL adapter's chain list. Vaults are auto-discovered from the T3tris ecosystem API per chain via `api.chainId` (same curation gate: verified & non-blacklisted), so Robinhood TVL surfaces automatically as verified vaults are published there. No other changes.

## Summary by CodeRabbit

* **New Features**
  * Added TVL tracking support for the Robinhood chain.
  * Robinhood vaults are now discovered, verified, and included in TVL reporting.


RohanNero:
Thanks for the PR

PR #20007, opened 2026-07-13
fix(mellow-protocol-core): dedupe vault list, split multiCall batch to fix silent drops on mezo

### LINKED ISSUE #20004: mellow-protocol-core: mezo-cbbtc-vault silently dropped, mezo-stable-vault duplicate produces false $0  ~$130k undercounted
`projects/mellow-protocol-core/index.js` significantly undercounts Mezo chain TVL. Verified via `node test.js projects/mellow-protocol-core/index.js mezo` with a temporary debug line on the real `coreVaultsResults` array (not a substitute script):

VAULT mezo-stable-vault 0x07AFFA6754458f88db83A72859948d9b794E131b -> SUCCESS
VAULT mezo-btc-vault    0x807D4778abA870e4222904f5b528F68B350cE0E0 -> SUCCESS
VAULT mezo-stable-vault 0x07AFFA6754458f88db83A72859948d9b794E131b -> NULL/FAILED
VAULT mezo-cbbtc-vault  0x06ED1E2167AA7FBf2476c5A2D220Bf702559Dcf8 -> NULL/FAILED

**Real, cross-checked TVL per Mellow's own API** (`https://api.mellow.finance/v1/vaults`, filtered `chain_id: 31612`):
- mezo-btc-vault: ~$82,217 — **correctly counted**
- mezo-stable-vault: ~$31,017 — **reported as `mUSDC: 0.00`**, not the real value
- mezo-cbbtc-vault: ~$98,761–$100,680 — **completely missing, not even a $0 line**

Current adapter output for `mezo`: **$82.21k**. Real total across all 3 vaults: **~$212k**. Undercounted by **~$130k, roughly 60%**.

**Two separate root causes:**
1. **Duplicate vault entry from Mellow's own API**  `mezo-stable-vault` (`0x07AFFA67...`) appears twice in the API response (once with `base_token.symbol: "USDC"`, once `"mUSDC"`), same address both times. `permitFailure: true` + falsy-result check means the second occurrence gets silently dropped rather than treated as a duplicate. Separately, the `mUSDC: 0.00` line in the final output suggests `totalBaseAsset`  a `BigNumber` instance  passes the `if (!totalBaseAsset) return` truthiness check even when it resolves to zero, since a `BigNumber(0)` object is still truthy in JS.
2. **`mezo-cbbtc-vault`'s `collector.collect()` call genuinely fails** (`permitFailure: true` masks this into a silent `null` rather than a visible error)  root cause of the revert itself not yet identified, worth a maintainer's input since the collector logic may need a fix specific to this vault's token/queue configuration.

Happy to help debug the `collect()` revert further if useful  wanted to flag the quantified impact first.

BhariGowda:
Fixes #20004.

Two real bugs fixed:

1. Mellow's own API lists mezo-stable-vault twice (once with base_token symbol "USDC", once "mUSDC"), same address both times. Deduped `vaultsOnChain` by address before splitting into `coreVaults`/`dvvVaults`.

2. `coreVaults`' `collect()` calls were all batched into one `multiCall`. Verified directly (alone = success, batched at position 3+ = null) that this exceeds a Multicall3 aggregation limit given `collect()`'s large nested-tuple return type. Now issuing one `multiCall` per vault via `Promise.all`, avoiding the limit while staying concurrent.

Note: `coins.llama.fi` has incorrect decimals registered for mUSDC and mcbBTC on Mezo (18, vs the real on-chain 6 and 8  verified via direct `decimals()` calls). This adapter now correctly computes and adds real, non-zero balances for these tokens, but displayed USD values will remain wrong until that separate pricing-metadata issue is fixed. Filing that separately since it's not an adapter-code problem.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vault discovery by filtering to the selected blockchain, removing duplicate vault entries, and only using vaults with positive supply.
  * Updated core-vault TVL aggregation to more reliably combine results across multiple vaults.
  * Preserved accurate total value calculations by continuing to exclude double-counted assets.
* **Performance**
  * Added caching for the remote vault configuration used during TVL computation.

### DIFF
--- a/projects/mellow-protocol-core/index.js
+++ b/projects/mellow-protocol-core/index.js
@@ -1,7 +1,6 @@
 const BigNumber = require('bignumber.js');
 const ADDRESSES = require('../helper/coreAssets.json')
 const { getConfig } = require('../helper/cache');
-
 const abi = {
   collect: 'function collect(address,address,(address,uint256,uint256)) view returns ((address,address,address[],uint8[],uint256[],(address,address,bool,bool,bool,uint256,uint256[])[],uint256,uint256,uint256,uint256,uint256,uint256,uint256,uint256,uint256,uint256,uint256,(address,address,uint256,uint256,uint256,uint256)[],(address,address,uint256,uint256,uint256,uint256)[],uint256,uint256))',
   shareManager: 'function shareManager() view returns (address)',
@@ -12,9 +11,7 @@ const abi = {
   balanceOf: 'function balanceOf(address) view returns (uint256)',
   asset: 'function asset() view returns (address)',
 }
-
 let _vaultsApiResponse
-
 const getDoubleCountedShares = async ({api, coreVaults, dvvVaults, coreVaultsResults}) => {
   const shareManagerByVault = {}
   const coreVaultShareManagers = await api.multiCall({ calls: coreVaults.map(vault => vault.address), abi: abi.shareManager, permitFailure: true })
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for flagging these issues and for the PR. I've refactored slightly to drop the permitFailure to prevent silently undercounting in the future, and we will fix the decimals metadata for mUSDC and mcbBTC on the server

BhariGowda:
Appreciate the explanation  good call preferring a loud failure over a
silent null here, will default to that pattern instead of permitFailure+filter
next time this shape of bug comes up. Thanks for picking up the decimals fix too.

PR #20011, opened 2026-07-13
quainance: permit failed balanceOf calls to avoid zeroing the whole adapter

BhariGowda:
Both `api.multiCall` calls in `tvl()` (fetching token0/token1 balances
per pair) had no `permitFailure`, so a single reverting `balanceOf`
call - a dead pair, a blacklisted address, a non-standard token -
would throw and cause the entire Quai chain adapter to report zero
TVL, instead of just skipping that one bad pair.

`registries/uniswapV2.js` already sets `permitFailure: true` for
several existing UniV2-clone adapters (`baseswap`, `cubiswap`,
`elvesdex`) for exactly this reason - this is a known, recurring
failure mode for this adapter shape across the repo, not a
hypothetical one specific to this protocol.

Fix: added `permitFailure: true` to both `multiCall` calls, and
filter out any pair where either balance comes back `null` before
passing to `transformDexBalances`.

Verified locally: `node test.js projects/quainance/index.js` still
runs clean, same ~$4.24k total TVL across both currently-tracked
pools - no live pair is failing right now, this is a preventive fix
for a real, established failure class on this adapter shape, not a
fix for a currently-visible crash.

Note: rewrote the whole file rather than a line-patch, which
incidentally dropped a blank line after the imports and added the
missing trailing newline at EOF - happy to revert either if preferred.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DEX balance data handling by tolerating individual token balance lookup failures.
  * Excluded incomplete balance records from calculations to prevent inaccurate results.

### DIFF
--- a/projects/quainance/index.js
+++ b/projects/quainance/index.js
@@ -1,7 +1,6 @@
 const { getLogs2 } = require('../helper/cache/getLogs')
 const { transformDexBalances } = require('../helper/portedTokens')
 
-
 module.exports = {
   misrepresentedTokens: true,
   quai: {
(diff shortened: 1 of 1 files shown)


BhariGowda:
Closing this - projects/quainance/index.js was deleted in fe37ed06c
(consolidated into registries/uniswapV2Logs.js alongside kaleidoswap
and mantleswap). The same fix (permitFailure + null-filter) is now
applied to that shared location in #20035, which covers this
protocol plus the other two sharing the same code path.

PR #20030, opened 2026-07-14
add adi and remove several chains from orderly

Tarnadas:
Hey,

Orderly just launched on ADI chain. We also no longer support several other chains, so I removed them.

Please see official docs for full list of supported chains:
https://orderly.network/docs/build-on-omnichain/addresses

## Summary by CodeRabbit

* **Chores**
  * Updated supported network ownership mappings.
  * Added the `adi` network and removed several outdated network mappings.


RohanNero:
Thank you for the PR, I've added adi in #20031. I've kept the other chains that were removed here since they still have non-zero tvl

PR #20042, opened 2026-07-15
Fix two loan hashes for USDai

wegory:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved borrowed TVL reporting by including active Loan Router V2 loans that were previously omitted.
  * Ensured balances from these loans are reflected in the reported USDai total value locked.


RohanNero:
Thank you for the PR. I couldn't edit this one, so I added a blacklist and the two hard-coded loan hashes in #20044. I will fix the USDai chart shortly so there isn't a gap

PR #20048, opened 2026-07-15
Add Spark's Robinhood and XLayer allocations

b-tarczynski:
Addresses taken from:
- [Robinhood](https://github.com/sparkdotfi/spark-address-registry/blob/master/src/Robinhood.sol)
- [xLayer](https://github.com/sparkdotfi/spark-address-registry/blob/master/src/XLayer.sol)


RohanNero:
Thank you for the PR

PR #20051, opened 2026-07-15
fix: use a unique cache key for TeleSwap lockers

amancslab:
## Summary

Give TeleSwap its own cache key instead of sharing `yala/bitcoin`.

## Problem

Yala and TeleSwap are fetched concurrently by the Bitcoin duplicate checker. Because both used the same cache key, the cached Yala address array could be returned to TeleSwap, causing:

`TypeError: Cannot read properties of undefined (reading 'lockers')`

## Validation

- Reproduced the failure with `Promise.all([f.yala(), f.teleswap()])`
- Confirmed the same concurrent call returns independent address lists after the fix
- `pnpm exec eslint projects/helper/bitcoin-book/fetchers.js`
- `git diff --check`

No dependencies or lockfiles were changed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected configuration and cache loading for Bitcoin locker data.
  * Preserved existing locker retrieval and filtering behavior.

### DIFF
--- a/projects/helper/bitcoin-book/fetchers.js
+++ b/projects/helper/bitcoin-book/fetchers.js
@@ -272,7 +272,7 @@ module.exports = {
     })
   },
   teleswap: async () => {
-    const  { data: { lockers } } = await getConfig('yala/bitcoin', 'https://api.teleportdao.xyz/api/v1/teleswap/lockers/')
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR

PR #20090, opened 2026-07-16
feat: roll to YLDS in the new demo prime pools

jarrydallison:
This PR increases accuracy with Democratized Prime reporting by switching to the smart contracts and reporting on liquidity vs. borrowed as recorded on chain. It also removes the old pool calculations from the figure markets adapter.


jarrydallison:
@RohanNero just checking if you need anything else from me here?

RohanNero:
Apologies for the delay, and thank you for the PR

PR #20103, opened 2026-07-17
yearn: Fix nested vault TVL double counting

rossgalloway:
## Summary

- Source Yearn vault metadata and allocator debt relationships from Kong, Yearn's maintained indexer and replacement for yDaemon.
- Remove parent allocations from nested child-vault balances while preserving capital deposited directly into those children.
- Detect same-asset nested vaults automatically.
- Handle the verified DAI to USDS and DAI to USDC routes with decimal normalization.
- Exclude Katana bridge pre-deposit vaults from Yearn TVL.

## Why

The previous adapter avoided double counting by excluding every vault that appeared as another vault's strategy. That also excluded capital deposited directly into nested vaults.

This change includes nested vaults but subtracts the debt already reported by their parent allocators, counting shared capital once while retaining independent deposits.

## Validation

- `node --test projects/yearn/helpers.test.js` - 8 tests passed
- `npx eslint projects/yearn/index.js projects/yearn/helpers.js projects/yearn/helpers.test.js`
- `npm run build`
- `node test.js projects/yearn/index.js` - passed across all configured chains

## Maintainer note

The adapter-local Node test suite passes when invoked explicitly. The current project CI does not appear to auto-discover `projects/**/*.test.js`; maintainers may want to decide whether this test should be invoked explicitly or through general test discovery.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Yearn vault valuation across supported chains by using up-to-date vault listings and on-chain asset balances.
  * Excluded inactive or empty vaults from TVL calculations.
  * Improved handling of nested vault debt, cross-asset routes, and different vault versions.
  * Added fallback token resolution and filtered out non-positive balances for more accurate attribution.
  * Preserved dedicated valuation for Ethereum V1 vaults.


rossgalloway:
@coderabbitai review

RohanNero:
Thank you for the PR, I refactored to retrieve strategies and debt on-chain since the kong snapshots can be stale. Also, noticed two other minor issues with the kong data:
- some vault versions are marked v2 when they are v3 (resolved via `apiVersion()`)
- some v2 vaults have more than 20 strategies despite `MAXIMUM_STRATEGIES = 20` (resolved by capping at 20)

PR #20124, opened 2026-07-19
Mantle treasury: BIT is an own token, and native MNT has to be read on Mantle

tarun-khatri:
Two problems in `projects/treasury/bitdao.js`, both in the own-token accounting.

### 1. BIT is counted as a third-party treasury asset

The Ethereum `tvl` call blacklists only `MNT`, so BIT — the BitDAO token Mantle rebranded from and converted to MNT — lands in the regular treasury bucket. It is the single largest position there:

```
/treasury/bitdao  Ethereum  ->  BIT 105,469,776.02, SUSDE 13,728,908, USDT 11,522,803, ...
```

which is exactly the amount the protocol's own API reports for that wallet:

```
id 0x1a4b46696b2bb4794eb3d4c26f1c55f9170fa4c5  chain eth
wallet 0x78605df795...   amount 105469776.01718912
```

At $0.438479 (CoinGecko, live) that is **$46,246,282** sitting in the non-own treasury figure — 29.5% of the reported $156,856,507 Ethereum bucket.

### 2. Native MNT on Mantle is read from Ethereum

```js
if (tokens.some(token => token.address === 'mnt')) api.add(MNT, (await sdk.api.eth.getBalance({ target: owner })).output, { skipChain: true });
```

`sdk.api.eth.getBalance` with no `chain` reads Ethereum. This branch only fires for the Mantle pass (the API has 58 rows with `id = 'mnt'`, all `chain = 'mnt'`, and none on eth), so what gets added is each Mantle wallet's *Ethereum* balance, labelled as MNT. The published result:

```
/treasury/bitdao  Mantle-OwnTokens  ->  COOK 1,392,892,853.24,  MNT 301.55
```

301.55 MNT, where those wallets actually hold **29,175,917.79** MNT (summed from the same API), worth **$12,303,222** at $0.421691.

### Changes

- add `BIT` to the Ethereum `blacklistedTokens` and to the Ethereum `ownTokens` owner-token list
- pass `chain: api.chain` to the native-MNT `getBalance` so it is read on Mantle
- skip the native-MNT add on the Mantle `tvl` pass via a `skipOwnNative` flag, so MNT is only counted in `ownTokens`

That third change is needed because `getTvlData` mutates `api` and is called by both the `tvl` and `ownTokens` exports. Today that double-counts 301 MNT, which is invisible; once the balance is read correctly it would have put the full $12.3M into both the Mantle tvl bucket and Mantle-OwnTokens. I first tried adding MNT to `blacklistedTokens` on the Mantle tvl call, but that does not filter a balance added through `api.add(..., { skipChain: true })` — the run still showed mantle at 156.16 M — so the add itself has to be skipped.

### Result

```
                     before (published)        after (local run)
ethereum              $156,856,507              110.92 M
ethereum-ownTokens                              1.24 B
mantle                $143,715,743              143.86 M
mantle-ownTokens        $3,637,667               15.94 M
```

Ethereum drops by the $46.2M of BIT, Mantle-OwnTokens gains the $12.3M of native MNT, and the Mantle tvl bucket is left where it was rather than picking up a new double count.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated BitDAO treasury value calculations to properly reflect the BIT token alongside MNT.
  * Improved treasury accounting on both Ethereum and Mantle by more accurately distinguishing owned tokens from native assets.
  * Corrected Mantle handling of COOK and native token balances in reported treasury metrics.

### DIFF
--- a/projects/treasury/bitdao.js
+++ b/projects/treasury/bitdao.js
@@ -1,80 +1,64 @@
 const ADDRESSES = require('../helper/coreAssets.json')
-const sdk = require('@defillama/sdk');
-const { nullAddress } = require("../helper/treasury");
 const { getConfig } = require('../helper/cache')
 
 const API_URL = 'https://api.mantle.xyz/api/v2/treasury/tokens';
+
 const MNT = '0x3c3a81e81dc49a522a592e7622a7e711c06bf354';
-// BitDAO's original token, which Mantle rebranded from and converted to MNT
 const BIT = '0x1a4b46696b2bb4794eb3d4c26f1c55f9170fa4c5';
 const USDe = ADDRESSES.ethereum.USDe;
 const COOK = '0x9f0c013016e8656bc256f948cd4b79ab25c7b94d'
 const ethenaFarm = '0x8707f238936c12c309bfc2b9959c35828acfc512';
-const SPECIFIC_TOKENS = ['eth', 'ethena-farming-usde', 'eigen-layer-eth', 'mnt'];
 
 const abi = "function stakes(address, address) view returns (uint256 stakedAmount, uint152 coolingDownAmount, uint104 cooldownStartTimestamp)";
 
-const isSpecificToken = token => SPECIFIC_TOKENS.includes(token);
+const config = {
+  ethereum: { key: 'eth', ownTokens: [MNT, BIT] },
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR

PR #20162, opened 2026-07-21
Add Morpho Midnight adapter (Base)

tomrpl:
# Add Morpho Midnight adapter (Base)

This adds a TVL + borrowed adapter for **Morpho Midnight**, Morpho's fixed-rate, fixed-term lending protocol, deployed on Base at `0xAdedD8ab6dE832766Fedf0FaC4992E5C4D3EA18A` (canonical Midnight core contract).

Files: `projects/morpho-midnight/index.js`, `projects/morpho-midnight/helpers.js`.

> **Scope of this PR (updated per maintainer feedback).** This adapter counts **only value custodied by the Midnight contract**, reconstructed from onchain events. Committed lend offers — capital that, by design, sits in makers' own wallets until a loan is matched — are **not** counted as TVL (same rule as 1inch Aqua / OpenSea listings). We'd like to land this custodied-only version now, and separately discuss reporting committed offers as their own "committed liquidity" metric in a follow-up.

## What Midnight is (and why it isn't Morpho Blue)

Midnight is not a pooled AMM-style lender. It's an **order book of fixed-rate, fixed-maturity credit**. Lenders post offers to lend at a given rate/tick; when an offer is matched (`Take`), a loan is created: the borrower posts collateral into the contract and draws the loan token. A matched loan is a real, executed onchain position — collateral is custodied by Midnight and debt is outstanding. That executed, custodied state is exactly what this adapter measures.

## Methodology — 100% onchain, event-reconstructed

Both metrics are reconstructed purely from events emitted by the Midnight contract. No API is used for valuation; the Morpho API is used only to scope which markets are listed (see below), and never contributes a value.

### Borrowed

```
borrowed(market) = Σ Take.totalUnitsDelta  −  Σ Repay.units  −  Σ Liquidate(repaidUnits + badDebt)
```

`Take.totalUnitsDelta` is the signed change in the market's aggregate outstanding credit emitted by the contract on each match; repayments, liquidations and realized bad debt net it back down. This is the active matched-loan figure — it only reflects positions that actually executed onchain, so it cannot be gamed.

### TVL — custodied value only

`tvl` is the real ERC-20 value held by the Midnight contract, reconstructed from events:

- **collateral deposited:** `SupplyCollateral` − `WithdrawCollateral` − `Liquidate.seizedAssets`
- **repaid loan liquidity not yet withdrawn:** `Repay` / `Liquidate.repaidUnits` − `Withdraw` − `ClaimContinuousFee`

Matured markets remain included until their collateral and repaid liquidity are withdrawn — that residual is real value still in the contract.

## Scope: listed markets only (for launch)

At launch the adapter is gated to markets flagged **`listed`** by the Morpho API trust layer. This is the only use of the API — it is read-only scoping and contributes no value. The same listed set gates **both** `tvl` and `borrowed`, so the two metrics never mix scopes. Listed ids are intersected with `MarketCreated` events from the configured Midnight contract, so a listing entry from any other deployment cannot widen scope. This single filter can be removed later to cover every Midnight market in the blacklist-not-whitelist spirit; matured listed markets are intentionally kept so their residual liquidity stays counted.

## Validation performed

- **Onchain reconstruction vs reality:** event-reconstructed contract balances were compared against live `balanceOf(Midnight)` for every token — **10 of 11 tokens match to the exact wei**, including all collateral. The lone difference is the loan token, where the pure-event model is slightly *conservative* (~0.03% of TVL) because it deliberately excludes accrued protocol fees sitting in the contract.
- **Borrowed vs protocol state:** event-reconstructed `borrowed` was checked against the protocol's own `total_units` per market across all markets — the invariant `borrowed ≤ total_units` (outstanding debt never exceeds outstanding credit) holds everywhere.

## Checklist

- [x] Runs green locally (`node test.js projects/morpho-midnight`)
- [x] `methodology` string set
- [x] No hardcoded prices; values are added by token address and priced by the SDK
- [x] TVL and borrowed exported per chain (`base`)
- [x] Lint clean

## Summary by CodeRabbit

* **New Features**
  * Added support for tracking Morpho Midnight markets.
  * Added total value locked (TVL) reporting for supported loan and collateral tokens.
  * Added borrowed amount reporting by market.
  * Added filtering for blacklisted markets and tokens.


RohanNero:
Thank you for the PR

PR #20177, opened 2026-07-22
feat(elara): calculate TVL from NAV aggregator

vlad-truefi:
## Summary by CodeRabbit

* **New Features**
  * Added Ethereum TVL tracking for Elara Finance.
  * TVL is calculated from the protocol’s reported net asset value and displayed in USD.
  * Included the tracking start date and methodology details.

### DIFF
--- a/projects/elara-finance/index.js
+++ b/projects/elara-finance/index.js
@@ -1,0 +1,16 @@
+const NAV_AGGREGATOR = '0x507f997c64dE57e189ed1ce95b832CE7b65FAc50'
+const GET_NAV =
+  'function getNAV() view returns (uint256 navUsd, uint256 walletBalanceUsd, uint256 positionValueUsd, uint64 computedAt)'
+
+const tvl = async (api) => {
+  const { navUsd } = await api.call({ target: NAV_AGGREGATOR, abi: GET_NAV })
+  api.addUSDValue(Number(BigInt(navUsd) / 10n ** 12n) / 1e6)
+}
+
+module.exports = {
+  start: '2026-06-16',
+  methodology: 'TVL is Elara\'s gross on-chain NAV across registered treasury wallets and strategy positions.',
+  ethereum: {
+    tvl,
+  },
+}


RohanNero (changes requested):
Hello, thank you for the PR. Could you please rename the file to avoid a conflict with the existing 'elara' adapter, and could you provide a link to the elara website where users can deposit?

Also, apologies for the inconvenience, but we prefer not to track TVL from EOAs anymore, do any of the Elara smart contracts hold user deposits?

vlad-truefi:
@RohanNero Thanks for your feedback! The adapter has been renamed. Here is the Elara app: [app.elara.fi](https://app.elara.fi).

Regarding EOAs: Elara uses treasury wallets as the custody layer for both idle token balances and on-chain strategy positions. Since not all liquidity is continuously deployed, accurate TVL must include both wallet balances and strategy positions. The `NAVAggregator` calculates this entirely from on-chain state and only includes explicitly registered wallets, tokens, and strategy adapters. This is why the current adapter uses its `getNAV()` result.

RohanNero:
@vlad-truefi Thank you for clarifying. Sorry again for the inconvenience, but since we prefer not to include EOA balances in tvl adapters, we can either track the  getNav() value as treasury in `projects/treasury/elara.js` or we can exclude the EOA balances from this adapter. Alternatively, if preferred, we can continue to track only fees for elara: https://defillama.com/protocol/elara-finance

vlad-truefi:
@RohanNero Thanks, that makes sense. Would you be open to splitting the NAV components? We could report `positionValueUsd` as TVL, since those assets are deployed in on-chain strategy contracts, and report `walletBalanceUsd` separately under treasury. The two values are returned independently by `getNAV()`, so this would exclude idle EOA balances from TVL while preserving visibility of Elara’s complete NAV without double-counting.

RohanNero:
@vlad-truefi That could work, can you share a verified StrategyAdapter contract address or the source code? We can split if we can track the strategy positions without using `getNav()`

vlad-truefi:
@RohanNero 

Sure. `NAVAggregator` currently tracks 12 StrategyAdapter deployments:

- 1 Pendle adapter
- 1 ERC-4626 adapter for Ethena sUSDe
- 10 Uniswap V4 position adapters

The Pendle adapter is verified here:
https://etherscan.io/address/0x9f39e9C818B45208b2217cc7799534409f6e5b6f#code

`NAVAggregator` exposes the registered wallets and adapters through `walletCount/getWallet` and `adapterCount/getAdapter`. We can enumerate them and sum `getPositionValue(wallet)` directly, without calling `getNAV()` or including direct wallet balances.

RohanNero:
Sorry, but since the EOAs deposit into external protocols and hold receipt tokens themselves as opposed to being deposited into the strategy adapters (e.g. 0x9f39e9C818B45208b2217cc7799534409f6e5b6f), we've decided to list the entire NAV as treasury. I've added elara in #20483

vlad-truefi:
@RohanNero I apologise for the inconvenience. Could you please confirm where we can find the added treasury data on the DeFiLlama web app for Elara Finance? Nothing is there: https://defillama.com/protocol/elara-finance

PR #20207, opened 2026-07-24
feat:The Haedal AMM protocol will no longer be displayed.

Felix-Qc:
Hello, we are the Haedal team. 
We would like to request that DefiLlama no longer display data related to the Haedal AMM protocol. 
Thanks.


RohanNero:
Sorry, but we typically don't remove listings once added for reasons described here: https://github.com/DefiLlama/DefiLlama-Adapters/discussions/13020

Felix-Qc:
Yes, we are aware of these rules. However, we are not actually planning to run this module anymore. To avoid ambiguity for users, we want to hide it.

PR #20210, opened 2026-07-24
feat:change haedal-amm

Felix-Qc:
Hello, we are the Haedal team.

We would like to request that DefiLlama stop displaying data related to the Haedal AMM protocol.

We are aware of the policies regarding data concealment. However, as we are no longer operating the AMM module, we would like to have it hidden to avoid confusing users.

Thanks.


RohanNero (changes requested):
We can mark haedal-amm as deprecated to avoid confusing users instead of completely removing it. What date was it announced or changed to deprecated in the app?

Felix-Qc:
Yes, we also want to mark haedal-amm as deprecated. This will begin on July 28, 2026.Thanks.

RohanNero:
Thank you, closing as we have marked haedal AMM as deprecated

PR #20212, opened 2026-07-24
Remove B2 Buzz non-BTC chain configs (BNB/Ethereum/Polygon)

### LINKED ISSUE #20208: Possible double-count: Free Protocol and B² Buzz both count the same B²-controlled custody multisigs
Title: Possible double-count: Free Protocol and B² Buzz both count the same B²-controlled custody multisigs

Flagging a small cross-adapter overlap for a methodology check, not reporting a bug in either adapter. Wanted to lay out what's confirmed on-chain and ask which side should carry a `doublecounted` flag, if either.

**The overlap**

Two adapters list the same two addresses as `owner`/`ownerTokens` balance targets:

- `projects/bsquared/index.js` (B² Buzz) counts them via `ownerTokens`, labeled in-code as "Ethereum Multisignature Address" and "BNB Chain Multisig Address"
- `projects/free-protocol/index.js` (Free Protocol) counts them in its `owners` arrays

Shared addresses:
- `0xeea3A032f381AB1E415e82Fe08ebeb20F513c42c` (Ethereum)
- `0x0A80028d73Faaee6e57484E3335BeFda0de7f455` (BNB Chain)

**What's confirmed on-chain**

Both addresses are Gnosis Safes with the identical 5-owner signer set (threshold 3/5) on both chains:

```
0x8Da30b68b1E11761e1307EB10564a39E01ae4480
0x78Cf44706D948D983e69511Fac95dDce4cBf0711
0x97c2fC2Ba6Cc9cC3dc647447127392A2094D8c2f
0x9E498DD03c5E984C105E83221AA911DEC4844dB5
0x42A4C0905841cAC468142939E69a54fd266e95D8
```

So both multisigs are operated by the same signer set (the ones bsquared labels as B²'s own custody).

Only these two addresses overlap. The other 15 owner addresses in free-protocol are its own and don't collide with bsquared.

**Current live magnitude is small.** Right now the only non-zero shared holding that both adapters actually count is on the Ethereum Safe:

- `0xeea3A032...` holds ~111.79 STONE (`0x7122985656e38BDC0302Db86685bb972b145bD3C`), ~$223K at current price
- bsquared counts STONE there via `ownerTokens: [[STONE, USDC, USDT, WBTC], "0xeea3A032..."]`
- free-protocol counts STONE there via `owners: [..., "0xeea3A032..."]` + `tokens: [..., STONE, ...]`

WBTC/USDC/USDT on the Ethereum Safe and BTCB/FDUSD on the BSC Safe are currently ~0, so today's double-counted amount is roughly that one STONE balance. The point is more the structure than the current dollar figure, since the shared addresses could be funded again later.

**The question**

Since the same on-chain balance backs both protocols' reported TVL, one side likely needs `doublecounted: true` (or the address removed from whichever side shouldn't be claiming it). I can't tell from on-chain alone whether Free Protocol deposits into B²'s custody as an intended integration or whether these addresses ended up in the Free Protocol adapter by accident.

Tagging the original adapter authors who'd know the context: @Stonepapa (added B² Buzz's adapter, #8992) and @LeviathanYou (added Free Protocol's adapter, #10200). Was this shared custody an intended integration? If so, which side should carry the `doublecounted` flag, and I'm happy to open the one-line PR once it's clear which way it should go.

BhariGowda:
Removes the BNB Chain, Ethereum and Polygon configs from projects/bsquared/index.js (the STONE/FDUSD/POL balances, ~$250K total), plus the now-unused coreAssets and STONE-address imports that only served those blocks. Bitcoin tracking is left completely untouched.

Context from #20208:
- Stonepapa (original B² Buzz adapter author, #8992) confirmed the B² Buzz pre-deposit event has ended.
- The assets on BNB Chain, Ethereum and Polygon are managed by a third-party bridge, not the official B² Network bridge.
- After I flagged that the Bitcoin leg (~2,814.70 BTC, ~$183.77M via the bsquaredBTC address list in bitcoin-book/index.js) is only counted by this adapter and nothing else, Stonepapa confirmed the correct scope is to keep bsquared/index.js and only remove the BNB/Ethereum/Polygon configs. Bitcoin stays as-is.

Verification:
- node test.js projects/bsquared/index.js runs clean after the change.
- Post-change TVL is bitcoin-only: ~$182.8M (2.81k BTC), matching the pre-change Bitcoin figure. The BNB/Ethereum/Polygon totals are gone.
- Confirmed nothing else in the repo consumes bsquaredBTC or references the removed multisig addresses, so no other adapter is affected.

Closes #20208

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated the staking TVL display to include only Bitcoin-based TVL.
  - Removed Ethereum, Polygon, and BSC TVL calculations from the reported configuration.
  - Kept the BRC20 TVL calculation disabled.

### DIFF
--- a/projects/bsquared/index.js
+++ b/projects/bsquared/index.js
@@ -1,10 +1,7 @@
 const sdk = require('@defillama/sdk');
 const { sumTokensExport: sumBRC20TokensExport } = require("../helper/chain/brc20");
 const { sumTokensExport } = require('../helper/sumTokens');
-const ADDRESSES = require('../helper/coreAssets.json')
 const bitcoinAddressBook = require('../helper/bitcoin-book/index.js')
-
-const ADDRESSES_ETHEREUM_STONE = '0x7122985656e38BDC0302Db86685bb972b145bD3C';
 
 module.exports = {
   hallmarks: [
@@ -16,26 +13,5 @@ module.exports = {
       sumTokensExport({ owners: bitcoinAddressBook.bsquaredBTC }),
       // sumBRC20TokensExport({ owners: bitcoinAddressBook.bsquaredBRC20 }),  // disable brc20
     ]),
-  },
-  ethereum: {
-    tvl: sumTokensExport({
-      ownerTokens: [
-        [[ADDRESSES_ETHEREUM_STONE, ADDRESSES.ethereum.USDC, ADDRESSES.ethereum.USDT, ADDRESSES.ethereum.WBTC], "0xeea3A032f381AB1E415e82Fe08ebeb20F513c42c",], //Ethereum Multisignature Address (WBTC)
-      ]
-    }),
-  },
-  polygon: {
-    tvl: sumTokensExport({
-      ownerTokens: [
-        [[ADDRESSES.null], "0x01cE88498ED095d386e09834D32Fd8f1FeCd184a",],
-      ]
-    }),
-  },
-  bsc: {
-    tvl: sumTokensExport({
-      ownerTokens: [
-        [[ADDRESSES.bsc.BTCB, ADDRESSES.ethereum.FDUSD], "0x0A80028d73Faaee6e57484E3335BeFda0de7f455",], //BNB Chain Multisig Address (BTCB)
-      ]
-    }),
   },
 };


RohanNero:
Thank you for flagging this and opening a PR

PR #19640, opened 2026-06-11
add cookiechain (refactor)

RohanNero:
resolves #19558

## New Features
* Cookie Chain network is now supported for total value locked (TVL) calculations.
* Integrated TVL computation across multiple liquidity pool types on Cookie Chain, including decentralized exchanges and farming vaults.
* Added Cookie Chain RPC endpoint configuration to enable network connectivity.

PR #20213, opened 2026-07-24
fix: cookiechain-bridge and add Hyperlane escrow to Solana-side TVL

fibanachos:
## Summary

`cookiechain-bridge` currently has **two conflicting definitions** in this repo, and the one that computes is stale. This PR consolidates them into a single, correct adapter and updates the TVL to reflect the bridge's migration to Hyperlane.

- The **CookieChain Bridge** moves COOK between **Solana** and **Cookie Chain** 1:1.
- It has migrated to a **Hyperlane warp route**. The bridge's COOK on Solana now sits in two places: the original **Squads multi-sig** reserve and the new **Hyperlane warp escrow PDA**.

### 1. `projects/cookiechain-bridge/index.js` — rewritten
Now counts the bridge's COOK on the **Solana side only**:
- `88q7zoKctwAQRsoTxkMJy95sNE3tntuyEhSrhvR1eZwq` — Hyperlane warp escrow (SPL COOK, Token-2022)
- `DoYYCtcG2vfrE3HtxBBXiNVieMutvWBXsgbF3SKtYCyx` — Squads multi-sig that custodies the bridge reserve

### 2. `registries/sumTokens.js` — removed duplicate entry
Deleted the `"cookiechain-bridge"` block. It was the **original** adapter (added in #19640, migrated into the registry by #19742) that summed only the `DoYYC` multi-sig. Because it shared the `cookiechain-bridge` key with the newer `projects/` file, it shadowed it — so the live protocol was running the stale registry entry, not the intended adapter. Removing it makes `projects/cookiechain-bridge/index.js` the single source of truth.

## Methodology & rationale

**TVL = COOK locked in the bridge on Solana** (Hyperlane warp escrow PDA + Squads multi-sig reserve).

- **Why Solana-only:** COOK bridges 1:1 and is the **same asset** on both chains. Counting the mirrored Cookie Chain balances as well would **double-count** the same COOK, so TVL is measured once, on the canonical Solana side.
- **Continuity:** this extends the bridge's original methodology (*"COOK locked in the Solana bridge"*, which counted the `DoYYC` multi-sig) by adding the **Hyperlane escrow PDA** introduced during the migration. It is a clean continuation of the existing series (~415.7M → ~450M), not a redefinition — no TVL cliff.

## Not counted (intentionally)
- **Cookie Chain side** (`CL2Jo…` Hyperlane PDA, `G3mm…` Squads multi-sig) — mirror of the same 1:1 COOK; would double-count.

## Note for maintainers
The `cookiechain-bridge` protocol is currently flagged **`deprecated`** on the server. That request was intended for the project's *old community bridge* (a separate wallet, not a DefiLlama protocol) — but there is only one `cookiechain-bridge` protocol, so the flag is hiding the **active** Hyperlane bridge. Please **remove the `deprecated` flag** so the live bridge shows.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated Cookiechain Bridge TVL reporting to measure Solana-side balances only.
  * Improved balance accounting to include bridged escrow and custody holdings without double-counting.
  * Removed the outdated Cookie Chain TVL view from available reporting.

### DIFF
--- a/projects/cookiechain-bridge/index.js
+++ b/projects/cookiechain-bridge/index.js
@@ -1,36 +1,31 @@
-const { PublicKey } = require('@solana/web3.js')
-const { getConnection, sumTokens2 } = require('../helper/solana')
+const { sumTokens2 } = require('../helper/solana')
 
-// Cookie Bridge is a Hyperlane warp route between Solana and Cookie Chain (a
-// Solana / Agave fork). It is collateral <-> collateral: SPL COOK is locked in
-// a warp escrow on Solana, native COOK is locked in a warp collateral PDA on
-// Cookie Chain, and the relayer releases 1:1 from the destination pool. Nothing
-// synthetic is minted, so summing both locked pools is the TVL with no
-// double-counting. Both chains are indexed and priced by DefiLlama (COOK,
-// confidence ~0.99), so each pool is reported under its own chain.
+// Cookie Bridge moves COOK between Solana and Cookie Chain (a Solana / Agave fork)
+// 1:1 — it is the same asset on both chains. TVL is measured on the canonical
+// Solana side only (counting the mirrored Cookie Chain balance too would
+// double-count the same COOK). This continues the bridge's original methodology
+// ("COOK locked in the Solana bridge"), which counted the Squads multi-sig; with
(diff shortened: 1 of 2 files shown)


RohanNero:
Thank you for the PR, we will remove the deprecated flag on the server

PR #20222, opened 2026-07-25
wealthville: value CLMM/DLMM/perp positions on-chain (fix TVL undercount)

amitesh-m:
Updates the existing `wealthville` adapter. The current version only sums idle SPL/SOL balances held directly in the vault PDAs (`sumTokens2`), missing the bulk of TVL — funds deployed into concentrated-liquidity and perp positions (~85% undercount: ~$1.7k reported vs ~$11.8k actual NAV).

It now reads the full on-chain NAV of every smart-vault PDA, all from mainnet state (no project API, independently verifiable):

- **Idle** SPL + native SOL, incl. JitoSOL — `sumTokens2`
- **Orca Whirlpool** + **Raydium CLMM** — unwrapped from position NFTs via the shared `addUniV3LikePosition` math
- **Meteora DLMM** — unwrapped from on-chain bin reserves
- **Jupiter Perpetuals** — idle balances in the per-vault `jupiter_owner` PDA + the open position's on-chain `collateralUsd`

Verified against the protocol's own on-chain NAV across all active vaults within ~0.2%. `eslint` passes.

## Summary by CodeRabbit

* **New Features**
  * Expanded WealthVille TVL from idle-balance counting to a full on-chain valuation pipeline across smart vault positions.
  * Added valuation for Orca Whirlpool and Raydium CLMM positions (unwrapped from position tokens).
  * Added valuation for Meteora DLMM liquidity positions.
  * Added Jupiter Perpetuals collateral (USD) held within positions to total value.
* **Metadata**
  * Updated the methodology description to reflect the new multi-venue valuation approach.
  * Disabled historical time-travel TVL reporting.

### DIFF
--- a/projects/wealthville/index.js
+++ b/projects/wealthville/index.js
@@ -1,5 +1,5 @@
 const { PublicKey } = require("@solana/web3.js");
-const { getConnection, sumTokens2, runInChunks, decodeAccount, TOKEN_PROGRAM_ID, TOKEN_2022_PROGRAM_ID } = require("../helper/solana");
+const { getConnection, sumTokens2, getMultipleAccounts, decodeAccount, TOKEN_PROGRAM_ID, TOKEN_2022_PROGRAM_ID } = require("../helper/solana");
 const { addUniV3LikePosition } = require("../helper/unwrapLPs");
 const { bs58 } = require("@project-serum/anchor/dist/cjs/utils/bytes");
 
@@ -34,19 +34,12 @@ const BINS_PER_ARRAY = 70;
 const DLMM_POSITION_V2_DISC = bs58.encode(Buffer.from([117, 176, 212, 199, 245, 180, 133, 182]));
 const JUP_POSITION_DISC = bs58.encode(Buffer.from([170, 188, 143, 228, 122, 64, 247, 208]));
 
-// One RPC call at a time with a pause in between. The shared Solana endpoint this repo falls
-// back to throttles per method (getTokenAccountsByOwner and getProgramAccounts in particular),
-// so fanning these out with Promise.all trips its rate limit; account reads still go out
-// batched via getMultipleAccountsInfo below.
-const serially = (items, fn) => runInChunks(items, (c) => Promise.all(c.map(fn)), { chunkSize: 1, sleepTime: 1200 });
-const getMulti = (conn, keys) => runInChunks(keys, (c) => conn.getMultipleAccountsInfo(c), { sleepTime: 200 });
-
 // A single pass over the owners' SPL accounts yields both things we need from them: the
 // position-NFT receipts (balance exactly 1) that key the CLMM positions, and the funded token
 // accounts carrying idle balances. Scanning once avoids re-reading the same accounts per use.
 async function scanTokenAccounts(conn, owners) {
   const queries = owners.flatMap((owner, ownerIndex) => [TOKEN_PROGRAM_ID, TOKEN_2022_PROGRAM_ID].map((programId) => ({ owner, ownerIndex, programId })));
-  const results = await serially(queries, (q) => conn.getTokenAccountsByOwner(q.owner, { programId: q.programId }));
+  const results = await Promise.all(queries.map((q) => conn.getTokenAccountsByOwner(q.owner, { programId: q.programId })));
 
   const nftMints = owners.map(() => []);
   const tokenAccounts = [];
@@ -73,7 +66,7 @@ async function addClmmPositions(api, conn, nftMintsByOwner) {
     }
   }
   if (!cands.length) return;
-  const infos = await getMulti(conn, cands.map((c) => c.key));
+  const infos = await getMultipleAccounts(cands.map((c) => c.key));
 
   const poolSet = new Set();
   const positions = [];
@@ -96,7 +89,7 @@ async function addClmmPositions(api, conn, nftMintsByOwner) {
   if (!positions.length) return;
 
   const poolKeys = [...poolSet];
-  const poolInfos = await getMulti(conn, poolKeys.map((k) => new PublicKey(k)));
+  const poolInfos = await getMultipleAccounts(poolKeys.map((k) => new PublicKey(k)));
   const pools = {};
   poolKeys.forEach((k, i) => {
     const info = poolInfos[i];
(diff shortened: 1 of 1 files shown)


amitesh-m:
Pushed a fix for the red CI — it was an RPC issue rather than the valuation logic, and it's worth writing up because it isn't specific to this adapter.

**What CI hit**

```
Request failed with status code 429
Error pulling TVL for solana: Failed to post https://api.mainnet-beta.solana.com
    at async tvl (projects/wealthville/index.js:193:3)
Error: Protocol doesn't have total tvl
```

Line 193 was `sumTokens2({ api, owners, solOwners })`. That helper's `getOwnerAllAccounts` path packs every owner into one JSON-RPC array batch, and the public endpoint `test.yml` falls back to (no `SOLANA_RPC` is set in CI) rejects those outright. Probing it directly:

| batch size | result |
|---|---|
| 52 | 429 |
| 26 | 429 |
| 10 | 429 |
| 4 | 429 |
| 2 | 429 |
| 1 | OK |

Any batch of 2+ `getTokenAccountsByOwner` calls is refused. Sent one at a time it also throttles per method — ~12 calls before a hard 429 that survives web3.js's backoff.

Worth flagging: **this reproduces on `master`.** `git checkout main -- projects/wealthville/index.js && node test.js projects/wealthville/index.js` fails with the same error, because the merged adapter calls `sumTokens2({ owners })` too. This PR widened the owner set (adding the `jupiter_owner` PDAs), so it tripped the limit harder, but the adapter on master can't run against the public endpoint either.

**What changed**

Rather than lean on a better RPC being present, the adapter now stays inside the public endpoint's budget:

- **One scan per owner, reused.** Each owner's SPL accounts are read once, serially, and that single pass yields both the position-NFT receipts (balance exactly 1) and the funded token accounts. Those token accounts go to `sumTokens2({ tokenAccounts })`, which resolves them through `getMultipleAccountsInfo` — batched, and not rate-limited. Previously the same accounts were read twice: once by `sumTokens2({ owners })` and again for NFT discovery.
- **Skip uninitialised `jupiter_owner` PDAs.** They're created and funded on a vault's first perp position, so an uninitialised one has nothing to value. Currently 4 of 13 exist; I verified the other 9 have 0 token accounts and 0 perp positions, so this drops 18 RPC calls without dropping any value.
- **Derive BinArray PDAs instead of scanning for them.** A DLMM position spans at most two `["bin_array", lbPair, i64 index]` PDAs, so they're derived and every pair + array is fetched in one batched read. This replaces a `getProgramAccounts` per position, which is the scalability point @coderabbitai raised — DLMM and Jupiter discovery are now serial-with-backoff rather than unbounded fan-out.
- **Use the repo's own layout decoders** — `meteoraPosition` / `meteoraLbPair` / `meteoraBinArray` and `raydiumPositionInfo` / `raydiumCLMM` — instead of the hand-rolled byte offsets. Only the Orca Whirlpool reads still use documented offsets, as there's no Whirlpool layout in `helper/utils/solana/layouts` yet. Happy to add one if you'd like it.

**Result** — green against the exact endpoint CI uses, no `SOLANA_RPC` set:

```
$ node test.js projects/wealthville/index.js
--- tvl ---
SOL         3.86 k
usdc        3.74 k
JITOSOL     2.03 k
USDT        1.55 k
HYPE        300.00
CBBTC       221.00
jup         165.00
CARDS       65.00
usdt        2.00
Total: 11.93 k

------ TVL ------
solana      11.94 k

real  1m13s
```

`npx eslint projects/wealthville/index.js` is clean. The run takes ~73s because of the deliberate pacing; with a real RPC configured it's far quicker, and it degrades gracefully as vaults are added since the expensive per-position scans are gone.

For context on the number: the adapter on master only sums idle balances (~$1.7k) and misses the Orca/Raydium CLMM, Meteora DLMM and Jupiter-perp positions holding most of the NAV. $11.93k is the full on-chain value, and it lines up with the per-vault `tvlUsd` in the companion yields adapter (DefiLlama/yield-server#2714), which reads the same state.

amitesh-m:
Gentle bump with a concrete number, since the impact is measurable rather than cosmetic.

`api.llama.fi/protocol/wealthville` currently reports **$1,832.71**. Actual on-chain NAV across the vaults today is **$12,090** — so the public figure understates the protocol by about **85%**.

The gap is exactly what this PR fixes: the merged adapter sums vault-PDA token accounts only, which catches idle balances but misses the Orca/Raydium CLMM, Meteora DLMM, JitoSOL and Jupiter-perp positions that hold most of the value. This PR reads the full on-chain NAV of every vault PDA, and CI here independently computes `solana 11.92k`, matching our own read to within price drift.

No action needed from your side beyond a review — `test` and `CodeRabbit` are both green and the branch is mergeable. Happy to rebase or split it if that helps it move.

amitesh-m:
Two notes, then I'll stop bumping this.

**On the CodeRabbit nitpick (sequential per-vault `getProgramAccounts`) — I'm going to decline it, with a reason.**

The suggestion is to either drop the per-owner memcmp filter for a single discriminator-only scan per program, or batch the per-vault calls through `Promise.all`. Both would break CI here, and the sequential shape is deliberate rather than an oversight.

This repo's CI (`.github/workflows/test.yml`) sets no `SOLANA_RPC` secret, so `getConnection()` falls back to `https://api.mainnet-beta.solana.com`. Measured against that endpoint:

- **Any JSON-RPC array batch of 2 or more calls is refused with 429.** Batch sizes 52 / 26 / 10 / 4 / 2 all 429'd; only size 1 succeeded. So `Promise.all` over vaults is precisely the shape that fails.
- `getProgramAccounts` **with** memcmp filters is fine when paced — 52 sequential calls at ~1.2s spacing ran clean.

A discriminator-only scan would also return every DLMM position account on mainnet rather than the handful belonging to these vaults, which is a much larger payload on the endpoint least able to serve it.

That is what the second commit ("make on-chain position valuation survive the shared Solana RPC") is doing. `node test.js projects/wealthville/index.js` with no `.env` takes ~73s green, and the pacing is the reason.

Happy to reconsider if you'd rather I bound concurrency at 2-3 on the assumption of a paid RPC, but the current shape is the one that passes CI as configured.

**On priority:** [DefiLlama/yield-server#2714](https://github.com/DefiLlama/yield-server/pull/2714) is now explicitly waiting on this one — @0xkr3p asked to "wait for the TVL Adapter fix to land then will re-review". Flagging only so the dependency is visible; no rush implied.

RohanNero:
Apologies for the delay, and thank you for the PR. As you noted, those tests are using the public solana rpc so they occasionally fail despite this not being an issue in prod

PR #20249, opened 2026-07-27
fix(aave-v3): robust Mantle TVL via getReserveData available liquidity

0xaicrypto:
## Summary

The reviewer correctly pointed out that lending TVL should be **available liquidity** (`supplied - borrowed`), not gross deposits. The default Aave helper derives this from `underlying.balanceOf(aTokenAddress)`, which works for most assets but **reverts for USDT0 on Mantle**.

This PR replaces the Mantle `tvl` calculation with a direct call to the protocol data provider's `getReserveData()` and uses:

```
availableLiquidity = totalAToken - totalStableDebt - totalVariableDebt
```

for every reserve. The existing Ethena blacklist lender subtraction is preserved.

## Why this matters

* **USDT0** (`0x779D...736`) reverts on `balanceOf(...)`. The default path therefore cannot read its available liquidity, and the only USDT0 balance that shows up in TVL is the negative Ethena blacklist adjustment.
* Using `getReserveData()` makes the adapter robust to tokens with non-standard `balanceOf` behavior and aligns with DefiLlama's lending methodology.

## Numbers

| Metric | Before (DefiLlama) | After (this PR) |
|---|---:|---:|
| Mantle TVL | ~$47.3M | ~$46.9M |
| Mantle Borrowed | ~$141.2M | ~$141.3M |

The headline TVL is essentially unchanged; this is a correctness/robustness fix rather than a numeric increase.

## Verification

```bash
node test.js projects/aave-v3/index.js
```

Result:
```
--- mantle ---
Total: 46.86 M
--- mantle-borrowed ---
Total: 141.28 M
```

No other chain is affected.

## Files changed

* `projects/aave-v3/index.js`: added a Mantle-specific `tvl` function that computes available liquidity from `getReserveData()` and keeps the Ethena blacklist subtraction.

### DIFF
--- a/projects/aave-v3/index.js
+++ b/projects/aave-v3/index.js
@@ -59,7 +59,63 @@ const CONFIG = {
   monad: ['0xB65A68B98274ef7D9a60E0C0747dD1BEc3D32fad']
 };
 
-module.exports = aaveV3Export(CONFIG)
+const aaveExports = aaveV3Export(CONFIG)
+
+// Mantle-specific supply calculation.
+// The default helper uses underlying.balanceOf(aTokenAddress), which under-reports
+// (or reverts for) several Mantle assets. For a correct deposit figure we read each
+// aToken's totalSupply() directly from the protocol data provider.
+const MANTLE_POOL_DATA = '0x487c5c669D9eee6057C44973207101276cf73b68'
+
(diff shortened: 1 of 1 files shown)


g1nt0ki (changes requested):
incorrect implementation, lending product tvl = supplied - borrowed

PR #20252, opened 2026-07-28
accountable: add aHYPER Looping Vault

pawel-accountable:
Adds the aHYPER Looping Vault (`0x23b148d8f389C5821739381f1FF87bB7e1162566`, Monad) to the Accountable adapter. It is deployed directly, not via a registered factory, so it isn't discoverable through `strategyProxies`/`strategyVaults` — added explicitly. TVL uses the same path as other vaults; verified +$7.55M on the test harness.

## Summary by CodeRabbit

* **New Features**
  * Added support for recognizing additional vault addresses during vault discovery on the Monad chain.
  * Included vaults discovered through these additional addresses in TVL calculations.

### DIFF
--- a/projects/accountable/index.js
+++ b/projects/accountable/index.js
@@ -21,6 +21,12 @@ const FACTORIES = {
     ],
 }
 
+const EXTRA_VAULTS = {
+    monad: [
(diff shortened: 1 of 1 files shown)


g1nt0ki (approved):
thanks for the PR

PR #20256, opened 2026-07-28
Add Textile (on-chain FX DEX) TVL adapter

tomer-ba-text:
Adds a TVL adapter for **Textile**, an on-chain FX liquidity network (FX DEX).

**Site:** https://app.textilecredit.com · **Docs:** https://docs.textilecredit.com

**Methodology:** Textile is non-custodial. Market makers ("Stitch" bots) quote both sides of each FX corridor from their own wallets, and swaps settle atomically through a stateless reactor, so no funds are custodied in protocol contracts. TVL is the corridor-token inventory those makers hold on each chain. The adapter reads the maker set and the corridor tokens from the Textile subgraph, then sums balances on-chain per chain, so it self-updates as makers and corridors change.

**Chains:** Ethereum, BSC, Base, Celo (per-chain, as DefiLlama expects).

Corridors include cNGN, wARS, wBRL (local-currency stablecoins), XAUt (gold), and WETH, quoted against USDT/USDC.

## Summary by CodeRabbit

* **New Features**
  * Added TVL tracking for Textile across Ethereum, BNB Chain, Base, and Celo.
  * Added support for calculating TVL from supported collateral and debt tokens.
  * Added methodology information describing Textile’s TVL calculation.

### DIFF
--- a/projects/textile/index.js
+++ b/projects/textile/index.js
@@ -1,0 +1,45 @@
+const { graphQuery } = require("../helper/http");
+
+// Textile is a non-custodial on-chain FX DEX. Market makers ("Stitch" bots) quote
+// both sides of each corridor from their own wallets, and swaps settle atomically
+// through a stateless reactor, so nothing is custodied in protocol contracts. TVL is
+// the inventory those makers hold in the corridor tokens on each chain. The maker set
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR

PR #20263, opened 2026-07-28
Add SUBFROST

Vdto88:
**SUBFROST** is a decentralized custodian on Bitcoin L1. It issues **frBTC**, a synthetic bitcoin pegged 1:1 to BTC, against BTC held by a FROST signer set (6-of-9 threshold Schnorr).

- Website: https://subfrost.io
- App: https://app.subfrost.io
- Chain: Bitcoin
- Category: Bridge

## Methodology

TVL is the BTC held in the SUBFROST custody addresses that back frBTC, read straight from the Bitcoin chain via `sumTokens`.

frBTC is issued on **two** Bitcoin metaprotocols, Alkanes and BRC2.0, each with its own custody address; both are included. The frBTC minted against those deposits is **not** counted separately, so nothing is double counted.

## On the two addresses

Neither is an arbitrary constant — each is the taproot output key of the signer set. The Alkanes one is re-derivable from Bitcoin L1 at any time: the frBTC alkane `[32:0]` returns the signer's 32-byte internal pubkey from opcode 103 (`GET_SIGNER`), and the address is the standard BIP341 tweak of that key with no script tree. The exact RPC call is in the code comment.

It is pinned in `bitcoin-book` rather than derived at runtime because this repo carries no bitcoin/bech32/secp256k1 dependency to perform the tweak with. Happy to change the approach if you'd prefer it otherwise.

## Testing

```
$ node test.js projects/subfrost.js

--- bitcoin ---
BTC                       6.83 M
Total: 6.83 M

------ TVL ------
bitcoin                   6.83 M
total                     6.83 M
```

`check-bitcoin-duplicates` reports no conflict with an existing entry. Category follows the existing precedent for wrapped-BTC issuers on `chains:["Bitcoin"]` (WBTC, FBTC, sBTC are all Bridge).

## Summary by CodeRabbit

* **New Features**
  * Added Subfrost Bitcoin TVL tracking.
  * TVL now reflects on-chain Bitcoin balances held in Subfrost custody addresses for both frBTC deployments.
  * Added support for two Subfrost Bitcoin custody addresses used to compute TVL.

### DIFF
--- a/projects/subfrost/index.js
+++ b/projects/subfrost/index.js
@@ -1,5 +1,5 @@
-const { sumTokens } = require("./helper/sumTokens");
-const bitcoinAddressBook = require("./helper/bitcoin-book/index.js");
+const { sumTokens } = require("../helper/sumTokens");
+const bitcoinAddressBook = require("../helper/bitcoin-book/index.js");
 
 // SUBFROST is a decentralized custodian on Bitcoin L1. It issues frBTC, a synthetic
 // bitcoin pegged 1:1 to BTC, against BTC held by a FROST signer set. frBTC is issued
--- a/projects/helper/bitcoin-book/index.js
+++ b/projects/helper/bitcoin-book/index.js
@@ -18,20 +18,6 @@ const bitomato = ["bc1qgmtx3caf8rlxmzw703ga2sljv3rkkj39e4ysk9"];
 
 const lbank = ["1MZwhQkkt9wy8Mwm4rx5W3AYiDCJLasffn"];
 
-// SUBFROST frBTC custody. Neither address is an arbitrary constant: each is the
-// taproot output key of the FROST signer set that holds the BTC backing frBTC.
-// The Alkanes one can be re-derived from Bitcoin L1 at any time: the frBTC alkane
-// [32:0] returns the signer's 32-byte internal pubkey from opcode 103 (GET_SIGNER),
-// and the address is the standard BIP341 tweak of that key (no script tree).
-// Any metashrew/alkanes node can answer it, e.g. https://mainnet.subfrost.io/v4/subfrost
-//   {"method":"alkanes_simulate","params":[{"target":{"block":"32","tx":"0"},
-//    "inputs":["103"],"alkanes":[],"transaction":"0x","block":"0x","height":<tip>,
-//    "txindex":0,"vout":0}]}
-const subfrost = [
-  "bc1p5lushqjk7kxpqa87ppwn0dealucyqa6t40ppdkhpqm3grcpqvw9s3wdsx7", // Alkanes
-  "bc1pxn3gr0hy70exhdqjzawtuygppzdrk3mer3wlaa2gzkmruk3rrt4qga2qaj", // BRC2.0
-];
-
 const stacksSBTC = [
   // https://docs.stacks.co/concepts/sbtc/clarity-contracts/sbtc-deposit
   "bc1pl033nz4lj7u7wz3l2k2ew3f7af4sdja8r25ernl00thflwempayswr5hvc",
@@ -1175,12 +1161,18 @@ module.exports = {
     "3AHghpZ5GAU7rjTXHv4Xmfe6BLavJxnzbo",
     "3Pr9uMzcEtmmCLShywSrsHq6Xqy9taEdXh",
   ],
-  subfrost,
   circleBTC: [
     '1JkKmG26nUBcPS99TsCVsReSXvLkEai4ca',
     '1KVBNjpYfJvASdzeTAwqNbe9WecpKyugM3',
     '1HkJ6hcN4h4PtUYHiSi1hrUEUKQJmedM6z',
     '1FXxhAa9yKCG8WgCTrbSsdGKuC6QzN3Gq9',
+  ],
+  subfrost = [
+    // The FROST signer set's taproot output keys that hold the BTC backing frBTC.
+    // Alkanes custody derived from Bitcoin L1 by reading [32:0] opcode 103 (GET_SIGNER) on any metashrew/alkanes node
+    // for the signer's 32-byte internal pubkey, with standard BIP341-tweak (no script tree).
+    "bc1p5lushqjk7kxpqa87ppwn0dealucyqa6t40ppdkhpqm3grcpqvw9s3wdsx7", // Alkanes
+    "bc1pxn3gr0hy70exhdqjzawtuygppzdrk3mer3wlaa2gzkmruk3rrt4qga2qaj", // BRC2.0
   ],
 };
 
--- a/projects/helper/bitcoin-book/index.js
+++ b/projects/helper/bitcoin-book/index.js
@@ -1167,7 +1167,7 @@ module.exports = {
     '1HkJ6hcN4h4PtUYHiSi1hrUEUKQJmedM6z',
     '1FXxhAa9yKCG8WgCTrbSsdGKuC6QzN3Gq9',
   ],
-  subfrost = [
+  subfrost: [
     // The FROST signer set's taproot output keys that hold the BTC backing frBTC.
     // Alkanes custody derived from Bitcoin L1 by reading [32:0] opcode 103 (GET_SIGNER) on any metashrew/alkanes node
     // for the signer's 32-byte internal pubkey, with standard BIP341-tweak (no script tree).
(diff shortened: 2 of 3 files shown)


projects/subfrost/index.js
```diff
@@ -0,0 +1,16 @@
...
+const owners = bitcoinAddressBook.subfrost;
+
+async function tvl(api) {
+  return sumTokens({ owners, api });
+}
+
+module.exports = {
+  bitcoin: { tvl },
+  methodology: `TVL is the BTC held in the SUBFROST signer custody addresses that back frBTC, read directly from the Bitcoin chain. Covers both frBTC deployments: Alkanes and BRC2.0. frBTC minted against these deposits is not counted separately, so nothing is double counted.`,
+};
```
RohanNero, line 16: Could you please rename/move this into `projects/subfrost/index.js`

RohanNero (changes requested):

RohanNero (approved):

RohanNero:
Thank you for the PR

Vdto88:
Hey team — following up on the pre-merge check note that this PR was missing the new-listing template fields. Here they are, so the protocol entry can be created:

**Name (to be shown on DefiLlama):** SUBFROST

**Twitter Link:** https://x.com/subfrost_news

**List of audit links if any:** — (the end-to-end SUBFROST protocol has not been audited yet; a Halborn engagement begins Sep 1, 2026. The underlying FROST-on-taproot primitive was reviewed by the Zcash team. We'd rather leave the audits field empty than overstate this — details are in the methodology.)

**Website Link:** https://subfrost.io

**Logo (High resolution, will be shown with rounded borders):** https://subfrost.io/subfrost-logomark.svg (SVG, square 320×320; happy to provide a raster PNG if preferred)

**Current TVL:** ~$6.8M (per this adapter's CI run)

**Treasury Addresses (if the protocol has treasury):** —

**Chain:** Bitcoin

**Coingecko ID (leave empty if no token):** —

**Coinmarketcap ID (leave empty if no token):** —

**Short Description (to be shown on DefiLlama):** SUBFROST wraps BTC into frBTC 1:1 on Bitcoin metaprotocols (Alkanes and BRC2.0), with mint/redeem executed by a 6-of-9 FROST threshold signer group.

**Token address and ticker if any:** —

**Category (full list at https://defillama.com/categories):** Bridge

**Oracle Provider(s):** —

**forkedFrom (does your project originate from another project):** —

**methodology (what is being counted as tvl, how is tvl being calculated):** BTC held at the two documented SUBFROST custody addresses backing frBTC issuance across both metaprotocol deployments (Alkanes and BRC2.0), summed via sumTokens. Custody has consistently held a reserve above outstanding frBTC supply. Protocol fees: 0.1% on wrap/unwrap; swaps on the native AMM charge 1% (0.8% LP / 0.2% protocol).

**Github org/user (Optional, if your project is open source):** https://github.com/subfrost

Vdto88:
Thanks for creating the entry — I can see it on the API now (`/protocol/subfrost`, id 8317, category Bridge). Two small things:

**1. Twitter handle.** The entry currently has `twitter: "gabe_subfrost"`, which is a team member's personal account. The protocol account is **@subfrost_news** — that's where the project publishes (e.g. https://x.com/subfrost_news/status/2071636563356692834). Could you switch it?

**2. TVL is still empty** (`tvl: []`, `currentChainTvls: {}`) and the public page doesn't render yet. The entry is only a couple of hours old so this is probably just the first-run delay — no rush at all. Flagging only in case something is needed on our side to link the entry to the adapter: it lives at `projects/subfrost/index.js` and the CI check on this PR reported `bitcoin 6.83 M`.

Everything else in the entry looks right. Thanks again.

Vdto88:
Following up now that the page is live — thanks, TVL populated and `defillama.com/protocol/subfrost` reads $6.87m.

One correction still outstanding: the entry's twitter is `gabe_subfrost`, which is a team member's personal account. The protocol account is **@subfrost_news**.

Adding the detail I did not have earlier, in case it helps prioritise it: the older `@subfrost_io` handle was **suspended by X**, so `subfrost_news` is not a preference between two accounts — it is the only live one. Anyone who follows the link on the entry today lands on a personal profile instead of the protocol.

Nothing else in the entry needs changing.

Vdto88:
Different topic from the SUBFROST listing above, and a question rather than a request, because the answer decides whether any code is worth writing.

**Would you consider Alkanes as a chain entry under Bitcoin?**

Alkanes is the metaprotocol SUBFROST is built on: a WASM smart contract environment whose state lives in Bitcoin transactions and is derived by indexing them. It has no consensus of its own, no validator set, no native token, and no gas token. Execution is paid for in BTC, in ordinary Bitcoin fees.

I checked what exists today before asking, so the question is narrow:

- Of the 461 entries in `/v2/chains`, the Bitcoin-adjacent ones are Bitcoin, Rootstock, Stacks, Merlin, Bitcoincash and Botanix. All of them are sidechains or L2s with their own consensus and their own gas token (`RBTC`, `STX`, `MERL`).
- No Bitcoin metaprotocol is listed at all. No Runes, Ordinals, BRC-20 or Alkanes.
- `Rootstock` is the only entry with `parent: { chain: "Bitcoin", types: ["gas"] }`, which is the grouping that produces `defillama.com/chains/bitcoin`.

So I am not asking for a new category. I am asking whether Alkanes fits the slot that already exists, `parent.chain: "Bitcoin"` with `types: ["gas"]`, which is exactly Rootstock's shape: gas inherited from the parent chain.

Two things I think are not blockers, and I would rather you tell me if I am wrong:

1. **No native token.** `Base` carries `geckoId: null` and `symbol: null` with $4.55b, so a chain entry clearly does not require one. Alkanes would be the same, with `types: ["gas"]` pointing at BTC.
2. **Double counting.** SUBFROST's TVL is BTC held in L1 custody and belongs on Bitcoin, where it is today. It would not move. What has no home right now is TVL denominated in Alkanes tokens: AMM pool liquidity, the FIRE vault, orbitals. That is the content of an Alkanes entry, and it is disjoint from the BTC figure you already have.

The honest difference from Rootstock is that Alkanes has no consensus of its own, its state is a deterministic function of Bitcoin blocks. That is the actual question: does an execution environment that inherits both consensus and gas from Bitcoin count as a chain to you, or do you prefer that this TVL keeps arriving as separate protocols filed under the Bitcoin chain, which is what our listing already does?

Two dependencies I already know I would owe you if the answer is yes:

- **Pricing.** frBTC pegs to BTC, so that part is easy. DIESEL and FIRE have no CoinGecko id, so pool TVL would need a price source you accept, or the adapter values pools by their BTC side only.
- **State reads.** Your helpers speak EVM RPC and Alkanes state is not reachable that way. Reading it needs an Alkanes indexer (metashrew), so the adapter would either query one or derive balances itself. I read the rule about no project-specific npm packages, which is part of why I am asking before writing anything.

Happy to do the work either way. I just do not want to build an adapter against a chain key that will not exist.

Vdto88:
Correction to my own comment above, plus the concrete shape I think this should take. I measured rather than assumed, and one claim I made does not hold.

**What I got wrong.** I said Alkanes TVL would be disjoint from the Bitcoin figure you already have. That is true of only half of it. The DIESEL/frBTC pool (`2:77087`) holds 99.26 frBTC against 121,361.63 DIESEL, while total frBTC custody is 107.85 BTC, so 92% of all issued frBTC sits in that single pool. Its frBTC leg is the same BTC your Bitcoin entry already counts through SUBFROST, about $6.34m. Counting it under an Alkanes entry would be double counting, and you would be right to reject that.

Both legs of an AMM pool carry equal value by construction, so the DIESEL side is also ~$6.34m, and that part exists nowhere on DefiLlama today. The FIRE vault (`2:77628`) holds 45.74 LP of that same pool, 1.72% of supply, so it is a claim on the pool rather than new value. frUSD (`4:1776`) is deployed but its supply is still zero. So the honestly uncounted figure today is one pool leg, not a portfolio.

**What I think follows.** Two separate things, and only one of them needs a chain:

1. Regardless of the chain question, the right structure for us looks like `parentProtocol`: SUBFROST as the parent, with the bridge, the AMM and the FIRE vault as children, and the overlap declared through `excludeTvlFromParent` rather than hidden. That keeps FIRE and, when it has supply, frUSD visible as products without inflating any total. If you agree, I will open that as an adapter PR and mail the metadata as support suggested.
2. The chain entry is a weaker ask on today's numbers and I would rather say so than oversell it: one pool, and DIESEL has no CoinGecko id, so the only price available is the ratio of the pool being measured, which is circular. If you have an accepted pattern for that, I will follow it.

The reason I still think the chain is worth considering is not our TVL. It is that Alkanes has other protocols that currently have nowhere to file, and no Bitcoin metaprotocol is represented on DefiLlama at all today. A chain key under Bitcoin would give them a destination instead of each one arguing its own case. If you would rather see that demand demonstrated first, say so and I will come back when more than one protocol is asking.

For completeness on the mail thread: support confirmed the twitter handle fix, and said chain metadata only matters once changes are accepted in an adapter PR here, which is why I am keeping this in the repo rather than by mail.

PR #20271, opened 2026-07-29
term-finance-vaults: clarify methodology

robotoer:
Went with keeping the vaults adapter liquid-only, so this is now just a methodology string change, no behaviour change.

The old string ("deposits allocated to other protocols") didn't describe what `totalLiquidBalance()` actually returns, and didn't say why deployed capital is left out. New string covers both.

TVL is unchanged:

```
------ TVL ------
ethereum                  6.64 M
bsc                       2.09 M
avax                      1.54 M
plasma                    7.92 k
arbitrum                  1.01 k
base                      0.00

total                    10.28 M
```

## Summary by CodeRabbit

* **Documentation**
  * Clarified how Term Strategy Vault balances are calculated: liquid balance now reflects idle assets plus reserves held in third-party ERC4626 vaults.
  * Explicitly notes that capital deployed into Term repo tokens is excluded to prevent double counting.

### DIFF
--- a/projects/term-finance-vaults/index.js
+++ b/projects/term-finance-vaults/index.js
@@ -58,7 +58,7 @@ const vaultsGraphStartBlock = {
 }
 
 module.exports = {
-  methodology: `Counts deposits in Term Strategy Vaults allocated to other protocols.`,
+  methodology: `Counts the liquid balance of Term Strategy Vaults: idle assets plus reserves parked in third party ERC4626 vaults. Capital deployed into Term repo tokens is left out here to avoid double counting, it is tracked by the term-finance adapter as locked purchase tokens and borrowed.`,
   // hallmarks: [['2020-05-04', "TermFinance Launch"]],
 };
 


projects/term-finance-vaults/index.js
```diff
@@ -58,7 +58,8 @@ const vaultsGraphStartBlock = {
 }
 
 module.exports = {
-  methodology: `Counts deposits in Term Strategy Vaults allocated to other protocols.`,
+  methodology: `Counts total deposits in Term Strategy Vaults via each strategy's totalAssetValue(): idle liquidity plus capital deployed into Term repo tokens and pending auction offers.`,
```
RohanNero: If using `totalAssetValue` includes tokens tracked by `projects/term-finance/index.js`, we can either continue tracking only the idle vault funds here, or we can exclude the `term-finance/index.js` tvl from the parent: https://defillama.com/protocol/termfinance. Do you have a preference on which? They would both result in the same ~13M currently

RohanNero (changes requested):

projects/term-finance-vaults/index.js
robotoer: Good catch, thanks. Went with the first option, keeping this one liquid-only.

The two aren't quite equivalent though. Collateral never shows up in `totalAssetValue` (the vault holds the repo token, the borrower's wstETH/cbBTC sits in the repo locker), so excluding `term-finance` from the parent would drop ~2M of collateral while counting the lent out portion as tvl instead of borrowed. Lands on a similar number today by coincidence, but the two errors won't keep cancelling.

So this is now just a methodology string fix, no behaviour change. Deployed capital stays out of the vaults adapter and is tracked in `term-finance` as locked purchase tokens + borrowed.

RohanNero (approved):

RohanNero:
Thank you for PR and the update

PR #20280, opened 2026-07-29
fix(dreamdex): restore USDso quote + native SOMI legs dropped in consolidation

vincent-improbable:
`dreamdex` currently reports **$198,228** against **$1,643,861** actually held on-chain — 87.9% of TVL is uncounted. This is a regression from the `consolidate` commit d949732bf, not a change on our side.

### What happened

d949732bf migrated `projects/dreamdex/index.js` into `registries/sumTokens/data3.js`. It carried over the six `tokensAndOwners` ERC20 base legs and the `methodology` string verbatim, but dropped the adapter's `ownerTokens` block:

```js
for (const owner of owners) ownerTokens.push([[USDso, ADDRESSES.null], owner])
```

That block counted **USDso** — the quote asset of all four markets — and **native SOMI**, across all 8 owners (4 `SpotPool`s + their 4 `SpotStopOrderRegistry`s). Dropping it removed the entire quote side of the protocol, and removed the **SOMI/USDso market completely**: its base token is native SOMI, so both of that market's legs lived in the dropped block and neither its pool nor its registry appears anywhere in the migrated config.

The `methodology` text still shipping in the registry entry states that native SOMI and USDso are counted, so this reads as migration loss rather than an intended change.

### Verification

`node test.js projects/dreamdex/index.js`:

| | Total | Tokens counted |
|---|---|---|
| before | **198.23 k** | USDC.e, WBTC, WETH |
| after | **1.64 M** | + USDso 1.37 M, SOMI 72.5 k |

- The pre-fix run reproduces `api.llama.fi/protocol/dreamdex`'s live **$198,228** to the dollar, and that endpoint's token breakdown lists only USDC/WBTC/WETH.
- The same breakdown included `USDSO` ($934,464) and `SOMI` ($56,492) through the **2026-07-21 00:00 UTC** snapshot and neither from **2026-07-22 00:00 UTC** onward. Reported TVL steps from $1,604,591 to $467,130 across that boundary; d949732bf landed on `main` 2026-07-21 23:18 UTC.
- Direct reads against Somnia mainnet (chain id 5031) over the 8 owners: **1,373,411 USDso** and **838,345 native SOMI**, none of it currently counted. Priced with DefiLlama's own feed that is the $1,445,379 gap.
- Owners and token addresses reconcile against `config/mainnet/deploy.json` and `config/mainnet/stop-order-registries.json` in the protocol repo.

Real on-chain TVL over this window went ~$1.60M → ~$1.64M, i.e. flat. The reported collapse is entirely the dropped legs.

### The change

One line. Restored as `owners` × `tokens` alongside the existing `tokensAndOwners` rather than reintroducing `ownerTokens`, since `sumTokens2` folds all three into the same `tokensAndOwners` list and this form is the most compact. I verified all three spellings produce the same $1.64M.

### Two asks

1. **Backfill from 2026-07-22.** Merging only corrects forward; without a refill the chart keeps a false 88% collapse.
2. **`projects/component/index.js` broke the same way** and is not touched here. It was the only other adapter in d949732bf whose TVL relied on `ownerTokens` — 2 of 142 migrated adapters used that key and both regressed. Its `tvl` is now a bare `{}` on ethereum/xdai/bsc (not the `{ __empty: true }` marker the file uses for deliberately-empty TVL) and it reports $0 from the same 2026-07-22 snapshot, previously $5,375 on ethereum. Left for someone who owns that protocol, but worth a look — and possibly worth a guard in the consolidation tooling, since `ownerTokens` is supported by `sumTokens2` and was silently not emitted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved total value locked (TVL) calculation accuracy for Somnia by refining supported token and address mappings.


g1nt0ki:
@vincent-improbable thanks for the the fix, sorry for introducing the bug

PR #20289, opened 2026-07-29
originarm: value active market for all ARMs + add 2 new multi-asset ARMs

apexearth:
- Value the active lending-market position (activeMarket.convertToAssets) for every ARM generation, not just multi-base ones. Legacy single-base ARMs previously left their market position uncounted; activeMarket() is now batched up front and valued on both paths.
- Add the new multi-asset WETH ARM (0x6802..., bases stETH/wstETH/eETH/weETH) and USDC ARM (0x9e3a..., bases PYUSD/USDG), and guard the legacy branch so multi-asset ARMs (no single-base fallback) skip it.

## Summary by CodeRabbit

* **New Features**
  * Expanded Ethereum ARM support to handle multi-asset configurations alongside legacy single-base setups.
  * Improved total value reporting by valuing active market positions and the correct underlying asset denominations.
* **Bug Fixes**
  * Improved balance/denomination consistency across ARM generations, including correct conversion from market shares to underlying assets.
  * Added explicit failure behavior when base-asset data is missing for non-legacy ARM configurations.


RohanNero:
Thank you for the PR

PR #20303, opened 2026-07-30
Add Obligate (oTFY) adapter

dawidstezycki:
Adds a TVL adapter for **Obligate** — a regulated on-chain debt capital markets platform — starting with its first eTracker product, **oTFY** (Obligate Trade Finance Yield).

**What oTFY is:** an SPL security token on Solana issued by a Luxembourg securitization vehicle, representing a claim on the NAV of a revolving portfolio of investment-grade trade-finance bonds. Obligate AG (Zurich) is the product sponsor.
Mint: `BwB3tNH92jKw6naNGDYDbDwRo8bvYxZVvZjRZRcoWR2h`

**Chain:** Solana. Launched 2026-06-29.

**Methodology:** total oTFY supply is read on-chain from the Solana mint and multiplied by the latest per-token net asset value published by Obligate. NAV is USDC-denominated and treated 1:1 with USD. Supply is verifiable on-chain; the NAV feed is public and unauthenticated:
https://api.obligate.com/platform/etracker/token/3cef0789-88b8-4374-b818-73f9fed942af

**Why the NAV is applied inside the adapter:** oTFY is not currently priced by DefiLlama (`coins.llama.fi` returns no entry for `solana:BwB3tNH92jKw6naNGDYDbDwRo8bvYxZVvZjRZRcoWR2h`), so adding the raw mint supply would report $0. The adapter therefore applies the issuer's published NAV and reports the result under USDC's Solana mint (the NAV is USDC-denominated), with `misrepresentedTokens: true` set — the same pattern as `projects/anemoy-capital` and `projects/alcum`. Happy to switch to plain supply reporting if you'd rather add oTFY to the pricing pipeline instead.

**`timetravel: false`:** the NAV feed only exposes the current valuation, so historical values can't be backfilled.

Only the tokenized representation is counted — the underlying bond portfolio is not included, per the RWA edge case in the TVL docs.

## Summary by CodeRabbit

* **New Features**
  * Added TVL tracking for Obligate’s oTFY token on Solana using the latest published NAV and the token’s on-chain supply.
  * Enhanced valuation handling by validating the NAV and token details before computing TVL.

* **Bug Fixes**
  * Improved NAV request reliability with a stricter HTTP timeout.
  * Tightened NAV checks to require a finite value strictly greater than zero, and added clearer errors when the NAV currency is not USDC.

### DIFF
--- a/projects/obligate/index.js
+++ b/projects/obligate/index.js
@@ -1,0 +1,44 @@
+const ADDRESSES = require('../helper/coreAssets.json')
+const { getTokenSupplies } = require('../helper/solana')
+const { get } = require('../helper/http')
+
+// https://explorer.solana.com/address/BwB3tNH92jKw6naNGDYDbDwRo8bvYxZVvZjRZRcoWR2h
+const OTFY_MINT = 'BwB3tNH92jKw6naNGDYDbDwRo8bvYxZVvZjRZRcoWR2h'
+const OTFY_DECIMALS = 9
+
+// oTFY isn't priced by DefiLlama, so its NAV is applied here and reported as USDC.
+const USDC = ADDRESSES.solana.USDC
+const USDC_DECIMALS = 6
+
+// Public, unauthenticated NAV feed for the oTFY eTracker, keyed by its stable token id.
+const NAV_API =
+  'https://api.obligate.com/platform/etracker/token/3cef0789-88b8-4374-b818-73f9fed942af'
+
+async function tvl(api) {
+  const otfy = await get(NAV_API)
+  if (!otfy || otfy.symbol !== 'oTFY') throw new Error('Unexpected Obligate eTracker API response for oTFY')
+
+  // Fail loudly if the eTracker is reissued, rather than misreporting TVL.
+  if (otfy.contract !== OTFY_MINT)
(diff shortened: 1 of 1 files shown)


projects/obligate/index.js
```diff
@@ -0,0 +1,47 @@
+const ADDRESSES = require('../helper/coreAssets.json')
+const { getTokenSupplies } = require('../helper/solana')
+const { get } = require('../helper/http')
+
+// https://explorer.solana.com/address/BwB3tNH92jKw6naNGDYDbDwRo8bvYxZVvZjRZRcoWR2h
+const OTFY_MINT = 'BwB3tNH92jKw6naNGDYDbDwRo8bvYxZVvZjRZRcoWR2h'
```
RohanNero, line 6: Hello, thank you for the PR. If the collateral backing OTFY is on-chain, can you please track that instead? And apologies for the inconvenience, but we are changing how we list RWA tokens, if the backing is off-chain, could you please fill out one of our RWA submission forms:
    - main: https://forms.defillama.com/rwa-submission
    - backup: https://forms.gle/CgM2tPaFfpK2KK7p8

RohanNero (changes requested):

RohanNero:
Thanks again for the PR, we've listed oTFY here: https://defillama.com/rwa/asset/oTFY

PR #20153, opened 2026-07-21
feat(peridot): add Stellar Soroban markets, move out of compound registry

AsyncSan:
Peridot (already listed, id 7223) has a deployment on Stellar that this PR adds.

The Stellar markets are Soroban `ReceiptVault` contracts, not Compound `cToken` markets, so they cannot be expressed through `registries/compound.js` (a `stellar` key there would be passed to `compoundExports2`). This PR therefore moves the `peridot` key out of the compound registry into `projects/peridot/index.js`. The protocol's `module` in the DefiLlama config is already `peridot/index.js`, so no config change should be needed on your side.

The BSC and Monad config is carried over unchanged (same comptrollers, same `blacklistedMarkets` / `cether`), so their numbers are unaffected — verified below.

**Changed files**
- `projects/peridot/index.js` (new)
- `registries/compound.js` (`peridot` entry removed)

No dependency or lockfile changes.

### Stellar methodology

TVL is the underlying custodied by each of the three ReceiptVaults (XLM, USDC, EURC), read on-chain with `callSoroban(underlying, 'balance', [vault])` via `helper/chain/stellar` — the Soroban equivalent of Compound's `getCash()`. Outstanding debt is read from the vault's `get_total_borrowed()` and reported in the separate `borrowed` bucket, not in TVL.

Note: the vault also exposes `get_total_underlying()` (cash + borrows + accrued interest − reserves − fees). That is deliberately **not** used — it counts accrued interest that is not backed by assets in the vault, which would overstate TVL.

### Validation

`node test.js projects/peridot/index.js`

```
bsc                       324.00
bsc-borrowed              11.32 k
monad                     738.00
monad-borrowed            491.00
stellar                   264.00
stellar-borrowed          1.00

total                     1.32 k
```

BSC and Monad match the values currently reported on https://defillama.com/protocol/peridot (Binance 323.12, Monad 733.57), i.e. no regression from the registry → folder move.

Stellar per-market breakdown: USDC 117.19, EURC 124.83, XLM 20.02. TVL there is small today (~$264) because the deployment is new; the three vaults are live and in production use.

### Contracts (Stellar mainnet)

| Market | ReceiptVault | Underlying (SAC) |
| --- | --- | --- |
| XLM | `CBU4Y7CJFOUZZE3QBOXTKM54UTUYW3SDJWTNMDGJBNCR5HS5UCEKV3BE` | `ADDRESSES.stellar.XLM` |
[...]

PR #20326, opened 2026-07-31
fix(peridot): count Stellar underlying held in DeFindex boosted vaults

AsyncSan:
This is a fix to the existing `peridot` adapter (not a new listing), so the new-listing template fields are not applicable. No new dependencies, no lockfile changes.

### Problem

Peridot's Stellar markets are Soroban `ReceiptVault` contracts. Each vault forwards idle underlying into a DeFindex vault to earn additional yield, so the underlying a vault owns lives in two places:

1. the SAC balance held directly at the vault address, and
2. the underlying value of the DeFindex shares the vault holds.

The adapter added in #20153 computed TVL as `balance(underlying, vault)`, which only sees (1). Because the sweep into DeFindex is automatic, (1) is just the un-swept remainder — so roughly 90–99% of each market went untracked, and the gap widened with every new deposit.

### Fix

Read `get_available_liquidity()` from the vault, which is the vault's own accounting of unborrowed underlying it owns, i.e. (1) + (2). This is the Soroban equivalent of Compound's `getCash()` and is the same value Peridot's own frontend reports. BSC and Monad are untouched.

### Verification against mainnet

Underlying units, read live:

| asset | `balance(underlying, vault)` (before) | `get_available_liquidity()` (after) | tracked before |
|---|---|---|---|
| XLM  | 1,486.76 | 14,867.57 | 10.0% |
| USDC | 9.96     | 989.65    | 1.0%  |
| EURC | 102.23   | 1,022.35  | 10.0% |

The new figures reconcile against each vault's independent `get_total_underlying()` accessor, and against the DeFindex vaults directly — e.g. for XLM the ReceiptVault holds 133,781,200,653 shares of `CCB2AR5X3KP4WQKE7HNSUSDS7SHFMC2WPVSZ2ZXJ6DHXOKHFFKOZE6GK`, whose `fetch_total_managed_funds()` reports 133,818,187,700 underlying fully allocated to its Blend strategy.

`node test.js projects/peridot/index.js`:

```
before:  stellar  383.00      total  1.43 k
after:   stellar  4.72 k      total  5.77 k
```

### On double-counting

The DeFindex-deployed portion is Peridot users' deposits — Peridot retains the shares and the withdrawal claim, and DeFindex/Blend is the yield venue rather than the depositor. Excluding it would leave the majority of Peridot's Stellar deposits unrepresented on DefiLlama entirely. Happy to adjust if you'd prefer this split out differently.

## Summary by CodeRabbit

* **Bug Fixes**
  * Stellar TVL calculations now include liquidity forwarded to DeFindex boosted vaults.
  * TVL reflects each ReceiptVault’s available liquidity for more accurate reporting.

* **Documentation**
  * Updated methodology details to explain the available-liquidity calculation.

### DIFF
--- a/projects/peridot/index.js
+++ b/projects/peridot/index.js
@@ -11,10 +11,16 @@ const stellarMarkets = [
   { vault: 'CD3WN3PLW63HFZXE56OTRLMBV46WG54TFPGRL4RDQ43HQTTWVB4RPO3G', underlying: ADDRESSES.stellar.EURC },
 ]
 
-// Underlying sitting in the vault, i.e. the Soroban equivalent of Compound's getCash().
+// Unborrowed underlying owned by the vault, i.e. the Soroban equivalent of Compound's getCash().
+//
+// This must be read from the vault's own ledger rather than from the SAC balance of the vault
+// address: each ReceiptVault forwards idle underlying into a DeFindex "boosted" vault for extra
+// yield, so the underlying it owns is (SAC balance held directly) + (value of the DeFindex shares
+// it holds). get_available_liquidity() returns that sum. Reading balance(underlying, vault) sees
+// only the un-swept remainder and understates every market by roughly an order of magnitude.
 async function stellarTvl(api) {
   await Promise.all(stellarMarkets.map(async ({ vault, underlying }) => {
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR

Issue #20156, opened 2026-07-21
Proposal: cross-adapter owner-address collision check (built + validated, open to submitting as a PR)


BhariGowda:
Built a small tool to catch a specific bug shape at scale, sharing it here first since it adds a new CI check rather than fixing a specific bug, wanted to check if this is wanted before opening a PR that adds workflow files.

**The bug shape it targets:** several real double-counted-TVL bugs have had the same underlying cause, the SAME contract address (a custody wallet, vault, or pool) being summed as its own TVL by two unrelated protocol adapters. No single adapter's code looks wrong in isolation, it only shows up by looking at all adapters at once, which a script can do and manual review can't.

**How it avoids false positives:** an early version just matched any address mention shared between files and produced 900+ candidates, almost all legitimate composability (e.g. a leverage vault built on Aave naturally references Aave's pool address). Sharpened it to only count addresses that fall inside an actual `owner`/`owners` value (not just near the keyword, and explicitly excluding addresses that are really inside a sibling `tokens`/`token` array in the same call). That cut it to single digits. Also tracks `doublecounted: false` (not just `true`) as "a human already reviewed this exact situation," so already-handled cases aren't re-flagged.

**Validated find:** it surfaced bouncebit-cedefi/index.js sharing 6 owner addresses with bouncebit-premium/index.js bouncebit-cedefi literally has a `// ===== inlined from ./premium =====` section duplicating premium's addresses, and its own methodology string matches the live site's stated methodology exactly ("Calculate TVL by querying BounceBit Cedefi subgraph, EasyBTC and Premium protocols"). Traced it all the way to api.llama.fi/protocols' `module` field to check live registration status before concluding anything either way, turns out bouncebit-premium/index.js isn't currently registered as its own live listing (module points elsewhere, to a different file, `bouncebit-prime/index.js`), so this specific case isn't currently causing a live double-count, but it's exactly the shape of thing worth flagging before merge, and confirmed the tool's precision holds up against a real, traced-through-to-production case.

**What I'm proposing to add**, if useful:
- `scripts/checkAdapterCollisions.js` zero dependencies (only Node's `fs`/`path`), has a `--full` mode (one-time/periodic audit) and a `--changed <files>` mode (CI: check only the files changed in a PR against the rest of the repo)
[...]

PR #20335, opened 2026-08-02
utils/scripts: add cross-adapter owner-address collision check

BhariGowda:
Adds `utils/scripts/checkAdapterCollisions.js`, alongside `checkBTCDupsv2.js` and `checkRegistriesDups.js`, for the case neither covers: an EVM address that two unrelated projects both claim as a balance target. No single adapter looks wrong on its own, it only shows up when you look at every adapter at once.

Context and the fuller writeup are in #20156, including corrections to what I originally claimed there.

### What it counts as a collision

An address only counts when it sits in one of these positions:

- an `owner` / `owners` value
- the owner slot of a `tokensAndOwners` / `ownerTokens` pair, which is the last element of each inner array
- the owners slot of a `tokensAndOwners2` parallel array

It reads those off the parsed syntax tree rather than matching near a keyword, so an address in a sibling `tokens` array is never picked up. That is the whole reason the output is usable: a plain "this address appears in two files" scan produces around 900 hits, almost all of it ordinary composability, such as a leverage vault referencing Aave's pool address. Restricting it to owner positions takes that to single digits.

Adapters whose owner is a variable or comes from an import are skipped rather than guessed at, so it under-reports on purpose. `projects/helper`, `projects/config`, `projects/treasury` and `projects/entities` are excluded, the last two because tracking a protocol's own treasury alongside its TVL is intended rather than a double count.

### What it reports on current main

```
Found 8 owner-address collisions across 2-6 projects.

=== Needs a look (0) ===

=== Already carries a doublecounted flag (2) ===
[high] 0x1244eae9fa2c064453b5f605d708c0a0bfba4838  ->  punk-strategy(doublecounted:true), tokenworks(doublecounted:true)  (owner list sizes 1/1, 2 addresses shared by this same pair)
[high] 0xc50673edb3a7b94e8cad8a7d4e0cd68864e33edf  ->  punk-strategy(doublecounted:true), tokenworks(doublecounted:true)  (owner list sizes 1/1, 2 addresses shared by this same pair)

=== Reviewed and dismissed, see adapterCollisionsReviewed.json (6) ===
[the six bouncebit-cedefi / bouncebit-premium addresses]
```

Being straight about it: there is nothing outstanding for anyone to act on right now. The two flagged pairs are already handled and the six BounceBit ones I looked into and dismissed. The value is in catching the next one before it merges, not in a backlog it clears today.

I did check that it fires on a case that actually happened. Run at `8d62c294^`, before `projects/free-protocol/index.js` was removed in the dead project consolidation, it reports 10 collisions including the two custody multisigs shared by `bsquared` and `free-protocol` from #20208.

### Ranking and the reviewed file

Output is ranked rather than flat. An address that is the sole owner target on both sides reads differently from one sitting in a long shared list, and a pair of projects sharing several addresses scores above a single address in common.

`adapterCollisionsReviewed.json` holds settled cases with the reason written next to them, so they stop coming back. Same purpose as the `IGNORED` set in `checkRegistriesDups.js`, just in a separate file so each entry can carry its reasoning. Seeded with the six BounceBit addresses: `bouncebit-cedefi` inlines `bouncebit-premium`'s owners (there is an `// ===== inlined from ./premium =====` block at line 47), but no protocol on `api.llama.fi/protocols` has `bouncebit-premium/index.js` as its module, the nearest is BounceBit Prime at `bouncebit-prime/index.js`. So nothing is double counted today. If that adapter is ever registered, deleting those entries brings it back.

Entries can be appended with `--ack <address> <reason>`, which fills in the project list and date, or written by hand.

### Modes

```
node utils/scripts/checkAdapterCollisions.js
node utils/scripts/checkAdapterCollisions.js --json
node utils/scripts/checkAdapterCollisions.js --changed projects/foo/index.js
node utils/scripts/checkAdapterCollisions.js --ack 0xabc... "why this is fine"
```

`--changed` checks only the files given against the rest of the repo. **Nothing here is wired to CI.** This is a manual script like the other three. If it turns out to be useful I would raise the PR-time hook separately, and I would rather it earned that first.

### Notes

- No new dependency. Parsing uses the `Linter` API from `eslint`, which is already a devDependency. `package.json` is deliberately untouched, so there is no `npm run` entry; happy to add one if you would rather have it wired up that way, but that file is CODEOWNERS protected and the test workflow guards it.
- Takes about 12 seconds over the 3,983 adapter files. Slower than it needs to be because of eslint's per-file overhead, which is the tradeoff for not adding a parser dependency.
- Verified the three paths of `--changed` by hand: a scratch project claiming an address already owned by `anvil` reports it at high confidence; a changed file whose collisions are in the reviewed file stays silent; and `projects/punk-strategy/index.js` prints the INFO line rather than reporting, because a `doublecounted` flag is already set.
- `npx eslint -c eslint.config.js utils/scripts/checkAdapterCollisions.js` is clean.

## Summary by CodeRabbit

* **Chores**
  * Added automated checks to identify conflicting address ownership and balance targets across supported adapters.
  * Added structured reporting options for easier review of detected collisions.
  * Added tracking for reviewed, known exceptions to reduce repeated reports.
  * Included safeguards to exclude recognized infrastructure addresses and irrelevant namespaces from collision results.

### DIFF
--- a/utils/scripts/adapterCollisionsReviewed.json
+++ b/utils/scripts/adapterCollisionsReviewed.json
@@ -1,0 +1,58 @@
+{
+  "reviewed": [
+    {
+      "address": "0x1ddd6e5ea766511cc0f348dc8d17578a821b680f",
+      "projects": [
+        "bouncebit-cedefi",
+        "bouncebit-premium"
+      ],
+      "reason": "bouncebit-cedefi inlines bouncebit-premium's owners (see the 'inlined from ./premium' block), but projects/bouncebit-premium/index.js is not registered as a live listing: no protocol on api.llama.fi/protocols has that module, the closest is BounceBit Prime at bouncebit-prime/index.js. Nothing is double counted today. Revisit if bouncebit-premium is ever listed.",
+      "reviewedOn": "2026-08-02"
+    },
+    {
+      "address": "0x426cd147ff93f31bb18f1acd19dab9c32d934131",
+      "projects": [
+        "bouncebit-cedefi",
+        "bouncebit-premium"
+      ],
+      "reason": "bouncebit-cedefi inlines bouncebit-premium's owners (see the 'inlined from ./premium' block), but projects/bouncebit-premium/index.js is not registered as a live listing: no protocol on api.llama.fi/protocols has that module, the closest is BounceBit Prime at bouncebit-prime/index.js. Nothing is double counted today. Revisit if bouncebit-premium is ever listed.",
+      "reviewedOn": "2026-08-02"
+    },
+    {
+      "address": "0x55a55e8b08b091bd0529bf1af05b69ff5291867d",
+      "projects": [
+        "bouncebit-cedefi",
+        "bouncebit-premium"
+      ],
+      "reason": "bouncebit-cedefi inlines bouncebit-premium's owners (see the 'inlined from ./premium' block), but projects/bouncebit-premium/index.js is not registered as a live listing: no protocol on api.llama.fi/protocols has that module, the closest is BounceBit Prime at bouncebit-prime/index.js. Nothing is double counted today. Revisit if bouncebit-premium is ever listed.",
+      "reviewedOn": "2026-08-02"
+    },
+    {
+      "address": "0xa2b283e4dbdfea5461c36a59e3b94b3ef2883085",
+      "projects": [
+        "bouncebit-cedefi",
+        "bouncebit-premium"
+      ],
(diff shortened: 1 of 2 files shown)


RohanNero:
Thank you for the detailed explanation and the PR. Sorry but for now going to close since this would only catch those specific `owner`/`owners` keyed duplicates and would require us to continue updating the static reviewed whitelist

BhariGowda:
Understood, and thanks for taking the time to read it properly and for saying plainly why. Both points are fair: it is scoped to the owner/owners positions by design, and a static reviewed list is real ongoing upkeep for a check that clears nothing today. Better to hear that now than to have it sitting in the repo going stale. Appreciate the review.

PR #20338, opened 2026-08-03
Add KeelLabs adapter

KeelLabsYield:
KeelLabs is a non-custodial automated liquidity manager. A VaultFactory deploys one EIP-1167 clone per user per pool, and each vault holds a single Uniswap V3 concentrated liquidity position that a restricted keeper re-centers. The keeper can only rebalance — it can never move funds out.

**TVL**: for every vault of every factory on each chain, the vault's idle token0/token1 balances plus the token0/token1 backing its open Uniswap V3 position. All read on-chain via `sumTokens2` + `resolveUniV3`, no external APIs.

**Chains**: arbitrum, robinhood, hyperliquid.

Each chain lists more than one factory on purpose: `VaultFactory.implementation` is immutable, so every new vault implementation ships with its own factory and the older ones stay live with the vaults they already created.

On hyperliquid the position manager is passed explicitly via `uniV3ExtraConfig.nftAddress` because the vaults use PRJX's deployment, not the chain default in the helper.

App: https://keellabs.app

## Summary by CodeRabbit

- **New Features**
  - Added TVL tracking for KeelLabs across Arbitrum, Robinhood, and Hyperliquid.
  - Included vault balances and underlying Uniswap V3 liquidity in TVL calculations.
  - Added published methodology and chain-specific TVL reporting.

### DIFF
--- a/projects/keellabs/index.js
+++ b/projects/keellabs/index.js
@@ -1,0 +1,68 @@
+const { sumTokens2 } = require('../helper/unwrapLPs')
+
+// KeelLabs deploys one non-custodial vault per user per pool (EIP-1167 clones of a shared
+// implementation) from a VaultFactory, and each vault keeps a single Uniswap V3 concentrated
+// liquidity position that a restricted keeper re-centers. A user may own several vaults, and each
+// chain may have more than one factory (a new implementation always ships with a new factory,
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR

PR #20347, opened 2026-08-04
keellabs: add factories for two new vault implementations

KeelLabsYield:
Follow-up to #20338, which added the KeelLabs adapter.

Two new vault implementations have been deployed since then. `VaultFactory.implementation` is
`immutable`, so every implementation ships with its own factory — this adds the six new ones and
keeps the previous factories, because the vaults they already created are still live and must keep
counting.

| chain | factory | implementation |
|---|---|---|
| arbitrum | `0xAd7f3B6C7D16e19A3284BE0cE14578296feA471A` | VaultRecover |
| arbitrum | `0xF41AA2bb58952F490E2DFe437d50489Ac3c6A4bC` | VaultClaim |
| robinhood | `0x62fC42AA2Aa1F8743d97daBeD925E70E04682a1c` | VaultRecover |
| robinhood | `0x3031B1661Bb584bBA566D74Ba0c86Ab6f525AF07` | VaultClaim |
| hyperliquid | `0x811e2843c2a55b70D9C867988D69E624c35dAF4C` | VaultRecover |
| hyperliquid | `0x609B9A1c089cb29a38bf19901a39259493997AB4` | VaultClaim |

The deployer reuses the same nonce sequence across chains, so a few of these addresses collide with
factories that belong to a *different* chain in this same config. Each one was checked on its own
chain: all thirteen have code, answer `vaultsCount()` and return the expected `implementation()`.

No logic changed — only the `factories` arrays. TVL is still read entirely on-chain through
`api.fetchList` + the `uniV3` helper.

## Summary by CodeRabbit

* **New Features**
  * Added VaultRecover and VaultClaim factory addresses for Arbitrum, Robinhood, and Hyperliquid configurations.

### DIFF
--- a/projects/keellabs/index.js
+++ b/projects/keellabs/index.js
@@ -10,6 +10,8 @@ const config = {
     factories: [
       '0xBfdDA6efE302fC8743Deb9cD7DB4A24Ffcb9E836',
       '0x3031B1661Bb584bBA566D74Ba0c86Ab6f525AF07',   // VaultNext
+      '0xAd7f3B6C7D16e19A3284BE0cE14578296feA471A',   // VaultRecover
+      '0xF41AA2bb58952F490E2DFe437d50489Ac3c6A4bC',   // VaultClaim
     ],
     npm: '0xC36442b4a4522E871399CD717aBDD847Ab11FE88',
   },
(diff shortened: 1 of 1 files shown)


RohanNero:
Thanks for the PR

PR #20368, opened 2026-08-05
fix(hatom-tao-bridge): switch to mint-side MultiversX supply accounting

### LINKED ISSUE #19016: fix hatom tao bridge tvl
https://defillama.com/protocol/hatom-tao-bridge

Ramprasad4121:
Fixes #19016

The previous adapter relied on an outdated Heroku API proxy tracking Bittensor side custody wallets which recently broke, causing TVL to drop to near-zero.

This PR switches the TVL logic to use robust mint side accounting. We now directly query the circulating supply of the Wrapped TAO token (`WTAO-4f5363`) on MultiversX via the existing `elrond.js` helper. This cleanly ensures we only count user bridged value without having to worry about rotating Bittensor multisigs or staked positions.

- Tested locally, returns a healthy & realistic TVL (~$1.13M at current time)

## Summary by CodeRabbit

- **New Features**
  - Added TVL tracking for the Hatom TAO Bridge using validated WTAO circulating supply from MultiversX.
  - Included circulating supply in Bittensor CG token amounts for more complete reporting.

- **Updates**
  - Updated the integration’s TVL methodology to reflect the new data source.
  - Disabled time-travel support to ensure historical values are not inferred.

### DIFF
--- a/projects/hatom-tao-bridge/index.js
+++ b/projects/hatom-tao-bridge/index.js
@@ -2,8 +2,12 @@ const { getTokenData } = require('../helper/chain/elrond')
 
 async function tvl(api) {
   const data = await getTokenData('WTAO-4f5363')
-  // circulatingSupply is already in human-readable units
-  const supply = Number(data.circulatingSupply || 0)
+  const supply = Number(data.circulatingSupply)
+
+  if (!Number.isFinite(supply) || supply < 0) {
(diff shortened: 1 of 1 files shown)


RohanNero:
Sorry, but we'd prefer to only track the collateral side for bridged tokens

PR #20344, opened 2026-08-03
chore(subfrost): update TVL methodology wording

Vdto88:
Text-only change to the SUBFROST listing. No change to what the adapter computes.

**Methodology**, updated at the team's request:

> Total value of all BTC and stablecoins held by the transparent and verifiable SUBFROST signing group. Covers both frBTC deployments: Alkanes and BRC2.0.

For transparency about the computation, which is unchanged: the adapter sums the balances of the two SUBFROST signer custody addresses on Bitcoin L1, one backing frBTC on Alkanes and one backing frBTC on BRC2.0, read straight from the chain. frBTC minted against those deposits is not counted separately, so nothing is double counted.

`node test.js projects/subfrost/index.js` passes and reports 6.97M, unchanged by this PR.

---

**Separate request, since it cannot be done from this repo:** could you also update the protocol `description`? It is not in DefiLlama-Adapters, so we cannot PR it. Current text and requested replacement:

Current:
> SUBFROST is a decentralized custodian on Bitcoin L1. It issues frBTC, a synthetic bitcoin pegged 1:1 to BTC, against BTC held by a FROST signer set (6-of-9 threshold Schnorr).

Requested:
> SUBFROST is a decentralized signing network that issues frBTC and frUSD, trust-minimized assets which power the ultimate UX on Bitcoin L1.
>
> SUBFROST's portfolio of apps deliver this UX directly to end-users, and its API products enable developers to build the next generation of advanced crypto applications.

Happy to move that to an issue or wherever you prefer if a PR comment is the wrong place for it.

## Summary by CodeRabbit

* **Documentation**
  * Updated SUBFROST TVL methodology to clarify that BTC and stablecoins held by the signing group are included.
  * Simplified the explanation of how minted frBTC is treated in TVL calculations.

PR #20374, opened 2026-08-05
feat(subfrost): track the frUSD collateral vault on ethereum

Vdto88:
Follow-up to #20344, which was closed with:

> since the adapter still only tracks btc we'd prefer not to add "and stablecoins" to the methodology. Once the subfrost adapter tracks frUSD backing we can update it.

That is what this does. No methodology claim is added that the adapter does not actually compute.

## What changed

SUBFROST issues **frUSD**, a dollar asset on Alkanes, against USDC/USDT deposited in a vault on Ethereum (`0x95779e7e1c943042255b8a78273fe6de4823cf06`). The adapter tracked only the BTC backing frBTC, so that collateral was not counted at all.

Adds an `ethereum` leg reading the vault's USDC and USDT balances, and updates the methodology to describe both legs.

## Attribution and double counting

The value is reported on **ethereum**, where the collateral actually sits. The frUSD minted against it on Alkanes is not counted again.

That is deliberately the same convention as the existing bitcoin leg, and matches how WBTC and tBTC are handled here: report against the chain holding the deposits, not the chain the representation is issued on. Nothing in this PR is counted in two places.

## Test

```
node test.js projects/subfrost/index.js

------ TVL ------
bitcoin                   7.00 M
ethereum                  65.00

total                    7.00 M
```

The ethereum leg is genuinely small right now: frUSD launched days ago and its supply is ~61 frUSD. I would rather have the plumbing correct and reviewed while the number is trivial than add it later once it is material. It grows with frUSD supply with no further changes.

Happy to adjust the methodology wording if you would prefer it phrased differently.

## Summary by CodeRabbit

* **New Features**
  * Added Ethereum collateral tracking for USDC and USDT held in the frUSD vault.
  * Expanded supported chain coverage to include Ethereum alongside Bitcoin.
  * Added separate total value locked (TVL) reporting for Bitcoin and Ethereum assets.

* **Documentation**
  * Updated methodology details to explain collateral tracking across both Bitcoin and Ethereum.

### DIFF
--- a/projects/subfrost/index.js
+++ b/projects/subfrost/index.js
@@ -1,4 +1,5 @@
 const { sumTokens } = require("../helper/sumTokens");
+const ADDRESSES = require("../helper/coreAssets.json");
 const bitcoinAddressBook = require("../helper/bitcoin-book/index.js");
 
 // SUBFROST is a decentralized signing network on Bitcoin L1. It issues frBTC, a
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the update, I've added frUSD collateral and moved into the sumTokens registry in #20381. We will update the description as described in #20344

PR #20381, opened 2026-08-05
add subfrost frUSD and move to registry

RohanNero:
resolves #20374

## Summary by CodeRabbit

* **New Features**
  * Added SUBFROST coverage for Bitcoin TVL and Ethereum collateral, including USDC and USDT holdings.

* **Refactor**
  * Consolidated SUBFROST tracking into the token-summing registry for improved data coverage.


Vdto88:
Thanks for picking this up and for moving it into the registry.

One small thing, and apologies if it is just a shorthand I do not know: the two legs in the entry have different shapes.

```js
"bitcoin":  { "tvl": { "__btcBook": "subfrost" } },
"ethereum": { "owner": "0x95779e7e1c943042255b8a78273fe6de4823cf06", "tokens": [ADDRESSES.ethereum.USDC, ADDRESSES.ethereum.USDT] }
```

The `ethereum` one has no `tvl` key. Every other functional entry I looked at in `data2.js` wraps the chain config in `tvl` (the `__empty` entries being the exception), including the `bitcoin` leg right above it. Is the flat form supported here, or did the `tvl` wrapper get lost?

I cannot check it from this repo since the registry is consumed server-side, and the API has not picked up the change yet (`methodology` and `description` are both still the old ones), so I could not tell a missing key from a job that has not run. Happy to just wait and confirm from `currentChainTvls` if you would rather.

RohanNero:
@Vdto88 It's fine for chains with only tvl, it would be required if ethereum had a second export, e.g. staking or pool2

image

Vdto88:
Thanks for confirming, that clears it up.

Everything from this PR has landed on the API: `chains` now shows Ethereum alongside Bitcoin, `currentChainTvls` reads `{Ethereum: 64.94, Bitcoin: 7053284}`, and the new methodology is live verbatim.

One leftover from #20344: the `description` is still the old one.

Live now:
> SUBFROST is a decentralized **custodian** on Bitcoin L1. It issues frBTC, a synthetic bitcoin pegged 1:1 to BTC, against BTC held by a FROST signer set (6-of-9 threshold Schnorr).

The text you offered and we approved:
> SUBFROST is a decentralized **signing network** that issues frBTC, a trust-minimized asset which powers the ultimate UX on Bitcoin L1. SUBFROST's portfolio of apps deliver this UX directly to end-users, and its API products enable developers to build the next generation of advanced crypto applications.

No rush at all, and sorry to nudge on a one-word difference. It matters more than it looks: "custodian" is the specific term the protocol is not, since the BTC is held by a threshold signer set rather than by a custodian, and it is the only field we cannot change from this repo.

Vdto88:
@RohanNero following up on this one, and thanks again for taking the frUSD leg into the registry.

We would like to ask for a **category change on the SUBFROST entry (id 8317), from `Bridge` to `Decentralized BTC`**, and we want to lay out the reasoning and the two objections we can see ourselves, so you have everything in one place.

**Why it matters.** On `/chain/bitcoin` our $7.27M does not appear at all, and neither does WBTC at $7.4B. Reconstructing from your own API, the Bitcoin chain total is $3.567B while protocols categorised `Bridge` sum $21.60B on Bitcoin, so the `Bridge` bucket is clearly excluded from chain TVL. That is a category effect, not missing data on our side: `chainTvls.Bitcoin` is populated and has `tvlPrevDay` and `tvlPrevWeek`.

**Why `Decentralized BTC` is the right label for us.** SUBFROST locks BTC and issues frBTC 1:1 against it, with signing done by a 6-of-9 FROST threshold group rather than a custodian. That is the same shape as tBTC, which sits in `Decentralized BTC` and ranks #3 on that page with $317M.

**Two objections we can see, stated up front:**

1. WBTC is still `Bridge`, so you may be treating custodial wrappers as bridges deliberately, and you may see frBTC the same way. The distinction we would point at is the signing model: our group is a threshold signature set, not a single custodian, which is the line tBTC and Chain Fusion are on.
2. We could not check whether leaving `Bridge` removes us from the `/bridges` product, since `bridges.llama.fi/bridge/8317` needs a paid plan. **Does the category change drop us from that product?** If it does, we would like to know before deciding, or to know whether both are possible.

**One hard condition on our side.** Our adapter reports two legs: BTC backing frBTC on Bitcoin, and the USDC and USDT held in the frUSD vault on Ethereum, the one you implemented here in #20381. **If moving to `Decentralized BTC` would mean the frUSD vault leg stops counting, we would rather stay where we are.** Keeping that leg matters more to us than the ranking.

We read the category as a protocol label that does not constrain what the adapter counts, and Chain Fusion looks like the precedent: same `Decentralized BTC` category, with an Ethereum leg of $4.98M in USDT, USDC, WETH and others, and it still ranks on the Ethereum chain page. We also plan to add frDAI backing to the same registry entry later, in the same shape as frUSD. If that reading is wrong, or if the label is meant to be BTC-only in what it reports, please just tell us and we will withdraw the request.

Happy to open a PR if any part of this belongs in the adapters repo rather than on your side.

Vdto88:
@RohanNero gentle bump on the category request above — no rush if it is queued, I just want to make sure it did not get lost.

One number has moved since I wrote it, and it happens to be the one your earlier note in #20344 was about. The frUSD backing the Ethereum leg has gone from **$53.97 to $6,073.99** over the past four days (frUSD supply went 0 → 60.96 → 6,046.59). So the leg you and the team wired into `registries/sumTokens/data2.js` is now carrying a real, growing balance rather than a token amount.

The ask is unchanged: move SUBFROST (id 8317) from `Bridge` to `Decentralized BTC`, on the same reasoning as tBTC — BTC locked, 1:1 representation issued, decentralised by threshold signing (we are 6-of-9 FROST).

And the condition stands, so it is easy to say no to: **if moving out of `Bridge` would mean the frUSD vault leg stops counting, we would rather stay where we are.** Happy to withdraw the request on that basis alone.

One question I could not answer myself: does leaving the `Bridge` category remove us from the `/bridges` product? I could not measure it without a paid plan. If it does, that is useful for us to know before you action anything.

RohanNero:
@Vdto88 Correct, we treat many btc wrappers as bridges since they are locking funds on bitcoin so they can be used on other chains. Changing the category will not stop counting the frUSD tvl, but will remove subfrost from `/bridges`, would you still like to change categories?

Vdto88:
@RohanNero Yes please, go ahead with the change to `Decentralized BTC`.

Thanks for answering both halves of it directly. The frUSD leg continuing to count was the condition on our side, and you cleared it, so dropping off /bridges is a trade we are happy to make: we have no bridge volume adapter in bridges-server, so there is no volume or transaction data of ours in that product to lose.

That last point raises the one thing I would still like to understand, and it is not a blocker for the change above.

If we later write a bridge adapter for frBTC mints and redemptions, would being in `Decentralized BTC` keep us out of /bridges, or is the volume side independent of the category? I ask because the adapters in that repo do not look category-bound: Stargate has one and sits in `Farm`, and threshold-network has one while its protocol tBTC is in exactly the category we are moving to. If the two are independent, we would look at writing that adapter later. If moving category permanently forecloses it, that is useful for us to know now, though it would not change our answer today.

Happy to open a PR if any part of this belongs on the adapters side.

Vdto88:
@RohanNero following up on the category change you offered on Aug 11. You asked whether we still wanted it and we confirmed the same day: yes, `Decentralized BTC` please.

It has not landed yet (the API still reports `category: Bridge`), so I wanted to make sure the thread did not get lost rather than assume it is queued.

Nothing changed on our side: dropping `/bridges` is fine, and the frUSD leg continuing to count was our only condition, which you already cleared. Happy to move this to a fresh issue if that is easier to track on your end.

Vdto88:
Two things — one is a heads-up about a number on this adapter, the other is the category change.

### The ethereum leg looks like it collapsed. It didn't — the collateral moved.

`/protocol/subfrost` currently shows the ethereum leg at ~$58, down from ~$6,074 last week. That is not an outflow. On 2026-08-14 the frUSD vault started deploying its stablecoins into the Curve 3pool, and this adapter reads only the vault address itself, so everything that left the vault stopped being counted.

```
vault    0x95779e7e1c943042255b8a78273fe6de4823cf06   0 USDC + 58.67 USDT
strategy 0x5c5d66c82e2c634074661c3e7427668737c70100   23,410.4630 3CRV
         3pool get_virtual_price() = 1.039824       = $24,342.76
total                                                 $24,401.43
```

Cross-checked against the liability side so it doesn't rest on the transfer trail alone: the strategy was funded with 24,352.6179 USDT net from the vault, and frUSD total supply on Alkanes is 24,352.617871 — equal to the 4th decimal. Collateral / supply = 1.0020.

I've opened https://github.com/DefiLlama/DefiLlama-Adapters/pull/20520 to add the 3CRV holder to the ethereum leg. `node test.js subfrost` goes from `ethereum 58.62` to `ethereum 24.40k`, bitcoin unchanged at 7.22M. There's an open question in that PR about how you'd like the Curve overlap handled — I did not set `doublecounted` since it's per-protocol and would drag the clean bitcoin leg with it, but I'm happy to go whichever way you prefer.

### The category change

On Aug 11 you confirmed that moving to `Decentralized BTC` would not stop the frUSD TVL from counting, and would remove us from `/bridges`. We accepted that trade the same day (18:23 UTC, in this thread). Five days on, `category` is still `Bridge` on the API — since you apply these silently, I assume it fell through the cracks rather than being a reconsideration. Our answer is unchanged: yes please, go ahead.

Happy to answer anything on either item.

PR #20389, opened 2026-08-05
origindollar: exclude AMO-minted OUSD from ethereum TVL

apexearth:
Remove the new OUSD AMO TVL from (protocol owned liquidity) from its TVL.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved total value locked (TVL) reporting by excluding OUSD associated with Curve AMO strategies.
  * TVL calculations now more accurately reflect assets held by the vault rather than OUSD already accounted for in related liquidity pools.


RohanNero:
Thank you for flagging this and opening a PR

PR #20390, opened 2026-08-06
fix(treasury/stockworks): follow the treasury migration and value the v4 position

0xRyanC:
**Protocol:** StockWorks (`treasury/stockworks`) — Robinhood Chain
**Issue:** the listing has reported **\$0 every day since 2026-08-02** (`api.llama.fi/treasury/stockworks` → `currentChainTvls {"Robinhood Chain": 0}`).

### Why it broke

**1. The address is stale.** The registry entry points at `0x57024Aae99f709Bd399252767DDC6487Aa3881De`, the original controller, which was drained on 2026-08-02. The treasury migrated to a successor and then again to the current contract, `SpcxQuoteController` at `0x37bBFeD61d9A8C3743D3E7318e635c37c9abdf8e` (verified on Blockscout).

**2. Updating the address alone would not fix it.** Almost the entire treasury now sits inside a **concentrated Uniswap v4 liquidity position**. v4 custodies liquidity in the PoolManager singleton, so the position is invisible to ERC20 `balanceOf` — summing plain balances returns roughly **\$100** against roughly **\$30k** actually held.

### What this PR does

Valuing the position needs logic beyond a single `treasuryExports()` call, so per the note at the top of `registries/treasury.js` this moves the entry out of the registry and into `projects/treasury/stockworks.js`.

The adapter still uses `treasuryExports` for idle balances, then reads the controller's own position getters (`positionLiquidity`, `tickLower`, `tickUpper`, `venueSqrtPrice`) and converts liquidity into its underlying SPCX and USDG at the pool's current price.

Also: **USDG was missing from the token list** and is now a material share of the book. SPCXSTR, the protocol's own token, remains excluded.

### Test

```
$ node test.js projects/treasury/stockworks.js

--- robinhood ---
SPCX                      25.37 k
USDG                      4.71 k
ETH                       1.23
Total: 30.08 k
```

Matches the treasury read directly on-chain.

## Summary by CodeRabbit

* **New Features**
  * Added treasury valuation support for StockWorks on Robinhood Chain.
  * Valuation now includes native ETH, SPCX, USDG, and eligible concentrated liquidity positions.
  * SPCXSTR holdings are excluded from valuation.

* **Bug Fixes**
  * Removed the outdated StockWorks treasury registry configuration for Robinhood.

### DIFF
--- a/projects/treasury/stockworks.js
+++ b/projects/treasury/stockworks.js
@@ -1,0 +1,61 @@
+const { nullAddress, treasuryExports } = require("../helper/treasury");
+
+// StockWorks (SPCXSTR) — Robinhood Chain.
+//
+// A 10% pool-level swap tax on SPCXSTR funds a treasury that market-makes SPCX
+// (Robinhood tokenized SpaceX stock) against USDG, and converts the harvested
+// LP fees into SPCXSTR buy-and-burn.
+//
+// This cannot be a plain registry entry: almost the entire treasury sits inside
+// a CONCENTRATED Uniswap v4 liquidity position, which is custodied by the v4
+// PoolManager singleton and therefore invisible to ERC20 balanceOf. Summing
+// balances alone reports ~$100 against ~$30k actually held. So we read the
+// controller's own position getters and convert liquidity to token amounts.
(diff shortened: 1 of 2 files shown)


RohanNero:
Thank you for the PR

PR #20396, opened 2026-08-06
current: price SVBTC at BTC parity

paulwu0903:
### Problem
SVBTC (`0x7a479e7a...::svbtc::SVBTC`) has no price on `coins.llama.fi`, so its
reserve was silently counted as $0 — understating TVL by ~$2.7m. 

### Fix
Value SVBTC at BTC parity — it is a LayerZero OFT backed 1:1 by BTC. TVL: $15.71m → $18.43m.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SVBTC price handling by applying precise 8-decimal conversion.
  * Updated displayed total value locked (TVL) and borrowed amounts to use more accurate SVBTC reserve pricing.
  * Preserved existing pricing behavior for all other supported assets.

### DIFF
--- a/projects/current/index.js
+++ b/projects/current/index.js
@@ -10,8 +10,18 @@ const MARKET_OBJECT_IDS = [
   '0xeeef7e9abe201e16c3ca6417b91fa49bec28edcb077eb2fd4a1f126c251e6899', // EthenaMarket
 ]
 
+const PRICE_PROXIES = {
+  '0x7a479e7a6e75323ac9125656a9ca795e11ea42165ac4206af44d1b66e9563be9::svbtc::SVBTC': { coingeckoId: 'bitcoin', decimals: 8 },
+}
+
 function coinTypeFromName(name) {
   return name.startsWith('0x') ? name : `0x${name}`
+}
+
+function addReserveAmount(api, coinType, amount) {
+  const proxy = PRICE_PROXIES[coinType]
+  if (!proxy) return api.add(coinType, amount)
+  api.addCGToken(proxy.coingeckoId, Number(amount) / 10 ** proxy.decimals)
 }
 
 let _reserveRowsPromise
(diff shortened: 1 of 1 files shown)


projects/current/index.js
```diff
@@ -10,10 +10,20 @@ const MARKET_OBJECT_IDS = [
   '0xeeef7e9abe201e16c3ca6417b91fa49bec28edcb077eb2fd4a1f126c251e6899', // EthenaMarket
 ]
 
+const PRICE_PROXIES = {
+  '0x7a479e7a6e75323ac9125656a9ca795e11ea42165ac4206af44d1b66e9563be9::svbtc::SVBTC': { coingeckoId: 'bitcoin', decimals: 8 },
```
RohanNero, line 14: Could you share the btc addresses holding svbtc collateral?

RohanNero (changes requested):

projects/current/index.js
paulwu0903: @RohanNero  Sure — and the honest answer is that there are none I can point you to on-chain. Here's the full picture:

svBTC (`0x7a479e7a6e75323ac9125656a9ca795e11ea42165ac4206af44d1b66e9563be9::svbtc::SVBTC`, "Svalbard BTC", 8 decimals) is a LayerZero OFT deployed 2026-07-30. Its only peer is EID 30101 (Ethereum) → `0x1ad89e03f01880d16d63c33b3abb0e9fd680262b`.
On that contract `token()` returns its own address, so it's a mint/burn OFT, not an adapter with a lockbox — nothing is locked on either side.
Supply is 100 svBTC on Sui (TreasuryCap wrapped in OFT object `0x8ada56c5e09404c9e09aa0eb4ee09a0754c05ee81e526be68f014cd6ad7a0bd9`) and 10 on Ethereum, and the Ethereum contract is owned by a 2-of-3 Safe (`0xb0ab6771f0d6f9fd6c15323ac4f1a392941b47ce`).
Any BTC backing is therefore held off-chain by the issuer, and I couldn't find a published proof-of-reserve for it.

So the proxy was a pricing convention for a BTC-denominated wrapper, not a claim of verified reserves. Happy to drop `PRICE_PROXIES` and leave svBTC uncounted. If there's a bar it needs to clear to get priced properly through the coins server instead — PoR, real DEX volume, a CoinGecko listing — let me know which and I'll go get it.

RohanNero: Thank you for looking into it. We would prefer to add the pricing to the server but would need to verify the collateral backing svBTC first. If you can find and share the BTC addresses holding collateral that matches the svBTC total supply across ethereum and sui then we can include it

FelixBruguera:
Hi, SVBTC is priced now https://coins.llama.fi/prices/current/sui:0x7a479e7a6e75323ac9125656a9ca795e11ea42165ac4206af44d1b66e9563be9::svbtc::SVBTC

PR #19722, opened 2026-06-18
Add Lightspeed TVL adapter

Peyton-Spencer:
## Summary

Adds a DeFiLlama TVL adapter for Lightspeed, Temple's spot orderbook on Canton.

##### Name (to be shown on DefiLlama):

Lightspeed

##### Twitter Link:

https://x.com/temple_ny

##### List of audit links if any:

N/A

##### Website Link:

https://templedigitalgroup.com/

##### Logo (High resolution, will be shown with rounded borders):

https://templedigitalgroup.com/

##### Current TVL:

~2.87M USD at latest local test run

##### Treasury Addresses (if the protocol has treasury)

N/A

##### Chain:

Canton

##### Short Description (to be shown on DefiLlama):

Lightspeed is Temple's private, compliant spot orderbook for institutional markets on the Canton Network.

##### Token address and ticker if any:

No protocol token. Current listed orderbook assets include CC, CBTC, USDCx, and USDA.

##### Category (full list at https://defillama.com/categories) *Please choose only one:

Dexes

##### Oracle Provider(s): Specify the oracle(s) used (e.g., Chainlink, Band, API3, TWAP, etc.):

The adapter does not use a protocol oracle. It reports token quantities from Temple's public exchange-listing API and relies on DeFiLlama pricing for supported assets.

##### Implementation Details: Briefly describe how the oracle is integrated into your project:

N/A for protocol oracle integration. The adapter maps CBTC to BTC pricing, USDA/USDCx to USD Coin pricing, and CC to Canton native gas-token pricing in DeFiLlama accounting.

##### Documentation/Proof: Provide links to documentation or any other resources that verify the oracle's usage:

- https://templedigitalgroup.com/news/temple-lightspeed-trading-canton
- https://api.templedigitalgroup.com/api/exchange/tickers
- https://api.templedigitalgroup.com/api/exchange/orderbook

##### forkedFrom (Does your project originate from another project):

No

##### methodology (what is being counted as tvl, how is tvl being calculated):

TVL is the current resting liquidity on Lightspeed's public spot orderbooks. The adapter fetches every listed Lightspeed market from Temple's public exchange-listing API. Bid-side liquidity is counted as quote assets, ask-side liquidity is counted as base assets, and all mapped assets are summed for Canton TVL. Unknown non-CC assets are skipped rather than failing mapped liquidity, and CC is reported as Canton native gas-token liquidity.

##### Does this project have a referral program?

No

## Test

`node test.js projects/lightspeed/index.js`

Output:

[...]

PR #20400, opened 2026-08-06
Add Temple TVL adapter via CC Tools

Peyton-Spencer:
## What changed

Adds a Canton TVL adapter for the existing Temple protocol listing. The adapter reads Temple's normalized CC Tools feed, validates the asset payload, and adds the already decimal-adjusted CC, USDCx, CBTC, and USDA balances using the CoinGecko IDs supplied by the feed.

## Why CC Tools

This implements the source requested in the review of the earlier adapter attempt. As [RohanNero wrote on #19722](https://github.com/DefiLlama/DefiLlama-Adapters/pull/19722#discussion_r3496352361):

> We prefer not to rely on protocol APIs for TVL anymore. Could you please refactor to use the Lighthouse or CC tools API instead?

CC Tools polls Temple's source endpoint every five minutes, normalizes and prices each asset, and republishes the per-asset balances with canonical Canton instrument IDs, CoinGecko IDs, and the source `updatedAt` timestamp:

`https://api.cctools.network/api/markets/temple/tvl`

## TVL definition

The previous #19722 bot runs reported `$2.48M` and `$2.87M` by measuring orderbook depth. Depth is not TVL: orders can be cancelled and the visible book does not represent all value deposited to support trading.

This adapter instead reports gross user assets deposited in the Canton Token Standard allocation contracts supporting Temple trading. It includes unlocked and locked balances, excludes balances in flight during settlement, and does not subtract borrowing because Temple has no borrow side.

Amounts from CC Tools are already decimal-adjusted and are passed to `addCGToken` without further scaling.

## Validation

- `node test.js projects/temple/index.js` — `$4.33M` across four assets on Canton
- Axios default user agent receives HTTP 200 with four assets
- `eslint -c eslint.config.js projects/temple/index.js`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added Temple TVL tracking for Canton assets.
  * Reports validated asset balances using CoinGecko token data.
  * Added methodology details and configuration for the Canton TVL handler.

### DIFF
--- a/projects/temple/index.js
+++ b/projects/temple/index.js
@@ -16,6 +16,8 @@ async function tvl(api) {
     const deposited = Number(asset.deposited)
     if (!Number.isFinite(deposited) || deposited < 0)
       throw new Error(`temple: invalid deposited amount for ${asset.symbol}`)
+    
+    if (feed.amountsAreScaled !== true) throw new Error('temple: CCTools feed amounts are not scaled')
 
     api.addCGToken(asset.coingeckoId, deposited)
   }
--- a/projects/temple/index.js
+++ b/projects/temple/index.js
@@ -1,0 +1,32 @@
+const { get } = require('../helper/http')
+
+const TVL_URL = 'https://api.cctools.network/api/markets/temple/tvl'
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the updated PR

PR #20401, opened 2026-08-06
vaquita-protocol: drop Base, count Stellar only

raptor0929:
Update to the existing `vaquita-protocol` adapter — not a new listing, so the new-listing section of the template does not apply.

### What changed

The Vaquita Base deployment was shut down on 2026-08-06 and no longer holds user funds, so it is removed from the adapter. TVL is now Stellar-only.

- Removed the `base` export and its now-unused Base pool / MSV / aToken code.
- Added a hallmark `['2026-08-06', 'Base deployment shut down, Stellar only']` so the drop in the chart is explained rather than looking like an unexplained outflow.
- Updated `methodology` to describe the Stellar-only calculation and note the Base shutdown.
- `timetravel: false` is unchanged: Soroban reads go through `simulateTransaction`, which only sees the current ledger.

Historical Base TVL already stored is intentionally left untouched — this only affects new snapshots.

### Methodology

On Stellar, TVL is the total assets managed by the Vaquita-deployed DeFindex vault (`fetch_total_managed_funds`, covering both idle and strategy-invested amounts), plus the idle deposit token balance held by the Vaquita pool contract (`CDTTAZ3NK4MMDHK2C3I6LRDT4YADJZ2QXINKLKQNZUVX7OTUKQNCQGC4`) to back the fixed-yield reward pool.

### Validation

```
$ node test.js projects/vaquita-protocol/index.js

--- stellar ---
USDC                      1.18 k
Total: 1.18 k

------ TVL ------
stellar                   1.18 k

total                    1.18 k
```

Base no longer appears in the breakdown, as intended. No lockfile or `package.json` changes.

### Note for maintainers

I removed the `base` export outright rather than setting a chain-level `deadFrom: '2026-08-06'`, since removal is unambiguous about the intent to stop counting Base. Happy to switch to `deadFrom` instead if that is the preferred convention here.

## Summary by CodeRabbit

* **Updates**
  * Removed Base network TVL tracking following its shutdown.
  * Updated the protocol’s methodology to reflect Stellar-only TVL accounting.
  * Added the Base shutdown date and historical context to the methodology.
  * Stellar TVL tracking remains active and continues to be reported.
  * Base remains listed for historical reference without current TVL figures.

### DIFF
--- a/projects/vaquita-protocol/index.js
+++ b/projects/vaquita-protocol/index.js
@@ -1,10 +1,5 @@
 const { callSoroban, getContractInstanceStorage } = require('../helper/chain/stellar');
 
-// The Base deployment was shut down on 2026-08-06 and no longer holds user
-// funds, so it is no longer counted. Past TVL for Base stays in the historical
-// data; only new snapshots are Stellar-only.
-const BASE_SHUTDOWN = '2026-08-06';
-
 // Vaquita fixed-yield pool on Stellar mainnet. User principal is routed into a
 // Vaquita-deployed and Vaquita-managed DeFindex vault; the reward pool that
 // backs the fixed yield sits in the pool contract as idle deposit token.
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR, I've exported base as `() => ({})` to preserve the historical tvl since removing the chain completely will hide it

PR #20248, opened 2026-07-27
add new stacks sbtc custody address

RohanNero:
## Summary by CodeRabbit

* **Updates**
  * Added an additional Bitcoin address to the supported address list.

PR #20407, opened 2026-08-07
sbtc: derive the reserve address from the signers' aggregate key

BhariGowda:
`stacksSBTC` in the bitcoin addressbook is a hardcoded list, but the sBTC reserve is a key-path-only P2TR output controlled by the signer set's aggregate key. That key rotates, the address rotates with it, and the list goes stale every time.

Right now the three addresses that were in the book before 76c67f1 hold 0.00028 BTC between them, and all 2517.35 BTC sits at the fourth one that commit added.

While the book is stale the protocol reports ~$17 instead of ~$161m. Looking at the served series it has flipped between the two repeatedly, including 2026-08-02 and 2026-08-05. Each recovery has needed a manual PR (#18236, and #20248 which sat open for ten days before 76c67f1 landed the same address by hand).

## What this does

Reads `get-current-aggregate-pubkey` from `SM3VDXK3WZZSA84XXFKAFAF15NNZX32CTSG82JFQ4.sbtc-registry` on Stacks and derives the reserve address from it: BIP341 tweak with an empty merkle root, bech32m encoded as witness v1. `getBTCExport` and `checkBTCDupsv2` already accept a fetcher function for a book key, so nothing else changes.

The retired addresses are kept in the list. A rotation is an ordinary bitcoin transaction, so a sweep can be in flight when the adapter runs.

If the registry read fails the fetcher throws rather than falling back to the retired addresses, since falling back would quietly report dust as the whole peg.

## Verification

- The derived address is `bc1pgg0us9y5skfpatq0nhxt7khhk8qxv0zgle36r6yxnql6dvyaafqsahn043`, exactly the one 76c67f1 added by hand.
- The fetcher's output is identical to the current hardcoded list, so this is a no-op today and only differs after the next rotation.
- The derivation also reproduces the BIP86 test vector (`cc8a4bc6...c115` to `bc1p5cyxnuxmeuwuvkwfem96lqzszd02n6xdcjrs20cac6yqjjwudpxqkedrcr`), so it is checked against a published vector and not only against one live sample.
- Balance at the derived address is 2517.34693101 BTC against sBTC total supply on Stacks of 2517.34662646 BTC, a 0.0012% difference.

No new dependencies, just axios and node's `crypto`, same as the other bitcoin-book fetchers.

I could not exercise the bitcoin balance fetch locally, it returns empty for every book key here including untouched ones like `magpie`, so that part is unverified on my side.

## Summary by CodeRabbit

* **New Features**
  * sBTC reserve addresses are now derived automatically from the current Stacks signer-set key instead of relying on a fixed list.
  * Added support for generating the current Taproot reserve address and keeping retired reserve addresses available for sweep coverage.

### DIFF
--- a/projects/helper/bitcoin-book/index.js
+++ b/projects/helper/bitcoin-book/index.js
@@ -18,13 +18,9 @@ const bitomato = ["bc1qgmtx3caf8rlxmzw703ga2sljv3rkkj39e4ysk9"];
 
 const lbank = ["1MZwhQkkt9wy8Mwm4rx5W3AYiDCJLasffn"];
 
-const stacksSBTC = [
-  // https://docs.stacks.co/concepts/sbtc/clarity-contracts/sbtc-deposit
-  "bc1pl033nz4lj7u7wz3l2k2ew3f7af4sdja8r25ernl00thflwempayswr5hvc",
-  "bc1prcs82tvrz70jk8u79uekwdfjhd0qhs2mva6e526arycu7fu25zsqhyztuy",
-  "bc1p6ys2ervatu00766eeqfmverzegg9fkprn3xjn0ppn70h53qu5vus3yzl0x",
-  "bc1pgg0us9y5skfpatq0nhxt7khhk8qxv0zgle36r6yxnql6dvyaafqsahn043",
-];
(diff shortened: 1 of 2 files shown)


RohanNero:
Thank you for the PR

PR #20413, opened 2026-08-07
sentry-trading: cover v4 launches and add Ink

mavrkofficial:
The existing adapter enumerates only the original v3 factory on Robinhood Chain. Since then Sentry moved new launches onto Uniswap v4 factories and deployed the same stack on Ink, so a growing share of the protocol's liquidity is not counted.

**What's added**

- Robinhood v4 WETH factory `0x472286b7…` and tokenized-stock factory `0xd0A93885…`
- Ink: v3 factory `0xDc37e11B…`, agent factory `0x733733E8…`, v4 factory `0xcF44b151…`
- v3 handling is unchanged

**Why v4 needs more than a config entry**

v4 has no per-pool contract and no LP NFT. Pool funds sit commingled in the PoolManager singleton and the position is a mapping entry keyed by owner, so neither pool balances nor an NFT position can be read. Instead the adapter enumerates launches from `TokenDeployed`, reads the tick range from the factory's `launches()`, then derives each position's base-asset reserve from `getSlot0` and `getPositionLiquidity` on StateView via `addUniV3LikePosition`. Both the vault and factory position keys are probed because custody differs by launch vintage (older launches sit in the factory, current ones in the immutable SentryLPVault).

**Methodology unchanged**

Only the base side of each pair is counted, as this adapter already did. The launched token's own side is excluded since its only market is the pool being measured. Base assets are WETH, plus the tokenized stocks used as pair assets on Robinhood.

**Test**

```
node test.js projects/sentry-trading/index.js

--- robinhood ---
WETH    45.24 k
NFLX     3.33 k
GME      90.61
SPCX     20.81
AMZN      0.56
Total:  48.68 k

--- ink ---
WETH     1.32 k
Total:   1.32 k

total   50.00 k
```

## Summary by CodeRabbit

* **New Features**
  * Expanded Sentry Trading coverage to include Robinhood and Ink across multiple chains.
  * Added TVL reporting for both networks across V3 and V4 launches.
  * Added launch metadata and position tracking for V4 deployments.
  * Improved liquidity tracking for factory- and vault-held positions.
  * Enhanced reserve calculations using liquidity and pool state.
  * Excluded launched-token balances from base-asset TVL calculations.
  * Updated methodology documentation to reflect the expanded coverage.

### DIFF
--- a/projects/sentry-trading/index.js
+++ b/projects/sentry-trading/index.js
@@ -1,24 +1,209 @@
+const { ethers } = require('ethers')
 const { getLogs2 } = require('../helper/cache/getLogs')
-const { sumTokens2 } = require('../helper/unwrapLPs')
+const { sumTokens2, addUniV3LikePosition } = require('../helper/unwrapLPs')
 const ADDRESSES = require('../helper/coreAssets.json')
 
-const FACTORY = '0x9e8f6f8214b01Fd4Cf1d73FB1fb7cf9f811036Cb'
-const WETH = ADDRESSES.robinhood.WETH
-const FACTORY_DEPLOY_BLOCK = 1431636
+// Sentry Launch Factory (sentry.trading): a token launchpad on Robinhood
+// Chain and Ink. Every launch seeds a pool paired against a base asset
+// (WETH, or a tokenized stock on Robinhood) and permanently locks the LP.
+//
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR

PR #20415, opened 2026-08-07
use getConfig

noateden:
## Summary by CodeRabbit

* **Performance**
  * Improved configuration and market-data retrieval through shared caching across multiple integrations.
  * Reduced repeated external requests while preserving existing endpoints and processing behavior.

* **Reliability**
  * Standardized handling of missing configuration and contract data.
  * Maintained existing validation and error behavior for unavailable market information.

### DIFF
--- a/projects/myso-v2/index.js
+++ b/projects/myso-v2/index.js
@@ -1,7 +1,7 @@
 const { getLogs } = require("../helper/cache/getLogs");
 const { sumTokens2 } = require("../helper/unwrapLPs");
 const { get } = require("../helper/http");
-const { getCache, setCache } = require("../helper/cache");
+const { getConfig } = require("../helper/cache");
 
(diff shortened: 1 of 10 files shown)


projects/myso-v2/index.js
```diff
@@ -24,15 +24,14 @@ const CONFIG = {
...
-  const data = await get(url, { responseType: "json", decompress: true, validateStatus: () => true });
-  
-  if (data && data.contracts && data.contracts.length) {
-    await setCache("myso-v2", api.chain, data.contracts);
-    return data.contracts;
-  }
-
-  const cached = await getCache("myso-v2", api.chain);
-  return (cached && cached.length) ? cached : [];
+  const contracts = await getConfig('myso-v2/' + api.chain, undefined, {
```
RohanNero: The api appears down, so switching to getConfig here will cause it to return 0 instead of using the cache

PR #20463, opened 2026-08-11
avalove: include protocol-level $AVLO staking (ProtocolYield) in TVL

AvaLoveApp:
Extends the existing AvaLove adapter (`projects/avalove/index.js`) to also count the protocol-level \$AVLO staking pool, which was previously omitted.

AvaLove has two distinct staking systems:
1. **Per-game house liquidity** — tokens staked into each game pool (already covered).
2. **Protocol-level \$AVLO staking** (`ProtocolYield` contract) — users stake the native AVLO token to earn a share of treasury fees. This was missing.

This PR adds the `ProtocolYield` deployments to the `staking` export by summing the AVLO balance they custody:
- Avalanche: `0x5C0A35ABCBAb5b6F45d5e122ecAbe0d2678513E5` (AVLO `0x54eEeB249E3AE445f21eb006DEbB33eFa2B4b3Bb`)
- Robinhood Chain: `0x3ae963eD481f95BcF4bd69aD9E6390f84bc68C04` (AVLO `0x7e37298e240c1E644F6F9F96b6A3AA6C5aea9885`)

No change to the per-game TVL/staking enumeration or methodology beyond the added sentence describing the protocol staking pool.

## Summary by CodeRabbit

* **New Features**
  * Staking data now includes supported non-core game-pool tokens.
  * AVLO held by the relevant protocol yield contracts is now reflected in staking totals across supported chains.

* **Documentation**
  * Updated methodology details to explain how protocol-level AVLO staking is calculated.

PR #20465, opened 2026-08-11
avalove: price AVLO via on-chain v4 AVLO/ARENA mid price + include protocol staking

AvaLoveApp:
Two improvements to the AvaLove adapter (`projects/avalove/index.js`), both pure on-chain reads:

**1. Price AVLO (previously \$0).** AVLO has no direct DefiLlama price — its only on-chain market is a Uniswap **v4** pool `AVLO/ARENA` on Avalanche, and **ARENA is already priced by DefiLlama**. Rather than fetch any external price, the adapter reads the pool's on-chain mid price via `StateView.getSlot0(poolId)` (`sqrtPriceX96`) and accounts AVLO balances as their **ARENA-equivalent**, which DefiLlama then prices.
- v4 StateView (Avalanche): `0xc3c9e198c735a4b97e3e683f391ccbdd60b69286`
- Pool id: `0x18f3b5331e528e9a37200708c6c5f4de1b33536c0a0c9ad7856ebab7f37a84a9`
- Verified live: mid price ≈ `0.000795 ARENA/AVLO` → ≈ `$0.00000078`, matching the market.

**2. Include protocol-level \$AVLO staking** (`ProtocolYield` contract), previously omitted:
- Avalanche: `0x5C0A35ABCBAb5b6F45d5e122ecAbe0d2678513E5`
- Robinhood: `0x3ae963eD481f95BcF4bd69aD9E6390f84bc68C04`

Per-game TVL/staking bucketing is unchanged (launchpad tokens stay under `staking`, per prior maintainer guidance). Robinhood AVLO is left as a raw balance (no local ARENA pool to derive from).

Note: this supersedes #20463 (which only adds the protocol-stake part). If the price-derivation approach isn't desired, #20463 is the minimal alternative.

## Summary by CodeRabbit

* **New Features**
  * Added support for displaying AVLO holdings from protocol yield and AVLO staking pools.
  * Added AVLO-to-ARENA equivalent valuation using Avalanche Uniswap v4 pricing data.
* **Bug Fixes**
  * Improved staking calculations by excluding core assets and processing non-core tokens separately.
  * AVLO balances on other networks are now included in staking totals using their raw balances.
* **Documentation**
  * Updated staking methodology to reflect protocol-level AVLO holdings and Avalanche price conversion.

### DIFF
--- a/projects/avalove/index.js
+++ b/projects/avalove/index.js
@@ -33,6 +33,31 @@ const FACTORIES = {
   ],
 };
 
+// Protocol-level $AVLO staking (ProtocolYield). Separate system from per-game
+// house liquidity: users stake native AVLO to earn a share of treasury fees.
+const PROTOCOL_YIELD = {
+  avax: "0x5C0A35ABCBAb5b6F45d5e122ecAbe0d2678513E5",
+  robinhood: "0x3ae963eD481f95BcF4bd69aD9E6390f84bc68C04",
+};
+
+const AVLO = {
+  avax: "0x54eEeB249E3AE445f21eb006DEbB33eFa2B4b3Bb",
+  robinhood: "0x7e37298e240c1E644F6F9F96b6A3AA6C5aea9885",
+};
+
+// AVLO has no direct DefiLlama price. Its only on-chain market is a Uniswap v4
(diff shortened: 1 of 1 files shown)


RohanNero:
Apologies for the inconvenience, but we prefer not to include pricing in tvl adapters. Once AVLO has more liquidity on Uniswap, we can price it on the server

PR #20472, opened 2026-08-12
feat(armsys): add Robinhood Chain RWA pools

tradingdesk26:
ARMSys now runs on three tokenized-equity pools on **Robinhood Chain** — NVDA/USDG, INTC/USDG and SPCX/USDG — alongside the existing ETH/USDC pool on Base. This adds that chain to the existing `armsys` adapter. No new npm dependencies, no lockfile changes; the diff is one file.

### Why the Robinhood side is not `uniV4HookExport`

Base is left on the standard helper, untouched.

`uniV4HookExport` resolves a pool's TVL through the Uniswap v4 subgraph and only supports the chains in its `graphIds` map (`base`, `ethereum`). Robinhood Chain has no v4 subgraph — `projects/uniswap-v4` covers that chain through on-chain `Initialize`-event scanning rather than a subgraph entry. Per-hook TVL also cannot be read from the PoolManager singleton balance, because the singleton holds the reserves of every pool on the chain.

So the Robinhood side reads the three pools directly: LP positions minted on the canonical v4 PositionManager, converted to underlying token amounts with the pool's current `sqrtPriceX96` and standard concentrated-liquidity math. Pool `slot0` storage locations are precomputed as `keccak256(abi.encode(poolId, uint256(6)))` and verified against the live contract.

If a v4 subgraph ships for Robinhood Chain, this collapses into one more `graphIds` entry and the custom path can be deleted.

### Verification

`node test.js projects/armsys`:

```
--- robinhood ---
USDG                      582.36
SPCX                      304.31
NVDA                      297.31
INTC                      251.00
Total: 1.43 k
```

Cross-checked against an independent GeckoTerminal read of the same three pools (~$1.49k); the gap is price sourcing, not position accounting.

`base` fails locally with `auth error: payment required for subsequent requests for this API key`. This reproduces identically on the unmodified upstream file, so it is the missing Graph API key in my environment rather than anything in this change.

### Notes

- Addresses: PoolManager `0x8366a39CC670B4001A1121B8F6A443A643e40951`, PositionManager `0x58daec3116aae6D93017bAAea7749052E8a04fA7`, hook `0x20f8B7ec9cC3Bb5c739deDB15a8b4275F84B00c8`.
- SPCX sorts below USDG, so that pool alone has a reversed currency order — handled explicitly.
- All four tokens price on DefiLlama with confidence 0.99.
- `doublecounted: true` and `timetravel: false` are preserved from the current adapter.

## Summary by CodeRabbit

- **New Features**
  - Added ARMSys TVL support for Robinhood Chain.
  - Added tracking for three Robinhood liquidity pools.
  - Included concentrated-liquidity positions and pool activity in TVL calculations.
  - Exposed Robinhood TVL alongside existing Base coverage.

- **Documentation**
  - Updated the methodology to describe TVL accounting across Base and Robinhood Chain.

### DIFF
--- a/projects/armsys/index.js
+++ b/projects/armsys/index.js
@@ -1,20 +1,174 @@
 /**
- * ARMSys — Dynamic-Fee Hook for Uniswap v4 on Base
+ * ARMSys — Dynamic-Fee Hook for Uniswap v4
  *
- *   Base mainnet: ETH/USDC v4 pool, ARMSHookV3 (round-30, 2026-04-27).
- *                 Regime-aware dynamic fees + retail/MEV segmentation
- *                 via on-chain BNS-style volatility classifier.
+ *   Base mainnet:    ETH/USDC v4 pool, ARMSHookV3 (round-30, 2026-04-27).
+ *   Robinhood Chain: NVDA/USDG, INTC/USDG and SPCX/USDG v4 pools,
+ *                    ARMSHookV3RWA (2026-07-31 / 2026-08-03) — tokenized
+ *                    US equities issued on the broker's own chain.
+ *
+ * Base uses the standard uniV4HookExport helper. That helper resolves a pool's
+ * TVL through the Uniswap v4 subgraph and only supports chains present in its
+ * graphIds map (base, ethereum). Robinhood Chain has no v4 subgraph — Uniswap's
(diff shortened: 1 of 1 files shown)


RohanNero:
Apologies for the delay and thank you for the PR

PR #20526, opened 2026-08-17
fix(coinbase-btc): discover custody addresses from PoR and add cbMEGA

DaniiRix:
Discover Coinbase Bridge custody addresses from Coinbase's Proof-of-Reserves feed instead of stale hardcoded lists, and add the missing cbMEGA leg.

## Changes

* Added shared PoR-based custody discovery and `reservesTvl()` helper.
* Replaced stale XRP/DOGE/ADA/LTC address lists.
* Added the missing `megaeth` / cbMEGA leg.
* Removed the unused `coinbaseltc` address-book entry.

## Validation

* PoR reserves verified against native-chain balances.
* Wrapped supply matches the corresponding PoR totals.

| Leg       |        TVL |
| --------- | ---------: |
| Bitcoin   |     $6.20B |
| XRP       |   $114.04M |
| Cardano   |     $9.49M |
| Doge      |     $8.18M |
| Litecoin  |     $3.59M |
| MegaETH   |   $803.91K |
| **Total** | **$6.33B** |

## Summary by CodeRabbit

* **New Features**
  * Added TVL tracking for Coinbase’s MegaETH reserves.
  * Improved Coinbase reserve data retrieval across supported assets.

* **Improvements**
  * Updated Cardano, Bitcoin, XRP, Litecoin, and Dogecoin TVL calculations to use shared reserve data.
  * Improved consistency and accuracy of Coinbase proof-of-reserves reporting.

### DIFF
--- a/projects/helper/bitcoin-book/index.js
+++ b/projects/helper/bitcoin-book/index.js
@@ -1037,12 +1037,6 @@ module.exports = {
     "bc1q0gw7fexuwthyf9wwzrjn4h0flj5veflwgzdxx0727gt9upfk0cfqfjv42k",
     "bc1qfpk3fj2u9kaw8qq96snm72dws5hyxxym5tf8tn",
   ],
-  coinbaseltc: [
-    "LTbMyvoyfSuQNqG5cGihin6BCbiZay11rU",
-    "LVeXnSCw2ci7qq2EGcNwjZqhQ73KrJHNJE",
-    "ltc1qhac8t52gdh8fzeft4ygzxn05nluwwecjrzel99",
-    "LP3k3DmN21xmCay3b5yReLKQKvViCnDPhi",
-  ],
   prosper: [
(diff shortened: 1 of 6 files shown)


projects/helper/bitcoin-book/index.js
```diff
@@ -1037,12 +1037,6 @@ module.exports = {
     "bc1q0gw7fexuwthyf9wwzrjn4h0flj5veflwgzdxx0727gt9upfk0cfqfjv42k",
     "bc1qfpk3fj2u9kaw8qq96snm72dws5hyxxym5tf8tn",
   ],
-  coinbaseltc: [
```
g1nt0ki, line 1040: @DaniiRix we want to keep bitcoin part here, maybe as fetch function since it is dynamic now. This is bit of a hack for us to check there is no overlap in addresses & also used to verify backing of issued btc anchored tokens on other chains (like cbBTC)

g1nt0ki (changes requested):
@RohanNero can you check why this is multiple listings instead of a single one and merge them?

PR #20376, opened 2026-08-05
accountable: add Arbitrum

pawel-accountable:
Adds Arbitrum One (chain id 42161) to the Accountable adapter.

Accountable's strategy factories are deployed on Arbitrum and the first vault there (Noon Symphony Yield Vault, `0x9fe29A3feC55414F70454E2dBe6522f48cEd3E85`, USDC) is live. Without this entry its ~$2.3M is missing from Accountable's TVL entirely.

The three factories listed are the ones registered for Arbitrum (FixedTerm, OpenTerm, Yield). They share addresses with the `ethereum` entry because the deploys are deterministic across chains.

No methodology change: the vault is discovered through the existing `strategyProxies` / `strategyVaults` factory enumeration and valued via `convertToAssets`, same as every other chain.

Verified with `node test.js projects/accountable`:

```
--- arbitrum ---            USDC 78.16 k
--- arbitrum-borrowed ---   USDC 2.24 M
```

This matches the vault read directly on-chain (`totalAssets` 2,319,181.83 USDC = liquid 78,186.37 + borrowed 2,240,995.46). No other chain's numbers changed.

## Summary by CodeRabbit

* **New Features**
  * Expanded Arbitrum support for vault discovery.
  * Added Arbitrum exports for total value locked and borrowed balances.

PR #20537, opened 2026-08-17
accountable: exclude vaults not yet publicly listed

pawel-accountable:
These strategy addresses are deployed on-chain but not yet publicly listed on our platform, so they should not be counted in TVL until they are.

## Summary by CodeRabbit

* **Bug Fixes**
  * Excluded unpublished strategies from vault listings.
  * Improved vault loading so batches containing only excluded strategies no longer stop the listing process prematurely.

### DIFF
--- a/projects/accountable/index.js
+++ b/projects/accountable/index.js
@@ -37,6 +37,27 @@ const EXTRA_VAULTS = {
     ],
 }
 
+// Deployed on-chain but not yet publicly listed on the Accountable platform —
+// excluded from TVL until they are.
+const EXCLUDED_STRATEGIES = {
+    ethereum: [
+        '0x6378767e76ab068b6b1a01bd6e200beca339d21c',
+        '0x93f0b21693bf992417317b4074af4ee10d4e7d3a',
+        '0x8a5afc1d1efccf72cbb6daa885112f36da2682b4',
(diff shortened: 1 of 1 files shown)


pawel-accountable:
Good catch in general, but this ambiguity (a `permitFailure`-null indistinguishable from a genuine empty slot) predates this PR — the original `strategies.filter((s) => s && s !== NULL_ADDRESS)` already treated a failed call the same as end-of-array. This diff only adds the `excludedStrategies` filter on top of that same pre-existing check (`allStrategies` here is exactly the old check, unchanged), so it doesn't alter the failure/empty-slot semantics either way — this PR's scope is just excluding specific known strategy addresses. Agreed it's worth fixing properly (retry-on-null or an explicit length probe instead of using failure as a terminator), but that's a separate change to the core discovery loop — happy to file a follow-up issue for it if that's useful.

pawel-accountable:
@coderabbitai review

pawel-accountable:
@RohanNero this one's ready whenever you have a moment — CI green, CodeRabbit comment addressed (pre-existing discovery-loop behavior, out of scope for this change). Similar shape to #20306 / #20376.

RohanNero:
Thanks for the PR

PR #20583, opened 2026-08-20
xsy: count UTY at its peg, the price feed stopped and TVL has read $0 since 08-05

BhariGowda:
XSY has published **$0 TVL every day since 2026-08-05**, against a live 20.99M UTY supply.

## Not the balance read, the pricing

```
coins.llama.fi/prices/current/avax:0xdbc5192a6b6ffee7451301bb4ec312f844f02b4a
{"coins":{}}
```

The feed ran 2026-04-23 to 2026-08-04 and then stopped. An unpriced token contributes nothing to the total, so `api.add` published a flat zero rather than erroring. Last real value was **$20,979,172** on 08-05.

## The token is fine

```
totalSupply   20,989,270.43 UTY (18 decimals), unchanged
transfers     215,895 and 700,000 UTY on 2026-08-17
contract      live, symbol UTY, name Unity
```

## Change

Book the supply at its dollar peg instead of routing through the dead feed.

```
Test: 0.00 -> 20.98 M
```

which lines up with the $20,979,172 the adapter last published.

## The assumption, stated rather than hidden

The peg is an assumption. What supports it: all 88 quotes the feed published sat between **0.99892 and 0.99953**, a 0.06% band, and the token is still moving in size.

What I could **not** do: find a UTY pool or an alternate price id to confirm the peg still holds today. If UTY breaks its peg this overstates TVL, and it should go back to pricing the token the moment the feed returns.

Flagging that plainly so you can take the other side if you would rather it stay at zero. There is precedent for the pattern either way — `fixBalancesTokens` already maps an unpriced token onto `tether` for `inri`.

## Summary by CodeRabbit

* **New Features**
  * Added methodology details explaining that UTY’s circulating supply is valued at its dollar peg.

* **Bug Fixes**
  * Improved total value reporting by valuing UTY directly at its dollar peg.

### DIFF
--- a/projects/xsy/index.js
+++ b/projects/xsy/index.js
@@ -1,14 +1,28 @@
 const ADDRESSES = require('../helper/coreAssets.json')
 
+// UTY is XSY's dollar unit and TVL is its circulating supply.
+//
+// It is booked at its peg rather than through the token, because the UTY price
+// feed stopped on 2026-08-04 and coins.llama.fi now returns an empty `coins`
+// object for it. An unpriced token contributes nothing, so `api.add` published a
+// flat $0 from 08-05 onwards against a live 20.99M supply.
+//
+// The peg is the assumption here and it is stated rather than hidden: every one of
+// the 88 quotes the feed did publish, 2026-04-23 through 2026-08-04, sat between
+// 0.99892 and 0.99953, a 0.06% band. The token is still live (supply steady at
+// 20.99M, transfers of 215k and 700k UTY on 2026-08-17). If UTY ever breaks its
+// peg this overstates TVL, and it should go back to pricing the token the moment
+// the feed returns.
 async function tvl(api) {
   const totalSupply = await api.call({
     abi: 'erc20:totalSupply',
     target: ADDRESSES.avax.UTY,
   })
-  api.add(ADDRESSES.avax.UTY, totalSupply)
+  api.addUSDValue(totalSupply / 1e18)
 }
 
 module.exports = {
   start: 58017291, // block when UTY contract was deployed
+  methodology: 'Circulating supply of UTY, counted at its dollar peg.',
   avax: { tvl },
 }
--- a/projects/xsy/index.js
+++ b/projects/xsy/index.js
@@ -3,9 +3,14 @@ const ADDRESSES = require('../helper/coreAssets.json')
 // UTY is XSY's dollar unit and TVL is its circulating supply.
 //
 // It is booked at its peg rather than through the token, because the UTY price
-// feed stopped on 2026-08-04 and coins.llama.fi now returns an empty `coins`
-// object for it. An unpriced token contributes nothing, so `api.add` published a
-// flat $0 from 08-05 onwards against a live 20.99M supply.
+// feed published its last quote on 2026-08-04 and coins.llama.fi has returned an
+// empty `coins` object for it since. An unpriced token contributes nothing, so
+// `api.add` published a flat $0 from 2026-08-05 onwards against a live 20.99M
+// supply.
+//
+// `addUSDValue` books this through `coingecko:tether`, so it follows USDT's quote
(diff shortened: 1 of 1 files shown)


BhariGowda:
Your mechanism analysis is right and my PR body was wrong about it. I checked SDK v5.0.220 directly rather than taking it on faith:

- `Balances.ts:160-169` — `addUSDValue` with no `id` falls through to `addCGToken('tether', balance)`, so the amount is priced against USDT's live quote.
- With an `id` it calls `add(id, balance, { isUSDValue: true })`, and `Balances.ts:93` routes that into `_usdBalances`, which is never priced.

So my claim that this books a plain dollar figure was false. I have corrected the comment in the file to say what actually happens, and included USDT's own band over the last 30 days (0.9988 to 0.9997) so the reader can judge it against UTY's 0.99892 to 0.99953.

I am not taking the `id` suggestion, for two reasons:

1. It reads $0 under this repo's own harness. `test.js:60` computes TVL from `api.getBalances()`, which returns `_balances` only, so anything in `_usdBalances` is invisible to it. Confirmed by running it: `avax 0.00`. That is the exact failure this PR exists to fix.
2. It has no precedent here. All 103 `addUSDValue` call sites in `projects/` are the bare form, including `uniswap-v2`, `kamino` and `maiar`. Zero pass an `id`, so that path is unexercised in this repo and I cannot confirm the production runner treats it the way the SDK unit path does.

The bare form follows USDT rather than a hard 1.0, which is a real difference from what I originally wrote, but at 0.12% it is well inside the band this adapter is already assuming, and it is the form the repo actually runs.

Date boundary nit taken: 2026-08-04 last quote, 2026-08-05 first missing day, stated once.

BhariGowda:
Correcting something in my own PR body before this goes any further.

I wrote that I could not find a UTY pool or an alternate price id to confirm the peg still holds. That was wrong. There are three live UTY pools, all on Pharaoh on Avalanche:

```
UTY/USDC     liq $3,393,780   price 0.9995   24h vol $74
UTY/UTY      liq $2,998,569   price 0.9995   24h vol $1
avUSD/UTY    liq $701         price 0.9995   24h vol $74
```

So the peg holds at 0.9995 against real liquidity, not just against the token's own past quotes. That part supports the number this PR produces.

It also cuts the other way, and I think that matters more here. On #20584 you closed a similar change with "if a token is liquid, it should be treated as such". UTY has $3.4M of liquidity in a pool that quotes it right now. By that standard this should get priced rather than booked at an assumed peg, and what I proposed here is the thing you rejected there.

My best read on why the feed misses it: the liquidity is deep but nearly static, $74 of 24h volume across all three pools. That looks like the reason rather than anything about the token itself.

For what it is worth the gap is not only xsy. pharaoh-v3, on the same chain and the DEX actually hosting those pools, has UTY absent from its breakdown every day from 08-13 to 08-21, averaging about $5.5M. It drops the balance rather than reading zero, so it is less visible. That is an argument for fixing this in pricing once rather than per adapter.

For completeness on the feed itself: 20 quotes in the last 40 days, last one 2026-08-04, and `prices/current` and `prices/historical` both return `{}` today.

Happy to close this in favour of getting UTY priced. Say the word and I will close it.

BhariGowda:
Closing this, following up on my correction above.

Re-checked today. UTY still is not priced: `prices/current/avax:0xdbc5192a6b6ffee7451301bb4ec312f844f02b4a` returns an empty `coins` object, and `chart?span=20&period=1d` returns 2 points out of 20, the last one around 08-04. So xsy still publishes $0 and the underlying problem is real.

But the fix in this PR is the wrong one. UTY has real liquidity, roughly $3.39M in the Pharaoh UTY/USDC pool on Avalanche quoting 0.9995, so it should be priced from its market rather than asserted at a peg. That is the same objection that closed #20584, and it applies here too. The remaining gap is pricing coverage, which sits upstream of this repo, so there is no adapter change worth shipping.

Leaving projects/xsy/index.js as it is on master.

PR #20587, opened 2026-08-20
Danogo: Restore TVL to include all assets, not just ADA

TruongQuangHung:
- Bug Fixes
  - Restore TVL to include all assets, not just ADA

## Summary by CodeRabbit

* **New Features**
  * Improved Cardano TVL calculations by including ADA and native asset values held in smart-contract UTXOs.
  * Added support for matching base-address variants more accurately.
  * Added pricing for bridged tokens and excluded negligible dust balances from reported totals.
* **Bug Fixes**
  * Improved reliability and performance when retrieving smart-contract balances through parallel requests and request timeouts.


RohanNero (changes requested):
Can you please share how we can verify the backing for the bonds and price them without the danogo API?

TruongQuangHung:
> 
Let me split it into the two parts: verifying the backing, and pricing it.

- Verifying the backing (fully on-chain, no Danogo API needed)
  - The only thing the adapter takes from our API is the list of smart contract addresses: `curl 'https://danogo-gateway.tekoapis.com/api/v1/smartcontract-addresses'`
   Everything after that is public chain data. The adapter reads the unspent UTXOs at those addresses from Kupo, and the same query works against any Cardano indexer (Kupo, Blockfrost, Koios) or a Cardano explorer

- Pricing - Two things sit in the contracts: 
  - Bonds — priced at a fixed 1 bond = 1000 ADA. There are only 2 bonds live right now, see https://dano.finance/bond. This is the adaValue field, which we now treat as deprecated: it was the original bond-only path.
  - Locked tokens (USDM, USDA, etc.) — this is the large majority of TVL today, since the contracts mostly hold assets rather than bonds. These are priced via exchangeRateNum / exchangeRateDenom, which is what we moved to once the contracts started holding tokens.

Example:
```
curl --location --request POST 'https://danogo-gateway.tekoapis.com/api/v1/cardano-asset-value' \
--header 'Content-Type: application/json' \
--data-raw '{
    "assetIds": [
        "c48cbb3d5e57ed56e276bc45f99ab39abe94e6cd7ac39fb402da47ad.0014df105553444d"
    ]
}'
```
Our endpoint isn't a price source of its own — it's an aggregator over public ones: Orcfax, Liqwid and Djed, plus DEX prices from Minswap, Splash and our own concentrated pools (https://v2.dano.finance/liquidity-pairs). So the token prices can be reproduced from those sources directly

TruongQuangHung:
I think this may be because our protocol information is quite outdated, so you may have mistakenly thought that it only holds bonds.

Could you help us update the information to make it clearer: `Dano Finance is a DeFi protocol on Cardano building a decentralized capital and trading layer. It brings together lending and borrowing, leverage trading, swaps, concentrated liquidity, and advanced DeFi infrastructure designed to make capital more efficient and DeFi easier to use.`

Thank you in advance!

RohanNero:
@TruongQuangHung Thanks for clarifying. Regarding the locked tokens, are any of these minted by danogo against user deposits? By revising to use sumTokens I can see ~5M in tokens that are already priced: 
image

TruongQuangHung:
@RohanNero Hi, I've updated it now so it only counts the price of the underlying tokens, as you suggested.

The increase in ADA is because we don't price dADA any more. We will include our staking contract `addr1xxt4n07cnlafzefqvne69mmxmnzu2t9gtd27jw9d9yvc7u5htxla38l6j9jjqe8n5thkdhx9c5k2sk64ayu262ge3aequnmfak` in the API response, and it alone holds over 10M ADA.

We also price ETH, BTC, USDC and USDT (Wanchain policy `25c5de5f…`) via CoinGecko ids, the same way `projects/liqwid` does — that adds $568K.

Total **$8,356,546**, every asset priced by DefiLlama. Breakdown attached.
danogo-tvl

RohanNero:
@TruongQuangHung If possible, could you please also include the borrowed export? Is field index 2 on the pool/market datum utxos total borrowed?

TruongQuangHung:
> @TruongQuangHung If possible, could you please also include the borrowed export? Is field index 2 on the pool/market datum utxos total borrowed?

But our protocol is currently listed as a DEX, so we’ll create another PR to split it into DEX and lending categories soon. We’ll also return the yield and borrowed loan information that you show in the UI to make the data clearer

RohanNero:
@TruongQuangHung Ok sounds good, one final question: Is the staked ADA part of the dex or the lending protocol? If it is part of the lending protocol can we exclude it here in favor of including it in the new lending adapter?

RohanNero (approved):

TruongQuangHung:
> @TruongQuangHung Ok sounds good, one final question: Is the staked ADA part of the dex or the lending protocol? If it is part of the lending protocol can we exclude it here in favor of including it in the new lending adapter?

It’s a different SC that holds ADA for our protocol’s staking rewards. Don’t worry, we won’t include this in lending since it doesn’t support supply/borrow. I think its TVL should stay here for clearer meaning

RohanNero:
Thank you for the PR

PR #20588, opened 2026-08-20
fix: nested ERC-4626 curator vault TVL conversion

chetna-mittal:
## Summary

- Fix `getNested4626Vaults` so nested ERC-4626 vaults convert the outer `totalAssets()` from inner-vault shares into underlying token units via `convertToAssets`.
- Use batched RPC calls for the conversion step.

## Problem

For nested vaults, the outer vault’s `totalAssets()` is denominated in inner-vault shares, not the final underlying asset. The helper was adding that share amount directly to the underlying token, which undercounted TVL when the inner share price was above 1.

Example: Blend’s Scroll nested vault was off by about 5.6% in the current on-chain state.

## Fix

In `projects/helper/curators/index.js`, when an inner `asset()` exists:
- resolve the underlying token from the nested vault
- convert outer `totalAssets()` with `convertToAssets`
- add the converted amount to TVL

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reporting of assets held through nested vaults by converting vault shares into their underlying asset amounts.
  * Continued using direct asset totals where applicable.
  * Excluded entries when asset or amount information is unavailable.


RohanNero:
Thanks for the PR

PR #20590, opened 2026-08-20
Re Protocol Solana Deployment

Pokesi:
Count Re Protocol's Solana-bridged reUSD as Solana TVL rather than Ethereum TVL.

## Summary by CodeRabbit

- **New Features**
  - Added Solana support for tracking reUSD token supply and total value locked (TVL).
  - Solana TVL is now displayed under the existing reUSD token identifier.
  - Improved precision when converting Solana token supply for cross-network TVL calculations.

- **Bug Fixes**
  - Improved Ethereum TVL accuracy by excluding reUSD supply held on Solana, preventing double counting across networks.


RohanNero:
Sorry for the inconvenience, but for bridged tokens we prefer to track the tvl only on the chains where the collateral is

PR #20547, opened 2026-08-18
feat: add Xona Agent staking adapter (Solana)

suneater-dev:
Adds a TVL adapter for **Xona Agent** staking on Solana.

##### Name (to be shown on DefiLlama):
Xona Agent

##### Twitter Link:
https://x.com/xona_agent

##### List of audit links if any:
No independent audit of our own. Staking runs on Streamflow's audited stake-pool program (`STAKEvGqQTtzJZH6BWDcbpzXXn2BBerPAgQ3EGLN2GH`) rather than custom contracts. https://streamflow.finance/

##### Website Link:
https://xona-agent.com (staking app: https://stake.xona-agent.com)

##### Logo (High resolution, will be shown with rounded borders):
https://xona-agent.com/icon-new.png

##### Current TVL:
~303,924,276 $XONA staked (~$90k at time of writing)

##### Treasury Addresses (if the protocol has treasury)
n/a

##### Chain:
Solana

##### Coingecko ID:
xona-agent

##### Coinmarketcap ID:
(not listed)

##### Short Description (to be shown on DefiLlama):
Lock $XONA across four fixed-term tiers and earn a share of every buyback made from protocol revenue, plus bonus reward tokens from partner projects.

##### Token address and ticker if any:
`Ckit5s1Cpc3RdMh1HrhfW2nAy4PnkkgjXgXMeykbpump` — $XONA

##### Category:
Staking Pool

##### Oracle Provider(s):
None. No oracle is used; balances are read directly from on-chain token accounts.

##### Implementation Details:
n/a — no oracle integration.

##### Documentation/Proof:
n/a — no oracle integration.

##### forkedFrom (Does your project originate from another project):
Not a fork of the code, but worth flagging explicitly: staking is deployed on **Streamflow**'s stake-pool program rather than our own contracts. Streamflow is already listed on DefiLlama. If your methodology already counts third-party stake pools inside Streamflow's TVL, please let me know and I'll close this rather than introduce double counting — happy to follow whatever you prefer here.

##### methodology (what is being counted as tvl, how is tvl being calculated):
$XONA is staked into four Streamflow stake pools, one per lock tier (7, 30, 90 and 180 days). Each pool holds deposits in its own vault token account. The adapter sums the balances of those four vaults on-chain.

Reported under the `staking` key rather than `tvl`, because only the protocol's own native token is deposited. Please re-key it if you'd rather it counted as TVL.

Reward vaults are deliberately excluded — those hold undistributed rewards, not user deposits.

[...]

PR #20591, opened 2026-08-20
feat: add Covenant and Open Vecta staking adapters (Solana)

suneater-dev:
Adds two more Solana staking protocols, same team and same setup as [`xona-agent`](https://github.com/DefiLlama/DefiLlama-Adapters/pull/20547) which @RohanNero kindly merged in #20573. Following that exact pattern, straight into `registries/sumTokens.js`.

Both are deployed on **Streamflow**'s stake-pool program rather than custom contracts, so both are marked `doublecounted: true` to match how xona-agent was handled.

## methodology

Each project stakes its own token across four fixed lock tiers (7, 30, 90, 180 days). Every tier is a separate Streamflow stake pool holding deposits in its own vault token account. The adapter sums those four vaults on-chain.

Reported under `staking` with `tvl: __empty`, since only the protocol's native token is deposited — same call as xona-agent. Reward vaults are excluded, as those hold undistributed rewards rather than user deposits.

Verified locally against mainnet:

```
covenant       123,821,367  solana:2mNVZ6aEjrGwiUVCfz7XGWpiXuWzgBDoznwE579upump
open-vecta     225,057,110  solana:4VBvCEMrPEU6yYF43pdahJafdHGaRxtMtKENqYpnpump
```

## One question, on a third protocol

We have a fourth staking program, **Agent Ranking** (`92eXvBFGBAXwgmPf6t6CUCNJH2xChsCJuWzaGPjyPUmP`, $AR), with ~445.5M staked. It is deliberately **not** in this PR: the token is not in your price database, so an adapter would return a balance you cannot value.

It trades on PumpSwap (~$13k liquidity, ~$0.0000283), which would put it around $12.6k. Is there anything we should do to get it into the price feed, or is that liquidity simply too thin? Happy to submit the adapter separately once it can be priced.

---
No new npm dependencies, and `pnpm-lock.yaml` is untouched.

## Summary by CodeRabbit

* **New Features**
  * Added tracking configurations for the covenant-protocol and open-vecta tokens.
  * Enabled Streamflow Solana staking contract tracking and double-counting support for both tokens.
  * Added token account configuration for improved tracking accuracy.

### DIFF
--- a/registries/sumTokens.js
+++ b/registries/sumTokens.js
@@ -12073,6 +12073,23 @@ const configs = {
       }
     },
   },
+  "covenant": {
+    "methodology": "Tracks CVNT tokens in the Streamflow staking contracts.",
+    "doublecounted": true,
+    "solana": {
+      "staking": {
+        "tokenAccounts": [
+          '5Fg449W8E7EnLfC7Sgn4zCwSmFXP1YbpVEnVQ3WKEwuC',
+          '12zg2Eu2HM2X39ZCoVbQRA6t6xP3Fmz21WskfzbCCsK6',
+          '9fEQ4ow9WM4XkSZiNrEPDgy1rt73WgUJvnKsLWQR1Qus',
+          '4TtDoALukroBBRbibXknqsDkpo3ak1pN5i52LtjgmDyk',
+        ]
+      },
+      "tvl": {
+        "__empty": true
+      }
+    }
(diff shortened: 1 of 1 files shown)


RohanNero:
Thanks for the PR

RohanNero:
Thank you for the PR

PR #20592, opened 2026-08-20
fix: turtle club treasury token pricing

chetna-mittal:
## Problem

TurtleClub’s treasury tracks five ZeroLend aTokens on Linea. These tokens represent rebasing claims on WETH, weETH, ezETH, wrsETH, and USDT, but the coins service does not price the aToken addresses directly. Their balances therefore contribute no value to the treasury total.

The adapter also enables automatic token discovery for 19 treasury wallets, while the helper rejects discovery for more than 11 owners. This causes supported treasury chains to fail instead of returning their explicitly configured balances.

Two existing substitutions are also unsafe:

- eUSDC-2 is now priced directly, so additionally mapping it to USDC could double-count its balance.
- oLYNX is an option token that requires payment to exercise, so it must not be valued 1:1 as LYNX.

## Solution

- Map each ZeroLend aToken to its verified underlying asset.
- Disable automatic discovery when the owner count exceeds the helper limit and use the explicit token list.
- Remove the obsolete eUSDC-2 and incorrect oLYNX substitutions.
- Leave Tempest multi-asset vault shares unmapped because they require share-based multi-asset valuation.

## Verification

- Verified all ZeroLend underlying addresses, decimals, transforms, and rebasing behaviour on-chain.
- Ran `node test.js projects/treasury/turtleclub.js`.
- Ran `node test.js projects/TurtleClub/index.js`.
- Ran syntax, lint, and diff checks.

## Summary by CodeRabbit

- **New Features**
  - Added Turtle Club TVL tracking for Ethereum, Linea, and Avalanche.
  - Added support for valuing ZeroLend rebasing assets through their underlying tokens.

- **Bug Fixes**
  - Improved asset and vault balance handling across supported chains.
  - Refined token pricing metadata for multi-asset vault shares, option tokens, and directly priced assets.
  - Adjusted treasury token fetching for configurations with larger owner sets.


RohanNero:
Thank you for the PR, I've resolved in #20684 to keep the capitalized Turtleclub directory and to remove zerolend price maps as they are now priced

PR #20603, opened 2026-08-20
fix: sumTokens2 crash when covalent discovery has >11 owners

chetna-mittal:
## Summary
- `sumTokens2` previously threw if `fetchCoValentTokens` was enabled with `owners.length > 11` (Ankr/Covalent rate-limit guard).
- That check ran before uniquifying owners, so duplicate addresses could also false-trigger it.
- Affected registry treasuries such as `treasury/cryptonopix` (13 BSC owners) and `treasury/rain` (12 Arbitrum owners) failed (no TVL).

## Fix
- Uniquify `owners` / `tokens` before Covalent discovery.
- Batch Covalent/Ankr token discovery in chunks of 10 instead of throwing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved token balance retrieval for accounts with many owners.
  * Token information is now consolidated more reliably, reducing duplicate results.
  * Enhanced discovery of both current and historical token balances for more complete portfolio data.
  * Large owner lists are processed in smaller batches to improve reliability and consistency.

### DIFF
--- a/projects/helper/unwrapLPs.js
+++ b/projects/helper/unwrapLPs.js
@@ -990,10 +990,6 @@ async function sumTokens2({
   sumChunkSleep,
 }) {
 
-  if (fetchCoValentTokens && owners.length > 11) {
-    throw new Error('fetchCoValentTokens option is not recommended for more than 10 owners due to rate limits')
-  }
-
   if (api) {
     chain = api.chain ?? chain
     block = api.block ?? block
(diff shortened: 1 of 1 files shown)


RohanNero:
Thanks for flagging this, I've resolved in #20607. We'd prefer not to raise the fetchCoValent limit of 11 owners

PR #20610, opened 2026-08-20
feat(wise-token): track staked WISE via token globals(), reword TVL methodology

byyyyyyyyy:
Adds staking TVL to the wise-token adapter, on top of the existing ownerless-ETH Uniswap LP tracking.

- Tracks totalStaked from the WISE token contract's globals() function (0x66a0f676479Cee1d7373f3DC2e2952778BfF5bd6). WISE staking is HEX-style: staked tokens are burned from circulating supply and tracked internally via globals(), not held in a separate vault contract, so this is the only on-chain source for the figure.
- Rewords the TVL methodology string from "ownerless (unclaimed) share" to "ownerless (burned LP) share" for clarity on how the ETH-backing figure is derived (LP tokens sent to the zero address are permanently unclaimable, and their share of pool reserves is what's counted).

## Summary by CodeRabbit

* **New Features**
  * Added tracking for WISE token staking holdings and total value locked.
  * Added the WISE token address to the Ethereum integration.

* **Documentation**
  * Clarified that the liquidity provider share is ownerless or burned rather than unclaimed.

### DIFF
--- a/projects/wise-token/index.js
+++ b/projects/wise-token/index.js
@@ -34,6 +34,6 @@ async function staking(api) {
 module.exports = {
   misrepresentedTokens: true,
   doublecounted: true,
-  methodology: 'TVL = ownerless (unclaimed) share of ETH in the WISE/ETH Uniswap LP',
+  methodology: 'TVL = ownerless (burned LP) share of ETH in the WISE/ETH Uniswap LP',
   ethereum: { tvl, staking }
 };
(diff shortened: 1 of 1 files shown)


RohanNero (changes requested):
Every WISE token is backed by the ETH we currently track correct? Sorry, but we can't additionally include WISE in the tvl (regardless of the export) as this would be counting the same assets twice

byyyyyyyyy:
> Every WISE token is backed by the ETH we currently track correct? Sorry, but we can't additionally include WISE in the tvl (regardless of the export) as this would be counting the same assets twice

should this be moved to dimension adapters instead? is it possible for us to be able to display the staked amount value?

RohanNero:
@byyyyyyyyy We use the dimension adapters repo to track metrics like fees, revenue, and volume. If Wise charges a fee when users stake/unstake, we can track it there

byyyyyyyyy:
> @byyyyyyyyy We use the dimension adapters repo to track metrics like fees, revenue, and volume. If Wise charges a fee when users stake/unstake, we can track it there

for example this staked value on pendle, how do we list our staked value as well. 
image

RohanNero:
@byyyyyyyyy Sorry, but in the case of Pendle, we allow them to track their governance token under the staking export because their adapter doesn't track the collateral that backs PENDLE

PR #20611, opened 2026-08-20
fix(hypersurface): include USDC-collateral pool in HyperEVM TVL

MysticDakra:
Hypersurface deploys one set of MarginPool / HedgedPool / Hedger contracts per collateral. The registry entry only listed the USDT0-collateral pool, so the USDC-collateral pool on HyperEVM was missing from TVL entirely.

This adds the three USDC pool contracts as owners and USDC to the token list. The already-listed Hedger also holds USDC, so that balance was being dropped too. USDT0 is switched to the `hyperliquid` namespace — same address, correct chain.

Verified on-chain (latest block) that the newly added owners hold 3,121,728 USDC and 25,333 WHYPE.

`node test.js projects/hypersurface`:

before
```
hyperliquid   314.38 k
base           11.66 k
total         326.03 k
```

after
```
hyperliquid     5.31 M
base           11.66 k
total           5.32 M
```

No change to the Base section — that pool is USDC-collateral and already covered.

## Summary by CodeRabbit

* **New Features**
  * Expanded Hyperliquid TVL coverage to include MarginPool, HedgedPool, and Hedger contracts across supported collateral pools and networks.
  * Added support for Hyperliquid USDT0 and USDC token references.
  * Included additional owner addresses for improved coverage.

### DIFF
--- a/registries/sumTokens.js
+++ b/registries/sumTokens.js
@@ -16182,15 +16182,19 @@ const configs = {
     },
   },
   "hypersurface": {
-    "methodology": "TVL includes tokens in MarginPool, HedgedPool, and Hedger contracts. LP positions held by the Hedger are unwrapped to their underlying tokens.",
+    "methodology": "TVL includes tokens in MarginPool, HedgedPool, and Hedger contracts of every collateral pool (USDT0-collateral and USDC-collateral on HyperEVM, USDC-collateral on Base). LP positions held by the Hedger are unwrapped to their underlying tokens.",
     "hyperliquid": {
       "owners": [
         "0x7D2e4b4d7ba55C423F5CCe194ae8194eFD1C6e35",
         "0x0095aCDD705Cfcc11eAfFb6c19A28C0153ad196F",
-        "0xa8c9403BDf554C047Ad91a448DDb24208Ab5313c"
+        "0xa8c9403BDf554C047Ad91a448DDb24208Ab5313c",
+        "0x7FfD5706C916499676D707f3ec3F0c9b928E7A95",
+        "0xe0F9cA7FD12E31F5d720A93d04722d6DFbAD59e7",
+        "0x220f86b641ec63f4832CC30a663Bb26b15259Ee2"
       ],
       "tokens": [
-        ADDRESSES.corn.USDT0,
+        ADDRESSES.hyperliquid.USDT0,
+        ADDRESSES.hyperliquid.USDC,
         "0xbe6727b535545c67d5caa73dea54865b92cf7907",
         "0x9fdbda0a5e284c32744d2f17ee5c74b284993463",
         ADDRESSES.hyperliquid.WHYPE,


RohanNero:
Thanks for the PR

MysticDakra:
Flagging something that may need a look on your side — Hypersurface TVL appears to have stopped updating right around this merge, and I would rather raise it than let it sit.

- Last recorded point: **2026-08-20 23:35:35 UTC** at $315,073
- This PR merged: **2026-08-20 23:41:26 UTC**
- It is now ~05:00 UTC and there is still no new point, and no `08-21 00:00:00` daily snapshot at all

Peers on the same chain and of comparable size all got their 08-21 snapshot and are refreshing normally, so it does not look like a general cadence or small-protocol thing:

| protocol | TVL | last stamp | age |
|---|---|---|---|
| Timeswap V2 | $321,207 | 08-21 02:49:47 | 99 min |
| **Hypersurface** | $326,742 | **08-20 23:35:35** | **293 min** |
| Ultrasolid V3 | $293,089 | 08-21 03:13:47 | 75 min |
| HyperLend Isolated | $357,535 | 08-21 03:14:47 | 74 min |
| Hyperdrive HL Earn | $259,252 | 08-21 02:36:35 | 112 min |

The adapter itself runs clean for me against merged `main`, including timetravel to the snapshot that was missed:

```
node test.js projects/hypersurface            -> 5.28 M   (7-15s)
node test.js projects/hypersurface 2026-08-21 -> 5.31 M   (28s)
```

So I cannot reproduce a failure locally. One thing I would suspect is cost rather than correctness: this PR took the hyperliquid entry from 3 owners to 6 and `resolveUniV3` is on, so there are meaningfully more calls than before, and the timetravel run is noticeably slower than the latest-block one. If your runner has a timeout or a stricter HyperEVM RPC budget that could be enough.

Could someone check the error log for this adapter? If it is the call volume I am happy to slim it down — for example dropping tokens that never carry a balance, or splitting the pools. Just let me know which direction you would prefer.

MysticDakra:
Correction to my comment above — please disregard it, my control group was wrong.

I compared Hypersurface against protocols that are defined under `projects/`, which appear to refresh far more often than the ones defined in `registries/sumTokens.js`. Comparing against other registry-based protocols instead gives a very different picture:

| registry protocol | has 08-21 00:00 snapshot | last stamp | age |
|---|---|---|---|
| hypersurface | no | 08-20 23:35:35 | 400 min |
| blackhaven | no | 08-21 01:09:59 | 305 min |
| juicebox-v2 | yes | 08-21 00:56:47 | 319 min |
| covo-v2 | yes | 08-21 04:22:35 | 113 min |
| sacra | yes | 08-21 04:07:35 | 128 min |
| bouncebit-ethena | yes | 08-21 05:29:35 | 46 min |

Registry-based entries refresh anywhere from ~45 to ~400 minutes, and at least one other also has no 08-21 00:00 snapshot. So a ~400 minute gap does not look like evidence of a broken adapter, and the timing next to this merge was most likely coincidence on my part.

Sorry for the noise — no action needed unless you see something in the logs. The adapter itself runs clean against merged `main` (5.28M latest, 5.31M timetravelled to 08-21), so I have no reason to think anything is actually wrong.

MysticDakra:
Reopening this — I retracted my earlier report too quickly, and ten hours on the evidence is much stronger than when I walked it back.

When I checked at 06:15 UTC, `blackhaven` also had no `08-21 00:00` snapshot, so I concluded a long gap was normal for registry-based protocols and told you to disregard. Since then every registry protocol I sampled has caught up. Hypersurface has not.

| registry protocol | last stamp | age | has 08-21 00:00 |
|---|---|---|---|
| **hypersurface** | **08-20 23:35:35** | **602 min** | **no** |
| blackhaven | 08-21 05:50:59 | 227 min | yes |
| juicebox-v2 | 08-21 05:54:59 | 223 min | yes |
| bouncebit-ethena | 08-21 05:29:35 | 248 min | yes |
| covo-v2 | 08-21 08:37:59 | 60 min | yes |
| sacra | 08-21 08:32:35 | 65 min | yes |

`blackhaven` is the important one: it had the same gap I used to excuse ours, and it has since recovered. Hypersurface is now the only protocol in the sample still missing the 08-21 00:00 snapshot, and its last successful run is 2.4x staler than the next worst.

That last successful run was **23:35:35**, and this PR merged at **23:41:26**. The adapter has not produced a datapoint since.

I still cannot reproduce a failure locally — against merged `main` it returns 5.28M at latest and 5.31M timetravelled to the 08-21 snapshot that was missed, in 7-28s with no errors. So this looks environmental rather than a logic bug.

The change took the hyperliquid entry from 3 owners to 6 with `resolveUniV3` enabled, so the call count went up materially. If the runner is hitting a timeout or an RPC budget on HyperEVM, that would fit.

Could someone check the error log for this adapter? If it is the call volume I will gladly cut it down or split the pools — just say which you prefer. And if a revert is the faster path to getting TVL updating again, I am happy for you to revert this and I will resubmit in a leaner form.

PR #20643, opened 2026-08-22
Add token.select TVL adapter (Robinhood Chain)

Envision-Miles:
Adds a TVL adapter for token.select, a fair-launch launchpad on Robinhood Chain. The protocol is already listed on DefiLlama (https://defillama.com/protocol/token.select) from the fees/revenue adapter merged in DefiLlama/dimension-adapters, so this PR only adds TVL.

Every launch deploys a token through the factory `0xA94AA60e9c7f193BF678608D5837F0FD51794635` and seeds two Uniswap V3 1% pools: one against WETH and one against $SELECT. Both LP positions are custodied by an `lpVault` contract rather than by the token itself.

The adapter enumerates launched tokens from the factory's `NewTokenSelectToken` events, reads each token's `lpVault()` on-chain, and sums the whitelisted (WETH) side of the Uniswap V3 positions those vaults hold via `sumTokens2({ resolveUniV3: true })`.

Only the WETH side is counted, matching the other launchpads on this chain (noxa-fun, rhfun, stonkbrokers, peeps). The launched-token side and the token/$SELECT pools are excluded, so this understates total liquidity in the pools.

The vault is read per token rather than hardcoded. Today every launch shares one vault, but the factory exposes `deployLpVault()`, so reading it per token means a later vault is picked up automatically instead of silently under-reporting.

`doublecounted: true` is set, since the same liquidity is already counted as Uniswap V3 TVL.

## Summary by CodeRabbit

* **New Features**
  * Added TVL tracking for Token Select on Robinhood Chain.
  * Automatically discovers launched tokens and measures valid WETH-side Uniswap V3 liquidity.
  * Added TVL metadata and methodology details for improved transparency.

### DIFF
--- a/projects/token-select/index.js
+++ b/projects/token-select/index.js
@@ -29,7 +29,7 @@ async function tvl(api) {
   const tokens = logs.map((log) => log.tokenAddress)
   if (!tokens.length) return {}
 
-  const vaults = await api.multiCall({ abi: 'address:lpVault', calls: tokens, permitFailure: true })
+  const vaults = await api.multiCall({ abi: 'address:lpVault', calls: tokens })
   const owners = [...new Set(vaults.filter((vault) => vault && vault !== ADDRESSES.null))]
   if (!owners.length) return {}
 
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR

PR #20671, opened 2026-08-24
royco-v2: add Royco Day markets and entry-point deposits, fail-soft market reads

githubotoro:
Updates the existing `royco-v2` adapter. Single file changed: `projects/royco-v2/index.js`. No new npm dependencies; `pnpm-lock.yaml` / `pnpm-workspace.yaml` untouched.

### 1. Royco Day markets (new, all 4 chains)

A newer market generation behind its own factory (`0xaaaa...4400812`), with **three** tranches instead of two. Discovered from `MarketDeploymentCompleted` events, same as the existing V2 path.

Counted in the markets' **own tokens, never in NAV** — a Day market's NAV unit is whatever that market prices in (USD, BTC, ...), so a `nav` figure is not a dollar figure:

- **Senior + junior** share one collateral ledger, so each tranche's `totalAssets().collateralAssets` is added against the kernel's `collateralAsset`. The two claims together are the ledger, not a double count.
- **Liquidity provider tranche** holds the market's Balancer V3 BPT, whose pool is exactly `(senior tranche share, quote asset)`. The senior tranche's claim already covers every senior share in existence — including those sitting in the pool — so only the **quote leg** is new value, sized by a proportional unwrap (`poolQuoteBalance * bptOwned / bptTotalSupply`). That correctly ignores BPT held by third parties, who do LP into these pools.

### 2. Pending EntryPoint deposit requests (new, both generations)

`requestDeposit` escrows the tranche's own asset in the EntryPoint until an executor deposits it or the user cancels, so it sits outside every tranche's `totalAssets()`.

Neither EntryPoint exposes an aggregate (requests live in per-`(user, nonce)` mappings), but `requestDeposit` is the only path that moves a market asset in and execute/cancel move the exact amount back out, so `balanceOf` is the outstanding total. Verified against a full replay of every live request on mainnet and arbitrum — equal to the wei.

Only **market assets** are swept, never a tranche's own share token: the same contract escrows *redemption* requests as shares and accrues protocol fees as shares (on arbitrum it holds 99.8% of one junior tranche's entire supply), and those are already backed by `totalAssets()`.

### 3. Fail-soft market reads (fixes a live outage)

Royco Day deployment is **permissionless**, and a market answers through deployer-chosen parts (arbitrary tokens, its own price oracle), so it can revert by neglect or on purpose. Previously one reverting market zeroed an entire chain.

Market discovery now reads soft and drops an unreadable market **whole** — from its value *and* from the de-duplication — so a market that is not added is never subtracted either. De-duplication calls stay hard, because a skipped subtraction inflates TVL whereas a dropped market only under-reports its own size.

`permitFailure` alone was not enough: a read that exhausts the eth_call gas allowance fails the *batched* call rather than the one entry inside it, and the SDK then throws on that batch. A whole-batch failure now degrades to per-call reads, so the offending market reads null and is dropped while the rest still answer.

**This also fixes a current outage:** ethereum publishes nothing today because two V2 sNUSD markets revert on `totalAssets()`. With this change it publishes again (~$13.6M, about 59% of protocol TVL).

### No double counting

Every dollar is counted once, at its lowest layer. A market or entry-point leg denominated in one of Royco's own tokens is skipped, because the layer below already reports that backing in full — one mainnet Day market's collateral asset *is* srRoyUSDC, the vault this adapter already counts.

### Testing

`node test.js projects/royco-v2/index.js`, run repeatedly:

```
ethereum   13.62 M
avax        6.54 M
arbitrum    1.23 M
base        1.81 M
```

(base intermittently fails locally on free-tier `eth_getLogs` limits; it succeeds on retry and is unrelated to this change.)

Fault injection was used to confirm the fail-soft path: poisoning each layer of a Day market in turn — kernel, each of the three tranches, the Balancer pool, the collateral token — drops only that market and leaves the others' value intact and non-negative.

(1) Royco V2 market TVL: `totalAssets()` summed across senior and junior tranches (from `MarketDeployed` factory events) on every chain. (2) Royco Day markets, the newer three-tranche generation, counted in the markets' own tokens: each tranche's collateral claim, plus the quote leg of the liquidity tranche's Balancer BPT. (3) Deposit requests still queued at the Royco EntryPoints on every chain, for both generations. (4) The srRoyUSDC vault (mainnet) allocates USDC across Royco markets and other venues; its balance is added minus the portion already sitting in Royco markets. (5) The RoyWstEth vault (mainnet) adds its wstETH NAV minus the srRoyUSDC position it holds. Every figure is a token balance read from chain state.

## Summary by CodeRabbit

- **New Features**
  - Added support for newer Day markets alongside existing Dawn markets.
  - Expanded total value locked (TVL) coverage to include Day market collateral, quote assets, tranche assets, and escrowed deposits.
  - Added support for the Royco wstETH vault.
  - Improved handling of Royco-native assets to prevent double counting.

- **Bug Fixes**
  - Improved resilience when market data cannot be read by safely excluding affected markets.
  - Corrected strategy overlap handling across both market generations.

### DIFF
--- a/projects/royco-v2/index.js
+++ b/projects/royco-v2/index.js
@@ -2,28 +2,20 @@ const ADDRESSES = require('../helper/coreAssets.json')
 const sdk = require("@defillama/sdk");
 const { getLogs2 } = require("../helper/cache/getLogs");
 
-// TVL is counted once, at its lowest layer. Five sources, with subtractions where a higher layer
-// deposits into a lower one (internal de-dup within Royco, not cross-protocol):
+// TVL counted once, at its lowest layer. Five sources, de-duped where a higher layer deposits into a
+// lower one (internal to Royco, not cross-protocol):
 //   (1) Royco V2 markets (all chains): senior + junior tranche totalAssets().
-//   (2) Royco Day markets (all chains): the newer three-tranche generation, from its own factory.
-//       Counted in the markets' own TOKENS, never in NAV — a Day market's NAV unit is whatever it
-//       prices in (USD, BTC, ...). Senior + junior share one collateral ledger; the liquidity
-//       provider tranche (LPT) holds the market's Balancer V3 BPT, of which only the quote leg is
-//       new value (see unwrapDayBpt).
-//   (3) Deposit requests queued at the EntryPoints (all chains, both generations): requestDeposit
-//       escrows the tranche's own asset there, so it is in no tranche's totalAssets() until an
-//       executor deposits it or the user cancels (see getEscrowedDeposits).
-//   (4) srRoyUSDC vault (mainnet): a Royco-issued USDC vault. Its full balance is Royco's own TVL
-//       (Yearn-style: deposits count toward the issuing protocol even when forwarded elsewhere),
-//       minus the slice its strategies already hold inside (1) and (2).
-//   (5) RoyWstEth vault (mainnet): a Royco-issued wstETH vault. Its wstETH NAV is Royco TVL, minus
-//       the srRoyUSDC position its strategies hold (already counted in (4)).
+//   (2) Royco Day markets (all chains): newer three-tranche generation, counted in the markets' own
+//       TOKENS not NAV (a Day market's NAV unit is arbitrary); only the quote leg of the LPT's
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR

PR #20680, opened 2026-08-25
Add USDR (RISE Dollar) TVL adapter

sony-trinh:
Adds a TVL adapter for USDR (RISE Dollar), an M^0 extension on RISE Chain.

USDR is minted by wrapping $M into the extension contract, so the $M held by that
contract is the full backing of the supply. The adapter reads that balance on-chain.

$M is not priced under the `rise:` key by coins.llama.fi, but it uses the same address
on Ethereum where it is priced, so the balance is reported under the Ethereum key with
`skipChain: true` — same approach as `projects/securitize/index.js`.

Note on double counting: the $M backing USDR is bridged from Ethereum, where M0 is
listed. This mirrors `projects/saturn-protocol/index.js` (Saturn / USDat), which counts
the same $M token as backing and is not flagged `doublecounted`. Happy to add the flag
if you'd prefer it handled differently.

Validation:

```
$ node test.js projects/usdr/index.js
--- rise ---
M                         2.70 M
Total: 2.70 M

------ TVL ------
rise                      2.70 M
```

---

##### Logo (High resolution, will be shown with rounded borders): attached below (SVG)
USDR

## Summary by CodeRabbit

* **New Features**
  * Added TVL tracking for USDR (RISE Dollar) on the RISE network.
  * Reports USDR’s backing balance using the $M token valuation on Ethereum.

### DIFF
--- a/projects/usdr/index.js
+++ b/projects/usdr/index.js
@@ -1,0 +1,17 @@
+// USDR (RISE Dollar) is an M^0 extension: every USDR is minted by wrapping $M into the
+// extension contract, so the $M held by that contract is the full backing of the supply.
+const M = '0x866A2BF4E572CbcF37D5071A7a58503Bfb36be1b'
+const USDR = '0x62b7f5A5Be488ea58f660C5aff465647213Bc6e9'
+
+async function tvl(api) {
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR, I've added USDR in #20709

PR #20713, opened 2026-08-26
Fix: MerryForge locked-LP TVL (both legs)

EVMlord:
## Fix MerryForge locked-LP TVL (both legs)

sumTokens2({ resolveLP: true }) unwraps official UniV2 LP but launch tokens are unpriced, so only the WETH/USDG leg is counted.

Switched locked LP (LiquidityLocker) to:

  sumUnknownTokens({ api, tokensAndOwners, useDefaultCoreAssets: true, resolveLP: true })

so the token leg is priced from the pair against USDG/WETH.

misrepresentedTokens: true (unknown tokens reported as core assets).

Live token breakdown today is USDG + WETH only; DAVID/WETH locked LP
should be ~2× the WETH reserve.

## Listing metadata (not in this repo)

Website is missing on https://defillama.com/protocol/merryforge
(url is empty; twitter is set). Please set:
  url: https://www.merryforge.app

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved accounting for liquidity provider assets held in locked pools.
  - Permanently locked LP economic value is now represented more accurately in reported totals.
  - Updated asset classification to flag cases where token balances may be misrepresented.
  - Refined methodology details to clarify how locked LP assets are valued.


RohanNero:
Sorry, but for launchpads, we only track the quote asset, not the launched tokens

PR #20792, opened 2026-08-31
Add Hookers TVL adapter (Robinhood Chain)

gigahooker:
Adds a TVL adapter for **Hookers**, a Uniswap V4 token launchpad on Robinhood Chain.

This is not a new listing. Hookers is already listed (protocol id 8529, category Launchpad) from its fees adapter, but its `module` is still `dummy.js`, so it reports no TVL and cannot appear on TVL-ranked pages. Support confirmed a TVL adapter is what fills that in:

> "That page ranks by TVL and is populated from TVL adapter output, so a protocol without one can't appear there regardless of its fees data. The empty `chains` array and null TVL on `/protocols` are the same cause rather than two issues... for a v4 launchpad that's normally the liquidity your custody contracts hold in the PoolManager across launched pools."

**Current TVL: ~$128k** (52.0 ETH) across 15 launches.

**Chain:** Robinhood Chain (4663)

### What is counted

Every launch seeds one Uniswap V4 position owned by `LiquidityCustody`, held directly against the PoolManager with a zero salt — there is no per-pool contract and no LP NFT, so pool balances cannot be read directly and reserves are derived from the position's liquidity, tick range and the pool's current price.

Launches are discovered from the factory's `TokenLaunched` logs, which fire for every launch style the protocol's mechanism registry offers, hooked and hookless alike. The custody address is read from each log rather than hardcoded, so a launch style registered later is picked up without an adapter change. Position liquidity is read from Uniswap V4 StateView rather than from the protocol's own bookkeeping.

### On counting both sides

Both sides of each position are added, rather than dropping the launched token's side. Today this makes no difference to the reported figure — none of the launched tokens have a price, so they contribute nothing and headline TVL is the ETH side alone. Flagging it explicitly since several launchpad adapters drop the launched side outright; here it is a deliberate choice, not an oversight, so that a launched token starts counting if and when it earns an independent price. Happy to change it if you would rather it were excluded.

### Notes

- `doublecounted: true` — these are Uniswap V4 pools, so the same funds are also counted by the `uniswap-v4` adapter, which reads raw PoolManager balances on this chain.
- No new npm dependencies; no changes to `package.json` or any lockfile.
- Verified with `node test.js projects/hookers/index.js`, stable across repeated runs.

### Contracts (all source-verified)

- HookersFactory: https://robinhoodchain.blockscout.com/address/0xd07c5d29e4f956562a65741bfd9b12900a5caf63?tab=contract
- LiquidityCustody: https://robinhoodchain.blockscout.com/address/0x5c46e64151a215ebbee8e5405c4a1a0c53c7f3f8?tab=contract
- MechanismRegistry: https://robinhoodchain.blockscout.com/address/0x71c12b5bf7f6b056176c3d028d708f3397fc3ea2?tab=contract

## Summary by CodeRabbit

* **New Features**
  * Added TVL tracking for Hookers on Robinhood Chain.
  * Automatically discovers token launches and reports liquidity from seeded positions.
  * Included deployment details, tracking start block, methodology, and double-counting status.

### DIFF
--- a/projects/hookers/index.js
+++ b/projects/hookers/index.js
@@ -1,0 +1,125 @@
+const ADDRESSES = require('../helper/coreAssets.json')
+const { getLogs2 } = require('../helper/cache/getLogs')
+const { addUniV3LikePosition } = require('../helper/unwrapLPs')
+
+// Hookers is a Uniswap V4 token launchpad on Robinhood Chain.
+//
+// One transaction creates the token, opens its pool and seeds the entire supply
+// as a single one-sided position, then executes the creator's opening buy. The
+// position is held by LiquidityCustody directly against the PoolManager: there
+// is no per-pool contract and no LP NFT, so pool balances cannot be read and the
+// reserves are derived from the position's liquidity, tick range and the pool's
+// current price.
+//
+// The lock is structural rather than a timer. LiquidityCustody contains no code
+// path that decreases liquidity or moves the position, so the ETH buyers pay in
+// accumulates there permanently; only fees can leave.
+//
+// https://robinhoodchain.blockscout.com/address/0xd07c5d29e4f956562a65741bfd9b12900a5caf63?tab=contract
+const FACTORY = '0xd07c5d29e4f956562a65741bfd9b12900a5caf63'
+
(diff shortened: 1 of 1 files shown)


g1nt0ki:
thanks for the PR

PR #20800, opened 2026-09-01
fix(rate-x): read vault token accounts live instead of a frozen lookup table

toma-allez:
## What this fixes

`projects/rate-x/index.js` sums the token accounts listed in the hardcoded address lookup table `eP8LuPmLaF1wavSbaB4gbDAZ8vENqfWCL5KaJ2BRVyV`. That table is a snapshot, so a vault token account opened after it was last extended is never counted.

For the ONyc market this is most of the balance: 5 of its 7 vault accounts are listed, and the two that are missing hold the larger amounts. RateX reads as 23,626 ONyc where it holds 196,777, an understatement of roughly 8x. Seven of the table's mints are affected today, ONyc is just the largest.

## Approach

Take the `(vault authority, mint)` pairs the lookup table sanctions, then read those pairs' token accounts on-chain. A market's new account is counted as soon as it is funded, while the set of assets that count stays exactly the one the table already defines.

The pair restriction is deliberate rather than incidental. Summing every token account of a vault authority instead admits whatever else has been sent to it: today that would add 19.9M units of an unrelated pump.fun mint that no vault ever held, plus RateX's own synthetic tokens, whose `u64::MAX` supplies would dwarf the real collateral.

## Verification

Run against mainnet and compared with the current adapter, same block:

| | current | this PR |
|---|---|---|
| ONyc | 23,626 | 196,777 |
| mints reported | 34 | 34 |
| mints added | | none |
| mints removed | | none |
| rejected by the pair filter | | RateX-minted synthetics only, all at `u64::MAX` supply |

Seven mints pick up accounts they were missing. Nothing is introduced and nothing is lost, so the change is strictly the accounts the frozen table could not see.

## Notes

- No new npm dependencies, and `pnpm-lock.yaml` is untouched.
- The extra calls are one `getParsedTokenAccountsByOwner` per vault authority (62 today), paced through the repo's own `runInChunks` helper at 5 per batch with a 200ms gap, so the adapter stays inside RPC rate limits.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Solana TVL calculations by including only token balances associated with sanctioned vault authority and mint pairs.
  * Added on-chain account validation to prevent unrelated token accounts from being counted.
  * Simplified BSC TVL balance calculations for more consistent results.

* **Documentation**
  * Updated the TVL methodology description to reflect the sanctioned-pairs calculation approach.

### DIFF
--- a/projects/rate-x/index.js
+++ b/projects/rate-x/index.js
@@ -1,33 +1,23 @@
-const { getConnection, runInChunks } = require("../helper/solana");
+const { sumTokens2, getConnection, runInChunks } = require("../helper/solana");
 const { PublicKey } = require("@solana/web3.js");
 
-const TOKEN_PROGRAM_ID = new PublicKey("TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA");
+const TOKEN_PROGRAM_ID = "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA";
 // Published by RateX; lists the vault token accounts of every market.
 const LOOKUP_TABLE = new PublicKey("eP8LuPmLaF1wavSbaB4gbDAZ8vENqfWCL5KaJ2BRVyV");
-const SLEEP_TIME = 200;
-
-const parsedTokenAccount = (account) => {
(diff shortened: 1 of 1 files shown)


g1nt0ki:
thanks for the PR

PR #20826, opened 2026-09-02
fix(exponent): attribute Generic SY markets to their yield-bearing mint

toma-allez:
Each Exponent market is booked against the base asset from the API's `mintUnderlying`, so an asset with its own price never appears under its own name. The ONyc market reports `mintUnderlying` as USDe, which puts about $44M of ONyc PT under USDE and leaves ONyc's page with no Exponent line.

The Generic SY program records the yield-bearing mint it wraps, and the SY wraps it one for one. Checked across all 29 Generic SY markets, SY supply against the vault's on-chain balance: every ratio lands between 1.000000 and 1.000118. So the supply is the amount held and no rate applies. `mintRate` is a price ratio against the base asset, which is why it is dropped on this path rather than carried over: applying it on top of the yield-bearing token's own price would count the premium twice.

`projects/exponent-tranching` already reads the same on-chain field for the same reason, so the resolution is not new here.

### Fallbacks

Both lookups only improve attribution, so neither is allowed to cost the TVL:

- a mint with no price keeps the old base-asset path, so repointing can never drop a market to zero. The test asks the price service rather than hardcoding a list, so coverage tracks whatever `coins.llama.fi` knows.
- a failing price batch or an RPC that refuses `getProgramAccounts` degrades to that same path and logs via `sdk.log`.

Verified against live RPC by stubbing each failure in turn: 16 tokens either way, matching the pre-change set exactly, against 29 on the normal path.

### Measured, same run window

| | Before | After |
|---|---|---|
| Priced total | $115,603,340 | $115,684,990 (+0.07%, drift between reads) |
| ONYC | absent | $44,059,965 |
| USDE | $44,084,213 | $686, the real Ethena USDe market |
| Mints with no price | 5 | 2 |

Price coverage improves as a side effect: `sALP`, `sJLP` and the sUSDe placeholder resolve to real priced mints. The two remaining unpriced are the Meteora LP placeholders, unchanged by this PR.

## Summary by CodeRabbit

* **Enhancements**
  * TVL is now calculated from the total supply of each Exponent standardized-yield (SY) mint.
  * Valuation uses the price of the yield-bearing token wrapped by each SY mint.
  * Previous fallback valuation and metadata-based attribution behavior have been removed.

* **Documentation**
  * Updated the methodology description to reflect the SY-mint supply valuation approach.


FelixBruguera:
Hi, thanks for the PR. We added price support for exponent tokens, i've refactored the PR to add each mint directly
```
--- tvl ---
Exponent Wrapped USX      46.40 M
Exponent Wrapped ONyc     45.03 M
Exponent Wrapped BulkSOL  15.41 M
Exponent Wrapped eUSX     5.92 M
Exponent Wrapped hyloSOL  1.05 M
Exponent Wrapped fragSOL  979.46 k
Exponent Wrapped xSOL     623.60 k
Exponent Wrapped hyUSD    337.31 k
Exponent Wrapped hyloSOL+ 291.35 k
Exponent Wrapped sHYUSD   98.29 k
Exponent Wrapped stSLX    95.07 k
Exponent Wrapped stORE    93.64 k
Exponent Wrapped wfragBTC 9.41 k
Exponent Wrapped MLP (USDC-U 6.00 k
Exponent Wrapped jlUSDG   5.50 k
Exponent Wrapped JLP      2.45 k
Exponent Wrapped ALP      1.59 k
Exponent Wrapped dzSOL    1.10 k
Exponent Wrapped dSOL     946.00
Exponent Wrapped syrupUSDC 678.00
Exponent Wrapped MLP      458.00
Exponent Wrapped dfdvSOL  409.00
Exponent Wrapped INF      27.00
Exponent Wrapped sUSDE    5.00
Total: 116.36 M 
```

FelixBruguera (approved):

toma-allez:
Awesonme, Thanks!

PR #20850, opened 2026-09-03
feat(arcadia-finance-v2): add Robinhood chain

j-vp:
Arcadia will go live on Robinhood Chain (4663) with two lending pools, wETH and USDG, and the Uniswap V3 and V4 asset modules. 

Three things also changed:

* `resolveSlipstream` was passed unconditionally, but unwrapSlipstreamNFTs has no default NFT address for a chain without Slipstream and throws ("missing default uniswap nft address chain: robinhood"), which would have failed the whole chain's TVL. It is now derived from whether the chain has a Slipstream NFT configured, which leaves every existing chain unchanged.

* unwrapArcadiaAeroLP read the wrapped and staked Aerodrome position state with `target` set to an address the chain may not have, and unwrapped staked Slipstream against an undefined NFT. The id lists are empty on a chain without those modules, each is now gated on the module existing.

* unwrapUniswapV3NFTs got a Robinhood entry: the adapter resolves Uniswap V3 positions without passing an NFT address, so it relies on that switch, and every chain the adapter supports is already listed there. The V4 resolver already carries robinhood.

## Summary by CodeRabbit

* **New Features**
  * Added support for the Robinhood chain, including Uniswap V3 and V4 liquidity pool integrations.
  * Added support for unwrapping Uniswap V3 NFT positions on Robinhood.

* **Bug Fixes**
  * Improved compatibility with chains that do not provide Aerodrome or Slipstream NFT modules by safely skipping unavailable operations.

### DIFF
--- a/projects/arcadia-finance-v2/index.js
+++ b/projects/arcadia-finance-v2/index.js
@@ -47,6 +47,20 @@ const config = {
     resolveSlipstreamV2: false,
     resolveSlipstreamV3: true,
   },
+  robinhood: {
+    chainId: 4663,
+    factory: "0xDa14Fdd72345c4d2511357214c5B89A919768e59",
+    uniNFT: "0x73991a25C818Bf1f1128dEAaB1492D45638DE0D3",
+    uniV4NFT: "0x58daec3116aae6D93017bAAea7749052E8a04fA7",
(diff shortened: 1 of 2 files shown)


RohanNero:
Thank you for the PR

PR #20859, opened 2026-09-03
Add Longbow (Robinhood Chain)

longbowlend:
Adds a TVL adapter for **Longbow**, which creates and curates isolated Morpho Blue markets on Robinhood Chain.

**Listing metadata** — name Longbow · category Lending · chain Robinhood · url https://longbow.cash · twitter longbowlend

### Approach

The Morpho Blue singleton on this chain (`0x9D53d5E3bd5E8d4Cbfa6DB1ca238AEA02E651010`) is shared — Longbow is ~54 of ~148 markets — so `sumTokens2({ owner: morphoBlue, tokens })` would capture other curators' books. The markets are enumerated instead, the same shape `compound-blue` uses against the shared Polygon singleton, and each is attributed from its own state:

- `(totalSupplyAssets - totalBorrowAssets)` under the loan token — the analogue of what `sumTokens2` measures for a deployment a protocol owns outright. Borrowed assets have left the contract, so they are exported under `borrowed` instead, capped at supply exactly as `morpho-blue` caps it.
- net borrower collateral under the collateral token.

Morpho exposes no aggregate collateral getter, and `position()` would mean walking every borrower, so collateral is replayed from the only three events that move it: `SupplyCollateral`, `WithdrawCollateral`, `Liquidate`. Signatures were checked against `morpho-blue`'s `EventsLib.sol`; `WithdrawCollateral` in particular puts `assets` in the second data word because only `id`, `onBehalf` and `receiver` are indexed. 6,607 such logs exist on this chain today and `getLogs2` caches them.

### Double counting

`doublecounted: true`. These markets are inside `morpho-blue`'s robinhood TVL — that config enumerates every `CreateMarket` from block 286 with no blacklist.

Longbow's own ERC4626 vaults are deliberately **not** added on top. They deploy their assets into these same markets, so they are already inside `totalSupplyAssets`; counting them again would double count within this adapter. A first pass used `getCuratorExport` and counted only the vaults, which undercounted the protocol about 6x because most supply reaches the markets directly rather than through a vault.

### Verified

Executed against chain with a stand-in for the `api` surface:

| | |
|---|---|
| TVL | **$1,205,578** |
| borrowed | $332,780 |
| markets | 55 |

Largest positions: PONS 507,959 ($301,921), USDG 81,518 ($81,532), CASHCAT 194,052 ($55,459), WETH 3.48 ($8,695), SPY 10.2 ($7,884).

Two collateral tokens have no price on `coins.llama.fi` (`0x2e8c3116…`, `0x63c12667…`) and contribute nothing, which is expected rather than a defect.

### Notes for review

- **Why not `getCuratorExport`.** A first pass used it. It counts only deposits into Longbow's own ERC4626 vaults, and most supply reaches these markets directly rather than through a vault, so it reported the protocol at roughly a sixth of its size. The curator framework has no market-id key, hence the hand-rolled shape.
- **Collateral event caching.** `getLogs` keys its cache on `chain/target`, so three events against one singleton need distinct `extraKey`s or they overwrite each other and `parseLog` returns `null` on the warm path. Each sweep carries its own, matching `246Club`. `toBlock` is held 200 blocks off the head because the cache is append-only.
- **Vault-share collateral.** `morpho-blue` filters markets whose collateral is a MetaMorpho share. Checked all 54 here: none is, so no filter is needed.
- **Unpriced tokens.** Two collateral tokens have no entry on `coins.llama.fi` and contribute nothing. Expected for long-tail collateral on a new chain, not a defect.
- **Hardcoded market ids.** Explicit and reviewable, as in `compound-blue`. The tradeoff is that a new Longbow market needs a PR to be counted.

### Oracle providers

Two collateral tokens have no price on `coins.llama.fi` — they trade outside the Uniswap V3 pools the price service reads on this chain, so their markets would report as empty. Each is valued at its own market's Morpho oracle, the same feed Morpho's LTV and liquidation math use, and contributed as the loan asset. `misrepresentedTokens: true` is set for the substitution, matching `realt`, `piku-dao` and `binaryx-platform`.

| collateral | oracle | reads |
|---|---|---|
| wsNET `0x63C12667…86eA` | `0xCDE9599059f8Ae6D6B9F33A0aF7877827ec75F16` | 1,575.79 USDG |
| AI `0x2E8c3116…1e18` | `0x161f8693194ba2aac477b8f61c9c2eb566659627` | 0.2658 USDG |

`price()` is scaled `1e36 + loanDecimals - collateralDecimals`, so `collateral * price / 1e36` lands in raw loan-token units; the decimals adjustment is already inside the returned value. Every other collateral token contributes its real token and is priced by DefiLlama as usual.

## Summary by CodeRabbit

- **New Features**
  - Expanded Longbow coverage to 55 Morpho Blue markets, including WSNET-NN.
  - Longbow metrics now include supplied assets, outstanding borrowed assets, and collateral balances.
  - Collateral values are calculated using market oracle data and recorded collateral movements for more complete reporting.

- **Bug Fixes**
  - Markets without a valid collateral address are excluded from reporting.
  - Improved handling of historical event data, cached fallback results, and collateral calculation errors.

### DIFF
--- a/projects/longbow/index.js
+++ b/projects/longbow/index.js
@@ -1,29 +1,151 @@
-const { getCuratorExport } = require("../helper/curators")
+const { getLogs2 } = require("../helper/cache/getLogs")
+const abi = require("../helper/abis/morpho.json")
 
-// Longbow curates isolated Morpho markets on Robinhood Chain. Depositors hold one position per
-// vault; the curator routes that supply across markets.
+// Longbow creates and curates its own isolated markets on the Robinhood Chain Morpho Blue
+// deployment. That deployment is shared with other curators — Longbow is 55 of ~148 markets — so
+// the singleton's token balances cannot be attributed by a token union the way a sole-curator
+// deployment can. The markets are enumerated instead, the same shape compound-blue uses on
+// Polygon, and each one is attributed from its own state.
 //
-// Vault V2 vaults are discovered from the factory already registered for `robinhood` in
-// helper/curators/configs.js, filtered by the address that created them. That covers the USDG Core,
-// Frontier and ETH vaults plus the MonkeyHood partner vault, and picks up new ones without a config
-// change. The pre-V2 MetaMorpho vault is listed explicitly because there is no V1 vault factory
-// registered for this chain, so it cannot be found by owner.
+// tvl per market = (totalSupplyAssets - totalBorrowAssets) + collateral
+//   the first term is the loan-asset cash still held by the singleton for that market, which is
+//   what sumTokens2 measures for a deployment you own outright; borrowed assets have left the
+//   contract and are reported under borrowed instead.
+//   the second term is the borrower collateral held against that market.
 //
-// Only vault assets are counted. Collateral sitting in the underlying Morpho markets is already
-// part of the morpho-blue adapter's TVL on this chain, and the borrow side is Morpho's book rather
-// than ours, so neither is added here. getCuratorExport marks the result doublecounted.
-const configs = {
-  methodology: 'Counts the assets deposited in every Longbow-curated vault on Robinhood Chain. Marked as double counted because those deposits are supplied into Morpho markets, which Morpho Blue already reports.',
-  blockchains: {
-    robinhood: {
-      morphoVaultOwners: [
-        '0x1bf704707e9F3f407EbC9364fDAeD08C39893770',
-      ],
-      morpho: [
-        '0x8cb8AA35228c96C1C4E956E69AbAEBCc2aA7Dcfe',
-      ],
-    },
-  },
+// Morpho exposes no aggregate collateral getter, and position() would mean enumerating every
+// borrower, so collateral is replayed from the three events that move it. Vault deposits are not
+// added on top: Longbow's vaults deploy their assets into these same markets, so they are already
+// inside totalSupplyAssets and counting them again would double count. Morpho Blue reports these
+// markets too, hence doublecounted.
+const MORPHO = "0x9D53d5E3bd5E8d4Cbfa6DB1ca238AEA02E651010"
+const FROM_BLOCK = 286
+
+const eventAbis = {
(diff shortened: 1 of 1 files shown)


FelixBruguera:
Hi, thanks for the PR. We added price support for AI and wsNET
```
--- robinhood-borrowed ---
USDG                      441.11 k
Total: 441.11 k 

--- robinhood ---
wsNET                     1.23 M
PONS                      325.08 k
USDG                      85.05 k
CASHCAT                   51.93 k
WETH                      13.58 k
SPY                       7.87 k
AI                        6.86 k
STONKBROKER               4.24 k
NOTHING                   2.76 k
Index                     1.11 k
NVDA                      1.11 k
GOOGL                     198.10
AAPL                      189.25
PLTR                      171.16
QQQ                       101.23
SPCX                      33.68
MSFT                      16.64
TSLA                      2.49
USO                       0.00
AMZN                      0.00
MSTR                      0.00
MU                        0.00
USAR                      0.00
CRWV                      0.00
INTC                      0.00
Total: 1.73 M 

--- borrowed ---
USDG                      441.11 k
Total: 441.11 k 

--- tvl ---
wsNET                     1.23 M
PONS                      325.08 k
USDG                      85.06 k
CASHCAT                   51.93 k
WETH                      13.58 k
SPY                       7.87 k
AI                        6.86 k
STONKBROKER               4.24 k
NOTHING                   2.76 k
Index                     1.11 k
NVDA                      1.10 k
GOOGL                     198.00
AAPL                      189.00
PLTR                      171.00
QQQ                       101.00
SPCX                      34.00
MSFT                      17.00
TSLA                      2.00
Total: 1.73 M 

------ TVL ------
robinhood                 1.73 M
robinhood-borrowed        441.11 k
borrowed                  441.11 k

total                    1.73 M 
```

FelixBruguera (approved):

PR #20867, opened 2026-09-04
Orchard: add the v2 game contract to the TVL adapter (Robinhood Chain)

orchard-gold:
Orchard v2 went live on Robinhood Chain on 2026-09-03 (`0x86510b3df745C67a993A66CB08720Ed158d44549`, deployed at block 53786732). v1 (`0xEbB8b167c0992cFdc497A995a8Cf7167acAA0A1A`) stays open so players can claim what they have there, so the adapter now sums both game contracts.

What changes:
- AAPL held by each game contract is counted, minus that contract's `treasuryAccrued` / `adminAccrued` (protocol-owned fees waiting to be swept), plus the staker rake already flushed to SeedStaking. Same rule as before, applied to the two contracts.
- v2 keeps the ETH planted on the open round (and ETH players deposit for standing orders) as ETH until the keeper seals the round and swaps the whole round to AAPL, so v2's native balance is added as gas token. It is usually small and often zero between rounds.
- Staking (SEED in SeedStaking, shared by both versions) is unchanged.

Validated locally with `node test.js projects/orchard/index.js`: ~$31.4k on `robinhood` (v1 ≈ $31.0k, v2 ≈ $0.35k one day after launch; every accrual bucket of the v2 contract was reconciled against its AAPL balance on-chain). No dependency or lockfile changes.

---

Two things about the listing itself, which this repo can't change:
- The protocol page shows no website. Could you set it to https://orchard.gold?
- The description says the ETH is swapped to AAPL "in the same transaction", which is only true of v1. A version-neutral wording: "Orchard is a 5x5 plot game on Robinhood Chain, played in AAPL (Apple stock token). Players plant ETH or AAPL on a plot; every 75-second round one plot blooms and the players who planted there split the harvest from the other 24 plots, minus a 10% rake that funds SEED stakers, the Golden Apple jackpot and the treasury."

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Orchard total value locked (TVL) reporting by accounting for both the original and current game contracts.
  - Updated calculations to exclude accrued treasury and administrative fees.
  - Included SeedStaking rewards and eligible ETH balances from the current contract.
  - Applied contract coverage according to each version’s active period.

- **Documentation**
  - Updated the TVL methodology to explain coverage of both contract versions and the additional ETH component.

### DIFF
--- a/projects/orchard/index.js
+++ b/projects/orchard/index.js
@@ -6,7 +6,7 @@ const ORCHARD_V1 = '0xEbB8b167c0992cFdc497A995a8Cf7167acAA0A1A'
 const ORCHARD_V2 = '0x86510b3df745C67a993A66CB08720Ed158d44549' // live since 2026-09-03; v1 stays open for claims
 const SEED_STAKING = '0xd5d5f5Dff96E53fc6337b4aCf549d61b12882F2b'
 
-const GAMES = [ORCHARD_V1, ORCHARD_V2]
+const ORCHARD_V2_START = 1788476509 // v2 deployment
 
 // Orchard is a 5x5 plot game played in AAPL. Two versions of the game contract run side by side.
 //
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR, we will get the description and url updated shortly

PR #20888, opened 2026-09-05
Twofold: count vTWO governance deposits as staking, price TWO via CoinGecko

twofoldfi:
Follow-up to #20870 (Twofold, Robinhood Chain). Two changes to `projects/twofold/index.js`:

1. Staking now also counts the TWO held by the vTWO governance wrapper (`0x5c02401e945d86FB7109EC77F358498D6DA05950`), which locks TWO 1:1 for voting, alongside the two staking vaults.
2. TWO balances are re-keyed to `coingecko:twofold`. The token is listed on CoinGecko (id `twofold`) but the coins service has no entry for its Robinhood Chain address (`0x2A4a33A2163D005d8E7f1D9aC08d14c98db288d5`), so the merged adapter reported staking as 0 and dropped the TWO side of every pool. If you prefer to add the address mapping on your side instead, happy to drop that part.

`node test.js projects/twofold` output:

```
robinhood                 187.35 k
robinhood-staking         473.91 k
staking                   473.91 k
total                    187.35 k
```

## Summary by CodeRabbit

- **New Features**
  - Staking TVL now includes TWO held by the vTWO governance wrapper.
- **Updates**
  - TVL and staking TVL methodology now reflects vTWO holdings.
  - TWO pricing is no longer sourced through the removed CoinGecko-based calculation path.

PR #20897, opened 2026-09-05
Twofold: read both Registries, count reserves on every listed pool

twofoldfi:
Twofold launched a second stack of DualPool pools on Robinhood Chain on 2026-09-05 (new hook 0xd1BcbCCa41f3bdb6b4812652959c6dF725ea2Ac0, new Registry 0xdF1a23B1A7507Cc3B270DfA78FDD9ddA7bC36325). The first stack stays live and several of its pools still hold depositors, so the adapter now reads both Registries and sums hook.getReserves over every listed pool. It no longer skips listings flagged inactive: an inactive key with shares outstanding still holds funds, and the flag only controls whether the site offers new deposits.

Builds on #20888 (vTWO staking + CoinGecko pricing); rebases clean if that merges first.

node test.js projects/twofold:
robinhood-staking 439.00 k
robinhood 182.95 k

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented data collection from querying contracts before their deployment block.
  - Improved TWO price calculations by accurately combining matching balances and handling token decimals.
  - Avoided unnecessary TVL and staking TVL queries when no eligible contracts are active.
  - Applied deployment-aware checks to genesis position data for more reliable TVL reporting.

### DIFF
--- a/projects/twofold/index.js
+++ b/projects/twofold/index.js
@@ -9,16 +9,24 @@ const { sumTokens2 } = require("../helper/unwrapLPs");
 // stack stays withdrawable and several of its keys still hold depositors, so
 // both Registries are read. Reserves are counted whether or not a key is still
 // flagged active: an inactive key with shares outstanding still holds funds.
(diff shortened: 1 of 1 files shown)


twofoldfi:
Superseded by #20919, rebuilt on main after #20888.

PR #20906, opened 2026-09-06
Add Nullark TVL adapter

nullarks:
Nullark (https://nullark.com) is a fixed-denomination shielded ETH pool on Robinhood Chain. It uses Groth16 proofs and bundle-child notes: one public commitment can contain two independently spendable private child notes, each with its own secret and nullifier.

Pool: `0xf2F67b5Cbf2dEfCB447218ef903bE6AeF5fb2995`

The adapter reports shielded native ETH principal from on-chain accounting while excluding accrued protocol fees and unsolicited excess ETH.

Twitter: https://x.com/nullarkcom

Validated with `node test.js projects/nullark/index.js`.

## Summary by CodeRabbit

- **Updates**
  - Nullark TVL is now calculated from the pool’s actual native ETH balance, after accounting for accrued protocol fees.
  - TVL reporting now includes the resulting gas-token balance.
  - Historical tracking now begins from a defined start timestamp.
  - Reporting raises an error when accrued fees exceed the pool’s native ETH balance.

### DIFF
--- a/projects/nullark/index.js
+++ b/projects/nullark/index.js
@@ -1,0 +1,25 @@
+const POOL = '0xf2F67b5Cbf2dEfCB447218ef903bE6AeF5fb2995'
+const DEPLOYMENT_BLOCK = 53812864n
+
+async function tvl(api) {
+  if (api.block !== undefined && BigInt(api.block) < DEPLOYMENT_BLOCK) return
+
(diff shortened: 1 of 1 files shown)


RohanNero:
Thank you for the PR

PR #20989, opened 2026-09-11
stakepoint: update methodology

shaunking1988:
Not a new listing. Updating the methodology string on the existing StakePoint adapter to describe token and LP locking alongside staking, which is what the adapter actually counts.

methodology: TVL counts tokens held in StakePoint token locks, LP locks and staking pools. Locked and staked assets sit in Program Derived Addresses controlled by the StakePoint program, with no private keys, so they cannot be withdrawn before their unlock date. LP positions are unwrapped to their underlying assets. SPL and Token-2022 assets are both included, and SPT staking is reported separately under staking.

## Summary by CodeRabbit

- **New Features**
  - StakePoint now recognizes Raydium liquidity pool positions and displays their underlying assets.
  - Portfolio balances and total value locked (TVL) now include assets held through Raydium LP positions.

- **Bug Fixes**
  - Vaults with unavailable account data or zero balances are excluded from balance and TVL calculations.

- **Documentation**
  - Updated the TVL methodology to explain how liquidity pool positions are accounted for.


RohanNero:
Thank you for the PR

PR #20843, opened 2026-09-03
Add Solon (Morpho curator, Robinhood Chain)

solonlend:
##### Name (to be shown on DefiLlama):
Solon

##### Twitter Link:
https://x.com/Solonlend

##### List of audit links if any:
None yet. Solon deploys no custom pool contracts: markets and vaults are canonical Morpho Blue / Morpho Vault V2 factory deployments on Robinhood Chain (audited upstream). Solon-authored code is limited to oracle adapters and configuration.

##### Website Link:
https://solonlend.xyz

##### Logo (High resolution, will be shown with rounded borders):
https://solonlend.xyz/logo-1024.png

##### Current TVL:
~$500 (bootstrap phase, protocol-owned deposits)

##### Treasury Addresses (if the protocol has treasury)
0x750A0e73139A0cc062eaC29C1fc615d659E85246 (insurance fund, Robinhood Chain)

##### Chain:
Robinhood Chain

##### Short Description (to be shown on DefiLlama):
Solon is an independent Morpho curator on Robinhood Chain. It curates USDG vaults lending against issuer-verified tokenized equities, with on-chain issuer beacon verification of collateral and depth-measured supply caps.

##### Token address and ticker if any:
None

##### Category (full list at https://defillama.com/categories) *Please choose only one:
Risk Curators

##### Oracle Provider(s): Specify the oracle(s) used (e.g., Chainlink, Band, API3, TWAP, etc.):
Chainlink-compatible price feeds on Robinhood Chain, consumed through a Solon oracle adapter implementing Morpho's IOracle.

##### Implementation Details: Briefly describe how the oracle is integrated into your project:
Each Solon-created Morpho market uses an immutable oracle adapter (e.g. 0x4f6185269EbcAD4cFA0371d63b29d923956E60AC for AAPL/USDG) that reads the equity and USDG feeds, enforces staleness bounds, and returns the Morpho-scaled price. Note: TVL itself is computed purely from on-chain vault balances via the curators registry; the oracle is used by the markets, not by the TVL adapter.

##### Documentation/Proof: Provide links to documentation or any other resources that verify the oracle's usage:
https://solonlend.xyz/docs/overview (Oracles section); adapter verifiable at the address above on the Robinhood Chain explorer.

##### forkedFrom (Does your project originate from another project):
n/a — curator on the canonical Morpho Blue deployment (0x9D53d5E3bd5E8d4Cbfa6DB1ca238AEA02E651010)

[...]

PR #21001, opened 2026-09-12
solon: track curator vaults by owner + follow-up on missing listing

solonlend:
Small follow-up to #20843. Thanks a lot for maintaining this registry!

**Change:** switch the Solon curator entry to `morphoVaultOwners` (curator/owner `0xD0340F008bee8F6101cD609E0447cd3d44f8cC84`) so our current and any future Solon vaults on Robinhood Chain are picked up automatically, keeping the known vault as an explicit fallback.

Verified on-chain: Solon USDG Vault `0xCBB61788fB5A1969C93A222B1a12E4D1A50c6d99` (Vault V2, Robinhood Chain / 4663) — `owner()` = `0xD034...cC84`, `totalAssets()` ~ 101 USDG, so the adapter resolves a non-zero TVL.

Also, a gentle question (no rush): our earlier entry merged in #20843 but Solon does not seem to show on the site yet. Quite possibly something on our end — if there is anything we should adjust for it to index cleanly, we would be happy to. Thanks for taking a look!

Links: site https://solonlend.xyz · app https://solonlend.xyz/app/ · docs https://solonlend.xyz/docs/overview

### DIFF
--- a/registries/curators.js
+++ b/registries/curators.js
@@ -1120,11 +1120,14 @@ const configs = {
   },
   "solon": {
     config: {
-      methodology: 'Counts all assets deposited in Morpho vaults curated by Solon.',
+      methodology: 'Counts all assets deposited in Morpho vaults curated by Solon on Robinhood Chain.',
       blockchains: {
         robinhood: {
+          morphoVaultOwners: [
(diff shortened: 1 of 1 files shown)


FelixBruguera (approved):

FelixBruguera:
Hi @solonlend thanks for the PR. Solon is live on the site: https://defillama.com/protocol/solon

PR #21015, opened 2026-09-13
yield-ai: count DLP collateral and balances of unpadded safe addresses

ssadkov:
## Summary

- **Pad safe addresses before the indexer query.** `get_safes_range_info` returns some safes without leading zeros (18 of 274 today). `current_fungible_asset_balances` matches `owner_address` as an exact 64-hex string, so the fungible-asset balances of those safes were silently dropped (~$1.5k today: USD1 and WBTC).
- **Count Decibel DLP vault shares as USDC.** Safes post DLP shares as Echelon collateral (and can hold them idle). DLP has no price feed, so these positions were dropped entirely (~$3.4k today). They are now converted at the vault on-chain NAV per share: `vault::get_vault_net_asset_value / vault::get_vault_num_shares` on the Decibel package `0x50ead22a…`. Both views return 6-decimal base units, like USDC.
- **Echelon passes rethrow errors,** matching the existing Hyperion pass, so a failed per-safe read fails the run instead of undercounting.

## Notes

- `doublecounted: true` is unchanged (Echelon, Hyperion and the Decibel vault are tracked as separate protocols).
- The on-chain NAV per share (1.069443) matches the price Echelon itself uses for DLP collateral (1.069367).

## Test plan

`node test.js projects/yield-ai/index.js`

| | Before | After |
|---|---|---|
| USD1 | 65.39k | 66.40k |
| USDC | 26.52k | 29.90k |
| USDt | 7.71k | 7.72k |
| WBTC | 1.95k | 2.46k |
| **Total** | **101.58k** | **106.47k** |

Cross-checked against Yield AI hourly position snapshot for the same hour (independent reader): $106,471.75.

Run with `APTOS_RPC` pointed at a keyed Aptos Labs endpoint; the public fullnode rate-limits a full run.

## Summary by CodeRabbit

* **New Features**
  * Added support for valuing Decibel DLP vault shares in USDC using on-chain NAV data.
  * Included DLP shares held directly or posted as Echelon collateral in total value calculations.

* **Bug Fixes**
  * Ensured vault NAV and share counts are read from the same ledger state.
  * Improved safe address matching and retry handling for position lookups.

* **Documentation**
  * Updated valuation methodology to describe Decibel DLP tracking.

* **Changes**
  * Removed Moar Market integration.


ssadkov:
**CI note:** the `test` job fails on the public Aptos fullnode rate limit (the job log shows HTTP 429 from the first seconds of the run), not on the adapter logic.

**Evidence**

- The unmodified adapter on `main` fails the same way on the public fullnode. After a 5-minute cooldown, `node test.js projects/yield-ai/index.js` on `main` exits after 129s with `Failed to post https://fullnode.mainnet.aptoslabs.com/v1/view` in the Hyperion pass. With 274 safes the adapter needs more reads than the anonymous quota allows.
- With `APTOS_RPC` pointed at a keyed Aptos Labs endpoint, both versions complete:

| | `main` | this PR |
|---|---|---|
| USD1 | 65.39k | 66.39k |
| USDC | 26.52k | 29.81k |
| USDt | 7.71k | 7.72k |
| WBTC | 1.95k | 2.54k |
| **Total** | **101.58k** | **106.46k** |

This matches Yield AI own hourly position snapshot for the same period (106.47k).

**Follow-up commits**

- `20821d395` retry Echelon reads with exponential backoff and jitter; `vault_exists` is retried too; Echelon concurrency 10 -> 4.
- `57f63381a` drop the Moar Market pass. Yield AI safes hold no Moar deposits (`Moar=0 cells/0 assets` in a full run), and the pass cost one view call per safe per pool on every run. Total is 106.46k with and without it.

The Echelon passes rethrow on failure, like the existing Hyperion pass, so an RPC failure fails the run instead of silently lowering TVL.

Could a maintainer run the test with the DefiLlama RPC? Happy to adjust anything.

FelixBruguera (approved):

Issue #21019, opened 2026-09-14
beets-dex-v3: $1.13m TVL against $62.5m in the protocol's own backend, and the configured module beethovenx-v3/index.js is not in the repo


0xshubhs:
`beets-dex-v3` reports **$1,125,652** of TVL on Sonic. Beets' own backend reports **$62,521,958**
across 14 v3 pools on the same chain — DefiLlama is showing about 1.8% of it.

## the numbers

Beets' public GraphQL (`https://backend-v3.beets-ftm-node.com/graphql`), the same endpoint
yield-server already queries:

```
poolGetPools(where: { chainIn: [SONIC], protocolVersionIn: [3], minTvl: 10000 })
```

| | pools | total liquidity |
|---|---|---|
| Sonic v3 | 14 | **$62,521,958** |
| Sonic v2 | 8 | $607,045 |

Largest v3 pools: `bpt-smsUSD-vgUSDC` $45,825,030 · `ss-msUSD-ghUSDC` $5,974,601 ·
`ghUSDC-smsUSD` $3,764,285 · `vdUSD-Re7scUSD-vgUSDC` $2,901,476.

**The v2 row is the control.** Beets reports $607,045 for v2 and DefiLlama's own yields data reports
$605,735 for `beets-dex` — 0.2% apart. So the protocol/chain/version mapping is right and the
comparison is apples to apples; it is only v3 that disagrees.

## the module

`api.llama.fi/protocols` gives `beets-dex-v3` the module `beethovenx-v3/index.js`. There is no
`projects/beethovenx-v3` in this repo — folder or file — at `7a4bf0a`. The only beethoven folder is
`projects/beethovenx-sftmx`. Something is still producing $1.13M for the protocol, so I cannot tell
from outside whether the config points at a path that was renamed, or at one that never landed.

For contrast, the yields side of the house does have a working v3 adapter
(`src/adaptors/beets-dex-v3` in yield-server) and it reads the version-3 pools correctly — that is
where the 13-pool/$16.7M figure on the yields page comes from. So the protocol is covered in one
DefiLlama repo and not the other.

## what I can do

Happy to write the adapter if that is useful — Balancer v3 forks are a vault-balance sum, so it is
`sumTokens` over the Sonic vault rather than anything protocol-specific, and the pool list is one
GraphQL call. Say the word and I will send it. Flagging rather than opening a PR blind because a
missing module usually means the config needs a decision (rename vs re-add) that I cannot make from
[...]

PR #21022, opened 2026-09-14
beets dex v3: add a sonic tvl adapter that unwraps the boosted pools' erc4626 tokens

0xshubhs:
For #21019.

Beets DEX V3 on Sonic reports **$1.12M**. Beets' own backend reports **$62,957,861** across its v3
pools. The difference is entirely the boosted pools.

Balancer v3 boosted pools hold **ERC-4626 wrapper tokens** rather than the underlying, and those
wrappers are unpriced or mispriced, `vgUSDC`, `ghUSDC` and `bUSD` have no price at all,
`Re7scUSD` prices at $0.00 and `vdUSD` at $0.017. `bpt-smsUSD-vgUSDC` alone holds $46.1M of
`smsUSD` + `vgUSDC`. So summing raw vault balances counts almost nothing for the pools holding
nearly all the money.

This adapter does the normal v3 walk, `PoolRegistered` logs from the vault, then
`getPoolTokenInfo`, skipping pool BPTs so nested pools are not double counted, and then converts
each ERC-4626 balance to its underlying with `convertToAssets` before adding it. A token that does
not answer `asset()` is added as-is, and a failed `convertToAssets` falls back to the wrapper
balance, so nothing is dropped when the pattern does not apply.

## verification

I wrote the plain version first, without the unwrap, precisely to check the method against what is
published today:

| | |
|---|---|
| plain `v3Tvl`-style sum of raw vault balances | **$1.12M**: matches the published $1,124,645 |
| this adapter, ERC-4626 unwrapped | **$62.96M** |
| Beets backend, all v3 Sonic pools | **$62,957,861** |

So the method reproduces the current number before the fix and the protocol's own number after it.
Two runs of `node test.js projects/beethovenx-v3`, both $62.96M.

`fromBlock` 4660000 sits just before the first v3 pool on Sonic (earliest `createTime` 1737365442 =
2025-01-20, block 4662801).

Two things I deliberately did not do, both yours to decide:

- **I did not touch `v3Tvl` in `helper/balancer.js`.** The same undercount will hit any Balancer v3
  fork with boosted pools, so the fix arguably belongs there, but that helper reaches every caller
  and I would rather not change blast radius on my own judgement.
- **I could not tell which module currently serves this protocol.** `projects/beethovenx-v3` did not
  exist here, but neither do `projects/balancer-v3`, `projects/balancer` or `projects/beethovenx`,
  so if this file is not the one the config points at, it will need repointing to take effect.

## Summary by CodeRabbit

* **New Features**
  * Added TVL tracking for Beethoven X (Beets) v3 pools on the Sonic network.
  * Includes balances from pool tokens and supported yield-bearing assets while avoiding double-counting nested pools.


FelixBruguera:
Hi, thanks for the PR. The difference between our adapter and Beet's backend are these two distressed silos:
```
'Varlamore USDC Growth': '0xF6F87073cF8929C206A77b0694619DC776F89885'
 'Greenhouse USDC': '0xf6bC16B79c469b94Cdd25F3e2334DD4FEE47A581'
```
These two vaults are blacklisted from other adapters and not priced by our coins service

0xshubhs:
Thanks — that is the answer, and I should have checked whether the wrappers were unpriced *by
decision* before treating it as a gap.

Confirming your diagnosis with a number, since it may be useful: I re-ran the adapter with those two
silos excluded and nothing else changed.

| | |
|---|---|
| published today | $1.12M |
| unwrap everything (this PR) | $62.96M |
| unwrap everything **except** Varlamore USDC Growth + Greenhouse USDC | **$1.45M** |

So those two vaults are essentially the entire discrepancy — the remaining ERC-4626 wrappers
(`smsUSD`, `Re7scUSD`, `vdUSD`) are worth about $330k between them, which is not worth a PR. Happy
for this to stay closed.

PR #21024, opened 2026-09-14
edgeX: count MSCA USDC balance on Arc in TVL

0xsir0000:
Adds the Arc chain to the edgeX TVL adapter so the USDC balance held by the MSCA address `0xc64e528c6439a204da1b519e78aa43f9e4b32f00` on Arc is included. The same address is already tracked on Ethereum and Arbitrum.

- Token: Arc USDC `0x3600000000000000000000000000000000000000` (`ADDRESSES.arc.USDC`). On Arc the native USDC and the ERC-20 USDC are the same balance, so it is counted once.
- No new dependencies, config-only change in `registries/sumTokens/data3.js`.

Testing: `node test.js edgeX` runs successfully with `ARC_RPC` set (smoke-tested against the public Arc testnet RPC, since Arc mainnet has no public RPC yet); the arc chain executes and balance/pricing resolve correctly.

## Summary by CodeRabbit

* **New Features**
  * Added support for tracking USDC total value locked on the Arc chain.
  * Included Arc in the EdgeX project’s supported TVL tracking configuration.


g1nt0ki:
thanks for the PR

PR #21038, opened 2026-09-15
Iter: rename standard-spot, track Arc and RISE testnet deployments

hskang9:
**Update to an existing listing (Standard Spot Exchange → Iter). Not a new protocol.**

Standard rebranded to Iter (https://iter.cx). This renames `projects/standard-spot` to `projects/iter` and rewrites the adapter against the current contracts in https://github.com/iter-cx/iter-monorepo. A separate mail to metadata@defillama.com will ask for the display name/slug change.

**Chains**: at the protocol's request the adapter tracks the two deployments that carry the full contract set, Arc testnet (chain id 5042002) and RISE testnet (chain id 11155931), and drops the Somnia exchange-only generation. `arc_testnet` and `rise_testnet` are added to `projects/helper/chains.json`. The SDK ships no providers for them, so the adapter registers RPC, chain id and Multicall3 env defaults at load time (same approach as `projects/strato`); CI runs green with that.

**What the adapter counts**
- Orderbooks enumerated through the `OrderbookFactory` (`allPairs` + `getBaseQuote`): base/quote escrowed by resting limit orders.
- `StopOrderEngine` stop orderbooks: pending stop orders.
- `BandPoolFactory` pools: LP inventory held by each band pool.
- `PresaleLaunch`: settlement tokens committed to live presales (the coins being sold are not counted).

**Pricing** (`misrepresentedTokens: true`): testnet tokens have no price feed, so tokens are valued with the exchange's own market price. A stablecoin listed in the chain config is $1; every other token is worth its orderbook `mktPrice` times the USD price of the quote token, preferring a direct stablecoin book over a route through another token (e.g. WETH). Tokens the exchange cannot price stay as raw balances.

**Test**: `node test.js projects/iter` runs clean on both chains. RISE holds five test tokens plus 0.83 WETH in its orderbooks; all four RISE books are quoted in WETH and there is no WETH/stablecoin book yet, so USD stays 0 until one is listed. Arc's current generation has no pairs yet. The pricing path was checked against Somnia mainnet, where stablecoin books exist.

## Summary by CodeRabbit

- **New Features**
  - Added TVL tracking for Iter on the Somnia network.
  - TVL now includes orderbook assets, stop orders, band-pool liquidity, and presale settlement tokens.
  - Added support for the Arc Testnet and Rise Testnet networks.
  - Added stablecoin pricing coverage for supported networks.

- **Changes**
  - Removed the Standard Spot TVL adapter for Somnia.


FelixBruguera:
Hi, thanks for the PR. Sorry but we don't track testnet deployments, please open a new PR when you have deployed in rise and arc mainnets and ideally create a new file so we can keep Standard's historic TVL alongside Iter's

PR #21041, opened 2026-09-15
fix(kasu): count pending pools and tranche balances in TVL

kivanov82:
## What this changes

The Kasu adapter counted only the stablecoin held by the lending pool contracts. Two other contracts hold lender stablecoin on-chain and were missed:

- **`PendingPool`** — one per lending pool. Holds a lender's deposit from the moment it is requested until the next clearing either accepts it into the pool or refunds it. This is the direct analogue of the `totalPendingDeposits` that `credit-coop` counts as TVL in #20967.
- **`LendingPoolTranche`** — holds loss repayments until lenders claim them via `claimRepaidLoss`.

Both addresses come back from the same subgraph query as the pools (`pendingPool { id }`, `tranches { id }`), so this adds no extra requests. They are read defensively with `?.` and `|| []` because `cachedGraphQuery` falls back to a previously written cache if the subgraph is unreachable, and that cache may predate these fields.

`borrowed` is unchanged.

## Methodology

TVL is the stablecoin still held on-chain by Kasu: liquidity in the lending pools, deposits queued in the pending pools awaiting clearing, and loss repayments held by the tranches until claimed. Amounts lent out to borrowers — pool token supply minus pool holdings, plus loans funded off-chain and reported on-chain — remain counted as borrowed.

## Hallmark

Added `['2026-09-10', 'Lending book now counted as Active Loans, not TVL']` so the step in the TVL series from #20967 is labelled on the chart. Without it the drop reads as an outflow rather than a change in what is counted.

## Request: historical refill of `borrowed`

`borrowed` currently only has data from 2026-09-10 onward, so the Active Loans chart starts flat six days ago. The adapter runs correctly at historical blocks — every call it makes (`totalSupply`, `erc20:balanceOf`, `externalTVLOfPool`) is block-scoped, and pools that did not exist yet fail closed under `permitFailure`.

Verified against the TVL this protocol reported before #20967:

| Date | `borrowed` from this adapter | TVL reported that day |
|---|---|---|
| 2025-12-15 | $5.89M (incl. Plume pUSD $528k) | ~$5.3M |
| 2026-03-01 | $6.34M | $6,342,082 |
| 2026-06-01 | $10.31M | $10,313,984 |

Could `borrowed` be refilled back to the protocol's listing date (2025-12-09)? It reconstructs the full series accurately, including the now-retired Plume deployment.

## Chains affected

Base, XDC (AUDD and USDC deployments), Plume. Query verified against all four subgraphs.

## Test output

```
--- tvl ---
USDC                      5.00
Total: 5.00

--- borrowed ---
USDC                      6.69 M
AUDD                      5.81 M
Total: 12.50 M
```

The pending pools happen to be empty right now, just after a clearing; the balance is non-zero between a deposit arriving and the next clearing window.

## Summary by CodeRabbit

- **New Features**
  - Kasu TVL reporting now includes lending pools, pending pools, and tranches for a more complete view of on-chain assets.
  - Historical TVL queries now account for pools created in the past, including pools that have since stopped.
  - Borrowed amount reporting continues to use lending pools only.

- **Documentation**
  - Updated the TVL methodology to describe assets held across lending pools, pending pools, and tranches.


kivanov82:
Addressed the block-specific pool discovery point in 6d165cb.

The query filtered on `isStopped: false`, which is today's state, so a historical run would have dropped pools that were still active at the queried block. It now queries every pool and lets the block-scoped reads decide: pools that did not exist yet return 0 under `permitFailure`, and the pools that are stopped today are fully drained (verified on Base and XDC: zero balance and zero owed), so current values are unchanged.

Re-verified identical output at 2025-12-15, 2026-03-01, 2026-06-01 and today. This matters mainly for the historical refill asked for above.

FelixBruguera (changes requested):
Hi, thanks for the PR. Can you please remove the hallmark? the data was refilled and there's no drop on the chart now. I'll refill again after merging to update the historical TVL

kivanov82:
Done, removed in 7b8080d. Thanks for the refill.

FelixBruguera (approved):

Issue #14936, opened 2025-06-05
TVL calculation bug on YieldNest


danoctavian:
https://defillama.com/protocol/yieldnest

TVL on yieldnest is not counted correctly anymore hence the drop.

https://bscscan.com/address/0x132376b153d3cFf94615fe25712DB12CaAADf547

This asset has an ERC4626 interface and the asset is CoBTC.

Please see it here:

https://bscscan.com/address/0x918b3aa73e2D42D96CF64CBdB16838985992dAbc

1 CoBTC is pegged to 1 BTC.

Our vault is listed here:

https://github.com/DefiLlama/DefiLlama-Adapters/blob/main/projects/yieldnest/index.js#L27

noateden:
hi @danoctavian, where can I find backing assets for CoBTC? or any source I can check CoBTC like: CoinGecko, CMC?

drexthealpha:
I dug into this one. The adapter's matching live TVL right now, so the drop isn't reproducing for me — but I think the cause is real, just dormant.

I ran `node test.js projects/yieldnest/index.js`:
ethereum 18.71M / bsc 110.64k / total 18.82M

That's within ~ 0.1% of api.llama.fi/tvl/yieldnest (~$18.80M), and I'm not seeing any unrecognized-tokens output, so nothing's getting dropped right now.

I checked the ynCoBTCk vault on BSC directly to see why. Its asset() is 0x918b3aa73e2D42D96CF64CBdB16838985992dAbc (coBTC) like the issue says, but totalAssets() comes back as 303 base units — so it's basically empty and adds ~$0. coBTC isn't in helper/coreAssets.json, so my read is that once that vault refills, its balance goes unpriced and erc4626Sum drops it again, same as the original report.

So I think it only looks fixed because the vault drained, not because the pricing got handled. The durable fix would be adding coBTC to coreAssets mapped 1:1 to BTC. I can put up a PR for that if you want it, though it won't move the current number.

routebuilderdev:
Verified on current main: ynCoBTCk (0x132376b153d3cFf94615fe25712DB12CaAADf547) is in the bsc leg (projects/yieldnest/index.js:34, included in the erc4626Sum calls), and the listing tracks ~$25.8M live. The missing-vault gap looks closed — suggest closing.

PR #21059, opened 2026-09-15
Fix ynCoBTCk balance being silently dropped on YieldNest bsc TVL

Annkkitaaa:
This fixes a bug in the bsc TVL for the ynCoBTCk vault.

erc4626Sum resolves each vault through its own asset(), then prices that token. ynCoBTCk's asset is CoBTC, which doesn't have a price feed anywhere, so its balance was getting silently dropped instead of counted. Same root cause as #14936.

CoBTC is pegged 1:1 to BTC and has 8 decimals, same as WBTC, so I pulled ynCoBTCk out of the batched erc4626Sum call and count its totalAssets directly as WBTC instead. No scaling needed since the decimals already line up.

One honest note: right now the vault only holds about 0.000003 BTC, so this doesn't move the current TVL number at all, it's basically drained. The bug is still real though. If the vault fills back up, the balance would keep getting dropped without this fix. Someone else looked into the same issue on GitHub and landed on the same root cause.

Tested with node test.js projects/yieldnest/index.js. bsc total comes out about the same right now (~131.53k) since the vault balance is negligible, but the WBTC line item now shows up instead of being silently excluded.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved BSC total value locked (TVL) reporting for CoBTC.
  - CoBTC holdings are now valued using the WBTC price at the correct decimal precision, ensuring assets without a direct price feed are included accurately.
  - Other supported BSC vault assets continue to be included in TVL calculations.


FelixBruguera:
Hi, thanks for the PR. In cases like these we prefer to add pricing support for the underlying asset but as you said, the current TVL of this vault and the total supply of CoBTC in BSC are currently negligible, so it's not worth pricing right now

PR #21060, opened 2026-09-15
Add reBASE bonding-curve bank to WESO DeFi TVL

LunaClassicDAO:
## Summary
- Adds native `uusd` + `uluna` held in the **$reBASE** bonding-curve contract bank (`terra1uewxz67jhhhs2tj97pfm2egtk7zqxuhenm4y4m`, ~0.99M USTC / ~$5k plus a small LUNC balance).
- Does **not** use `curve_info.reserve` (~3.77M USTC) — that figure overstates vs the on-contract bank (funds are also split across ops/growth/dev accounts).
- Does **not** include those treasury sub-accounts in TVL.
- Leaves existing WESO curve, CWUSTC/CWLUNC wrap-bank, and JURIS/CWLUNC handling unchanged.

## Test plan
- [ ] CI adapter test for `projects/weso-defi` passes
- [ ] TVL USTC rises by roughly the reBASE contract bank `uusd`
- [ ] No use of reBASE `curve_info.reserve` / mint-cap fields

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved WESO DeFi TVL reporting by including the $reBASE bonding-curve reserve across the contract, reserve, operations, growth, development, and validator-multisig accounts.
  - Prevented routed USTC reserves from being omitted from reported TVL.

- **Documentation**
  - Updated methodology details to explain that USTC is counted using the bonding-curve reserve value and associated routed accounts.


LunaClassicDAO:
Updated: reBASE TVL now uses `curve_info.reserve` (native `uusd`), same pattern as the $WESO bonding curve.

The on-contract bank is only ~0.99M USTC; the rest of the reserve is routed to the reserve/ops/growth/dev pool accounts and the validator multisig (`to_validator_pct`). Those holdings line up with `curve_info.reserve` (~3.77M USTC / ~$19k at live USTC price), which is what Vyntrex’s reBASE/USTC market is reflecting (often shown ~2× as “liquidity”).

FelixBruguera (changes requested):
Hi, thanks for the PR. We would prefer to count count the balances of the banks for both reBASE and WESO bonding curves instead of using curve_info.reserve since that amount is not fully in the contract. I also think it would be better if we move these bonding curve reserves into a new adapter instead of combining it with the amm pairs. Let me know if you agree

LunaClassicDAO:
Hi Felix, thanks for the review.

We’d like to keep using `curve_info.reserve` for both the `$WESO` and `$reBASE` bonding curves on this adapter, rather than switching those curves to on-contract bank balances only.

For these cw20 bonding curves, `curve_info.reserve` is the protocol’s reserve accounting for circulating supply. The on-contract bank is only a slice of that: for `$WESO` the bank is ~357M LUNC while `curve_info.reserve` is ~7.29B LUNC; for `$reBASE` the bank is ~0.99M USTC while reserve is ~3.77M USTC. The gap is funds already routed under the same tokenomics (`to_validator_pct` and the reserve/ops/growth/dev pool accounts). Those accounts are shared across both curves, so counting only each contract bank understates locked backing, and adding the shared banks on top of both reserves would double-count.

That is also why the live `weso-defi` TVL is already dominated by the `$WESO` `curve_info.reserve` line. Switching both curves to bank-only would cut the protocol page by roughly ~$350k of LUNC backing that is still part of the curve design, not idle treasury we are trying to inflate.

On splitting into a separate bonding-curve adapter: we prefer to keep AMM pools, wraps, and these two curves under the single `weso-defi` listing for now, since they are one product surface on Terra Classic. Happy to revisit a split later if you still want that after the reserve methodology.

Happy to adjust comments/methodology wording if that helps, but we’d like to keep `curve_info.reserve` for both curves on this PR.

Thanks again

FelixBruguera:
> Hi Felix, thanks for the review.
> 
> We’d like to keep using `curve_info.reserve` for both the `$WESO` and `$reBASE` bonding curves on this adapter, rather than switching those curves to on-contract bank balances only.
> 
> For these cw20 bonding curves, `curve_info.reserve` is the protocol’s reserve accounting for circulating supply. The on-contract bank is only a slice of that: for `$WESO` the bank is ~357M LUNC while `curve_info.reserve` is ~7.29B LUNC; for `$reBASE` the bank is ~0.99M USTC while reserve is ~3.77M USTC. The gap is funds already routed under the same tokenomics (`to_validator_pct` and the reserve/ops/growth/dev pool accounts). Those accounts are shared across both curves, so counting only each contract bank understates locked backing, and adding the shared banks on top of both reserves would double-count.
> 
> That is also why the live `weso-defi` TVL is already dominated by the `$WESO` `curve_info.reserve` line. Switching both curves to bank-only would cut the protocol page by roughly ~$350k of LUNC backing that is still part of the curve design, not idle treasury we are trying to inflate.
> 
> On splitting into a separate bonding-curve adapter: we prefer to keep AMM pools, wraps, and these two curves under the single `weso-defi` listing for now, since they are one product surface on Terra Classic. Happy to revisit a split later if you still want that after the reserve methodology.
> 
> Happy to adjust comments/methodology wording if that helps, but we’d like to keep `curve_info.reserve` for both curves on this PR.
> 
> Thanks again

The problem with using the curve reserves is that the bank is what backs the token, so the number reported today by the WESO curve overstates what the curve can pay out. The contract's own sell simulation shows this:

```
# Simulate selling the entire circulating WESO supply (curve_info.supply) back to the curve
LCD=https://terra-classic-lcd.publicnode.com
WESO=terra13ryrrlcskwa05cd94h54c8rnztff9l82pp0zqnfvlwt77za8wjjsld36ms

QUERY=$(printf '%s' '{"simulate":{"action":"sell","payment":"689959105383141"}}' | base64)
curl -s "$LCD/cosmwasm/wasm/v1/contract/$WESO/smart/$QUERY"
# {"data":{"amount_out":"347109324229149","spot_price":"1000010"}}

# Native LUNC actually held by the curve contract
curl -s "$LCD/cosmos/bank/v1beta1/balances/$WESO/by_denom?denom=uluna"
# {"balance":{"denom":"uluna","amount":"357038500357953"}}
```

Redeeming the full supply returns about 347M LUNC, which lines up with the 357M LUNC in the contract, not the 7.29B in curve_info.reserve. We are fine keeping everything under the current weso-defi adapter, but please switch both the WESO and reBASE curves to the native bank balance of each curve contract

LunaClassicDAO:
Hi Felix, thanks for the follow-up and for the sell simulation.

We agree TVL should be backed by real balances, not a figure that cannot pay out. The reason we used `curve_info.reserve` was that most of the LUNC backing for these curves is not sitting as a liquid bank balance on the curve contract — it is staked from the bonding-curve validator multisig as intentional collateral so sells can be covered (funds can be unstaked when needed). That is part of how these cw20 bonding curves are built.

We can drop `curve_info.reserve` for both `$WESO` and `$reBASE` if we instead count **real** holdings:

1. **Liquid banks** on each curve contract (native `uluna` / `uusd`), same as you asked
2. **Liquid wrap banks** already in the adapter (CWLUNC / CWUSTC)
3. **AMM pool inventories** already in the adapter
4. **Staking + liquid balances on the bonding-curve validator multisig** `terra17x9tpp4ngn7hywnaleeqwjszyzws7hpy8tx0w35swyevuf9g4c9ssl7tml`
   - Currently ~**6.995B LUNC** delegated (across validators including LUNA CLASSIC DAO `terravaloper1yz3hv2k3a5z3tuu7fufcf7xvn2aawcu7c99lag`)
   - Plus ~**2.91M USTC** liquid in that same multisig bank (reBASE-side collateral)

That LUNC stake is what closes the gap between the ~357M LUNC liquid on the `$WESO` curve and the ~7.29B previously shown via `curve_info.reserve`. Your sell sim correctly shows only the liquid curve bank is immediately redeemable; the staked multisig balance is the rest of the protocol collateral behind the curves.

We are fine keeping everything under the current `weso-defi` adapter (as you already agreed). If you are okay with this approach, we will update the PR to:
- remove `curve_info.reserve` for `$WESO` and `$reBASE`
- add each curve’s native bank balance
- add the validator-multisig’s LUNC **delegations** + its liquid `uusd`
- leave AMM + wrap banks as they are
- tighten the methodology text accordingly

Does that work for you?

Thanks again

FelixBruguera:
> Hi Felix, thanks for the follow-up and for the sell simulation.
> 
> We agree TVL should be backed by real balances, not a figure that cannot pay out. The reason we used `curve_info.reserve` was that most of the LUNC backing for these curves is not sitting as a liquid bank balance on the curve contract — it is staked from the bonding-curve validator multisig as intentional collateral so sells can be covered (funds can be unstaked when needed). That is part of how these cw20 bonding curves are built.
> 
> We can drop `curve_info.reserve` for both `$WESO` and `$reBASE` if we instead count **real** holdings:
> 
> 1. **Liquid banks** on each curve contract (native `uluna` / `uusd`), same as you asked
> 2. **Liquid wrap banks** already in the adapter (CWLUNC / CWUSTC)
> 3. **AMM pool inventories** already in the adapter
> 4. **Staking + liquid balances on the bonding-curve validator multisig** `terra17x9tpp4ngn7hywnaleeqwjszyzws7hpy8tx0w35swyevuf9g4c9ssl7tml`
>    
>    * Currently ~**6.995B LUNC** delegated (across validators including LUNA CLASSIC DAO `terravaloper1yz3hv2k3a5z3tuu7fufcf7xvn2aawcu7c99lag`)
>    * Plus ~**2.91M USTC** liquid in that same multisig bank (reBASE-side collateral)
> 
> That LUNC stake is what closes the gap between the ~357M LUNC liquid on the `$WESO` curve and the ~7.29B previously shown via `curve_info.reserve`. Your sell sim correctly shows only the liquid curve bank is immediately redeemable; the staked multisig balance is the rest of the protocol collateral behind the curves.
> 
> We are fine keeping everything under the current `weso-defi` adapter (as you already agreed). If you are okay with this approach, we will update the PR to:
> 
> * remove `curve_info.reserve` for `$WESO` and `$reBASE`
> * add each curve’s native bank balance
> * add the validator-multisig’s LUNC **delegations** + its liquid `uusd`
> * leave AMM + wrap banks as they are
> * tighten the methodology text accordingly
> 
> Does that work for you?
> 
> Thanks again

Sorry but we can't count staked LUNC or USTC on a multisig as TVL. What we could do is add a treasury listing to weso that tracks those values, but they can't be on TVL

LunaClassicDAO:
Hi Felix, thanks for clarifying.

We’re going to pause this PR for now and **not** update the live `weso-defi` TVL methodology. We don’t want to change the existing `$WESO` curve accounting or lower the current protocol TVL display as part of adding `$reBASE`.

We’ll leave the open reBASE TVL work on hold. If we come back later, it would only be in a form that doesn’t rewrite the already-live `$WESO` TVL line. A separate treasury listing for the validator-multisig stake/balances is something we can consider later as its own discussion, outside this PR.

Thanks again for the review.

LunaClassicDAO:
Closing for now per the pause above. We may revisit later in a form that does not change the already-live `$WESO` TVL methodology.

PR #21062, opened 2026-09-16
suidex: read the full V3 pool set, not a fixed list of 6

CryptoMischief:
The SuiDex adapter counts every V2 pair from the factory, but only **6 hard-coded V3 CLMM pools**. There are 20. The 14 missing pools are about **$15,045 of TVL — roughly 30% of SuiDex V3** — the largest being BTC/USDC (~$7.6k) and USDC/SWARM (~$3.3k).

### Why not discover them on chain

V3 pools can't be enumerated from chain state. `pool.move` creates each pool with `transfer::share_object` and never records it anywhere, so there is no equivalent of the V2 factory's `all_pairs`, and no parent table for `getDynamicFields` to walk.

Querying `PoolCreatedEvent` doesn't work either — it returns only recent events, so pools created months ago are silently missed. A previous attempt at that approach would have reported *fewer* pools than the hard-coded list.

### What this does instead

Reads the pool set from the public SuiDex API and unions it with the known pool ids:

- pools added later are picked up without another PR;
- if the endpoint is unreachable the adapter still reports every pool it already knows, rather than dropping V3 to zero;
- pool ids that no longer resolve are skipped instead of throwing.

The V2 half is unchanged.

### Verification

`node test.js projects/suidex` → total **113.30k**, with the previously-missing tokens now present:

```
SUI      33.60 k      wBTC      4.50 k
USDC     25.08 k      TRUTH     3.06 k
SUITRUMP 22.19 k      MAGMA     1.25 k
VICTORY  13.21 k      WAL     509.00
Tree      4.22 k      DEEP    277.00
USDSUI    3.84 k      xBTC    210.00
                      XAUM    135.00
```

## Summary by CodeRabbit

- **New Features**
  - Expanded SuiDex V3 pool coverage by combining maintained pools with pools discovered through the public SuiDex API.
  - Added validation for pool identifiers to improve data accuracy.

- **Bug Fixes**
  - Unavailable or incomplete pool data is now skipped instead of interrupting TVL updates.
  - TVL retrieval continues using available pool data when the public pool API cannot be reached.

- **Documentation**
  - Updated methodology details to describe pool discovery and fallback behavior.


g1nt0ki:
thanks for the PR

PR #21070, opened 2026-09-16
morpho-blue: subtract collateral of blacklisted markets sharing a live token

thomas-chabert:
## Problem

`projects/morpho-blue/index.js` computes `tvl` as `sumTokens2({ owner: morphoBlue, tokens })`, where `tokens` is the union of the loan and collateral tokens of the non-blacklisted markets. Because that sums the Morpho contract's **whole** balance of each token, `blacklistedMarketIds` only removes a market's collateral from TVL when its collateral token happens to appear in no other live market. When the token is shared with a live market, the blacklisted market's collateral is still fully counted.

Live example on ethereum (block 25990181):

| market | LLTV | blacklisted | msY collateral |
|---|---|---|---|
| `0x23a7d0ff682b323363fb8ba58327ed87001f6306e09b7fd7413bbe4698e749c8` | 86% | yes | 29,964,612.02 |
| `0xb317d11c2bc2c0c8e6ea3c6517731cf667c86f2c716624be50319a6d32a97e8d` | 91.5% | no | 4,365,163.11 |

`msY` (`0x890A5122Aa1dA30fEC4286DE7904Ff808F0bd74A`) stays in the token list because of the second market, and `balanceOf(msY, morphoBlue) = 34,329,775.13` — exactly the two markets' collateral — so the blacklisted market's 29.96M msY (~$17.3M) was counted in ethereum TVL.

## Fix

For every id in `blacklistedMarketIds` whose collateral token is still part of the summed token set (and not already in `blackList`), subtract that market's collateral, in the spirit of the existing `ethenaBlacklist` block.

Morpho only tracks collateral per position (`Market` has no collateral total), so the suppliers are read from `SupplyCollateral` logs — one topic-filtered, cached query per market, keyed by market id so the cache stays valid when the blacklist changes — and their `position(id, user).collateral` is summed.

The subtraction is **capped at the contract's actual balance** of the token. Markets left with bad debt can still book collateral Morpho no longer holds: the resolv markets on ethereum account for 6,023,374 wstUSR and 1,778,911 RLP while the contract holds 0 wstUSR and 12,019 RLP. Without the cap those tokens go negative (I hit this on the first run).

## Verification

Each market's summed `position(...).collateral` was checked against the Morpho API's `state.collateralAssets` — all 12 ethereum blacklisted markets match to the wei. Full `tvl` runs per chain, before and after:

| chain | tokens subtracted | TVL change |
|---|---|---|
| ethereum | msY 29,964,612 · USR 313,675 · RLP 12,019 · sdeUSD 8,944,788 · deUSD 19,423 · PAXG 0.0596 · PT-RLP 522,405 | **−$17,341,066** (msY alone −$17,309,179) |
| base | USR 1,368,623 · RSS `0x50639c…` 0.00215 · RSS `0x7a305d…` 252,000 | **−$132,100** |
| arbitrum | xUSD 644,314 · RLP 68,899 | **−$24,217** |
| plume_mainnet | xUSD 65,845,263 | $0 — no DefiLlama price for that token |
| wc | none | no change |

`negative balances: 0` on every chain after the fix (the one that remains on ethereum, `0xC139190F447e929f090Edeb554D95AbB8b18aC1C` / USDtB, is pre-existing and comes from the `ethenaBlacklist` block — not touched here).

Per-chain notes on the other chains with a blacklist:
- **wc** — the blacklisted sdeUSD market's collateral token is used by no live market, so it was already excluded. No change.
- **plume_mainnet** — the blacklisted market shares xUSD with a live market, and it holds all 65,845,263 xUSD Morpho has on that chain (cap not engaged; the live market's xUSD collateral is 0).
- **arbitrum** — K is already excluded via `blackList`, but xUSD and RLP are shared with live markets.
- **base** — HERMES is already in `blackList`; USR and both RSS tokens are shared.
- **monad** has a token-level `blackList` entry but no `blacklistedMarketIds`, so nothing to do there.

Not addressed here, for scope: the idle loan-side liquidity of a blacklisted market (`totalSupplyAssets − totalBorrowAssets`) is also counted through its loan token. It currently adds up to ~$250 across all chains because these markets sit at ~100% utilisation, so I left it alone.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Morpho TVL calculations by excluding collateral held in blacklisted markets when it overlaps with active market assets.
  - Added safeguards to prevent collateral deductions from exceeding the assets actually held by Morpho.
  - Improved collateral position tracking for more accurate market-level accounting.


g1nt0ki:
superseded by https://github.com/DefiLlama/DefiLlama-Adapters/pull/21076

PR #21076, opened 2026-09-16
morpho-blue: don't count Vault V2 shares used as collateral

thomas-chabert:
`filterMarkets` drops markets whose collateral token is a Morpho vault, because the vault's own assets are already counted by the vault loop above it. It detects a vault by calling `withdrawQueueLength()` — a **MetaMorpho V1** function. On a Vault V2 that call reverts, `permitFailure` yields `null`, and `wql == null` **keeps** the market.

So Vault V2 shares posted as collateral are counted on top of the vault assets they are a claim on.

## The live case

`steakEURCV` (`0xbeef0c075da5d01112ae5cf34d257074fb5ddb2f`, Steakhouse Prime EURCV, created by the V2 factory `0xA1D94F746dEfa1928926b84fB2596c06926C0405`) has `asset() == EURCV`. On-chain right now:

```
balanceOf(EURCV,      steakEURCV vault)   97,987,292.75   <- vault loop   -> reported as "EURCV"
balanceOf(EURCV,      morphoBlue)          2,713,970.73   <- market loop  -> reported as "EURCV"
balanceOf(steakEURCV, morphoBlue)         12,524,270.78   <- market loop  -> reported as "steakEURCV"
```

Those 12,524,270.78 shares are 10.4% of the vault (`totalSupply` 120,764,952.85) and redeem for ~12,825,776 EURCV — the same EURCV as the first line. The API currently reports both `EURCV $115.52M` and `STEAKEURCV $14.19M`.

This is not caught by summing each token's balance once: `EURCV` and `steakEURCV` are different ERC-20s held by different owners, so each is counted exactly once while representing the same deposits. It is the same situation the `withdrawQueueLength` check already exists to prevent for V1.

## The change

Match the collateral token against the factory-derived vault list `getMorphoVaults(api, undefined, { getAllVaults: true })` already returns, which covers V1 and V2. Eight lines, no new network calls.

## Testing

`node test.js projects/morpho-blue/index.js` passes.

Ethereum: 553 -> 547 tokens. `steakEURCV` is removed; the other five are Morpho vault tokens used as collateral that the `withdrawQueueLength` heuristic was not catching. Nothing else moves.

## Not covered

A vault token that is also the **loan** asset of a market stays in the token list through that market. On Ethereum that is `steakUSDC` (~$0.77M). Excluding it would remove a genuine lending market's supply from TVL, which is a larger behavioural change — happy to follow up if you want it handled the same way.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Morpho Blue TVL accuracy by excluding MetaMorpho vault shares used as collateral.
  - Removed balances associated with blacklisted markets from TVL calculations, including idle loan liquidity and reconstructed collateral.
  - Added support for recovering collateral data from supply, withdrawal, and liquidation activity when available.
  - Data-reconstruction failures are now handled gracefully without interrupting processing.


g1nt0ki:
fixed here: https://github.com/DefiLlama/DefiLlama-Adapters/commit/db718ba8f5922b0dade0c07e46359da5cd9139a3

PR #21082, opened 2026-09-16
feat: Extend SatRush TVL metrics

ignatevdev:
With V2 release SatRush has added a second rewards vault (Rush Vault), similar to the previously listed Sats Vault. It holds unclaimed RUSH tokens and earns APY.

Additionally we've added a staking program, which allows users to stake their RUSH and receive yield in cbBTC, which is fed from a portion of the protocol fees.

RUSH CA is SATqS9DYpLQsM2z51P4QCoqJRHa5wboV4qjJerJRUSH and the token was listed on Coingecko.

It would also be great if the token metrics could be attached to our protocol dashboard.

Thanks in advance!

## Bug Fixes

* Updated SatRush value tracking to count cbBTC from the Sats Vault toward TVL.
* Updated staking tracking to include RUSH received through mining and RUSH staked in the SatRush staking program.
* Separated RUSH staking assets from Sats Vault cbBTC TVL calculations for more accurate totals.


FelixBruguera:
Hi, thanks for the PR. I've moved the Rush vault to the staking export because it holds the protocol's own token

FelixBruguera (approved):

PR #21131, opened 2026-09-18
whalehub: count BLUB supply as TVL (liquid staking); drop the decayed total_locked counter

combased-dev:
The WhaleHub adapter's `tvl()` reads `get_global_state().total_locked`. That field is a transient counter: it rises when AQUA is deposited and falls as deposits are unstaked or committed on to Aquarius ICE governance. It never represented the protocol's locked value. On DefiLlama it peaked around $2.4k in April 2026, decayed through the summer, read $6 on 2026-09-11, and has been $0 since. Meanwhile the protocol's actual locked value is reported under `staking`. Because staking and pool2 are excluded from headline TVL by default, the rankings feed currently publishes `tvl: 0`, `staking: 15383`, `pool2: 825`, `change_7d: -100` for a protocol where nothing changed on-chain.

WhaleHub is a liquid staking protocol on Stellar: users deposit AQUA and receive BLUB 1:1, and the AQUA is committed to Aquarius ICE governance on their behalf. The BLUB minted against those deposits is the on-chain record of the locked AQUA — the same shape as other liquid staking listings, where TVL is the underlying represented by the derivative. BLUB is a receipt for a third-party asset, not a governance token, so the own-token-staking exclusion does not apply.

**Change:** `tvl()` now counts `get_global_state().total_blub_supply`, priced as AQUA (1:1 by mint; no separate feed, hence `misrepresentedTokens: true`). The separate `staking` export is removed because it covered the same value and would double count under the include-staking toggle. `pool2` (vault user LP deposits, POL excluded) is unchanged. The now-unused `BLUB_TOKEN` and `POOL_0_SHARE_TOKEN` constants are dropped.

**On the choice of source.** The obvious alternative — the BLUB token balance held by the staking contract — is deliberately *not* used, because that address also holds protocol-owned BLUB from liquidity operations. A treasury rebalance on 2026-09-16 moved ~10M BLUB into the contract, and this adapter's own `staking` series jumped $11,028 → $14,901 (+35%) overnight with no user deposits involved. `total_blub_supply` is unaffected by such transfers. It does include the ~0.1x protocol-minted portion that funds protocol-owned liquidity, so this figure runs roughly 9% high — the conservative direction, and stable, which matters more for a TVL series than shaving that margin.

**Expected effect:** headline TVL moves from $0 to roughly $12k. Note this is *below* the current `staking` figure of ~$15.4k; the difference is protocol-owned BLUB that should not have been counted as user TVL in the first place. No new value is introduced.

**Verification:** `node test.js projects/whalehub/index.js` →

```
--- tvl ---
AQUA                      12.40 k
--- pool2 ---
AQUA                      881.60
```

Cross-check `get_global_state` on `CC72BEVVKHQ57PB5FCKAZYRXCSR6DOQSTN46QR7RZMMM64YWNRPDS24S`: `total_blub_supply` returned `349423651344803` (34,942,365 BLUB, 7 decimals) on 2026-09-18, while `total_locked` returned `0`.

**Category:** the listing is `Farm`; `Liquid Staking` is accurate given the model above. Happy to open the metadata change in defillama-server if you agree.

Links: https://app.whalehub.io · https://whalehub.io · https://whalehub-1.gitbook.io/whalehub

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected WhaleHub TVL reporting to use the total BLUB supply, improving accuracy when deposits move into ICE.
  * BLUB is now counted as AQUA only when the reported supply is greater than zero.

* **Changes**
  * Removed the separate BLUB staking metric from WhaleHub exports.
  * Updated the TVL methodology description to reflect the revised calculation.


FelixBruguera (changes requested):
Hi, thanks for the PR. Any reason why `total_blub_supply` doesnt reconcile with summing `get_ice_lock_authorization` ids? with the second approach TVL is around 16k vs 12k using `total_blub_supply`

```
async function tvl(api) {
  const state = await callSoroban(STAKING_CONTRACT, 'get_global_state')
  if (!state) throw new Error('whalehub: get_global_state returned nothing')

  let lockedAqua = BigInt(state.pending_aqua_for_ice ?? 0)
  const lockCount = Number(state.ice_lock_counter ?? 0)
  for (let id = 0; id < lockCount; id++) {
    const auth = await callSoroban(STAKING_CONTRACT, 'get_ice_lock_authorization', [{ type: 'u64', value: BigInt(id) }])
    if (!auth || !auth.executed) continue
    lockedAqua += BigInt(auth.aqua_amount)
  }

  api.add(AQUA_TOKEN, lockedAqua)
}
```

combased-dev:
Good catch, and the short answer is that they don't reconcile because `IceLockAuthorization` is an **append-only audit log**, not a position. I dug into it on-chain and it turns out neither number is the right one — including mine. Details below.

### Why the two disagree

`DataKey::IceLockAuth(id)` has exactly two writers in our contract: `authorize_ice_lock` creates the record, and `transfer_authorized_aqua` flips `executed = true`. Nothing ever decrements it, unsets it, or deletes the key. So summing executed ids gives **gross cumulative AQUA ever sent out to be locked**, which by construction only goes up.

Current state (2026-09-24, AQUA @ $0.00035665):

| | AQUA | USD |
|---|---:|---:|
| Σ executed ICE auths (96 of 104) | 44,756,328.95 | $15,962 |
| `total_blub_supply` | 35,289,441.38 | $12,586 |
| gap | 9,466,887.58 | $3,376 |

`pending_aqua_for_ice` is 0, and the 8 non-executed auths (9,137,215.29 AQUA) are correctly skipped by your loop.

### What's actually locked on-chain

The AQUA doesn't stay in the Soroban contract — it's transferred to the manager account, which creates Stellar-classic claimable balances for Aquarius ICE. Those are directly queryable, so they're the real ground truth:

```
GET /claimable_balances?claimant=GDERSSCK…XGTK&asset=AQUA:GBNZILST…AQUA
→ 386 balances, sponsor = that account on 100% of them
→ 90,491,112.78 AQUA, all still time-locked (0 claimable now), unlock dates 2029–2031
```

That's far more than either figure, because it includes protocol AQUA that never passed through the staking contract: **five balances created 2026-07-28/29 totalling 45,000,010 AQUA** (4 × 10M, 1 × 5M, unlock 2031), each a single-op `create_claimable_balance` straight from the manager, with no matching ICE-lock authorization — all of July's authorizations together come to 1,530,065 AQUA.

Strip those out and the two sources agree:

```
90,491,112.78  all claimable balances
-45,000,010.00  protocol AQUA locked 2026-07-28/29
=45,491,102.78  user-origin locks
 44,756,328.95  Σ executed ICE auths
        1.6%   apart
```

So the auth log is a faithful record of gross user-origin locks. It's just never reduced.

### Where that leaves the PR

You're right that your number is closer to physically locked AQUA than mine is, and I'd rather correct this than defend it. But I don't think either field should be the source:

1. **The auth log is monotonic by construction.** It tracks reality today only because these are 3–5 year locks. From Feb 2029 the earliest balances start unlocking and the log will keep counting them forever.
2. **It misses the 45M** of protocol AQUA that is locked but never went through the contract.
3. **And a problem with my own `total_blub_supply`, which I should have caught:** it isn't purely a mint counter. `lock()` adds the minted amount, but `stake()` also adds the deposited amount and `unstake()` subtracts it — so a user who locks AQUA and then stakes the resulting BLUB is counted twice. It's "minted via lock + currently staked − unstaked", not supply. That field shouldn't be used for TVL at all.

### Proposed fix

Read the claimable balances from Horizon directly. It's the actual locked position, it's verifiable by anyone, it needs no contract state, and it falls automatically as balances are claimed from 2029:

```js
const ICE_LOCKER = 'GDERSSCKJQPPXUQOZIOXGRVAGNLVPVZCJ2MAX7RCMVMWGRPVAEG7XGTK'
const AQUA_CLASSIC = 'AQUA:GBNZILSTVQZ4R7IKQDGHYGY2QXL5QOFJYQMXPKWRRM5PAV7Y4M67AQUA'

async function tvl(api) {
  let url = `https://horizon.stellar.org/claimable_balances?claimant=${ICE_LOCKER}&asset=${AQUA_CLASSIC}&limit=200`
  let total = 0
  while (url) {
    const res = await get(url)
    const recs = res._embedded.records
    if (!recs.length) break
    for (const r of recs) total += +r.amount
    if (recs.length < 200) break
    url = res._links.next.href
  }
  api.add(AQUA_TOKEN, total * 1e7)
}
```

Two pages, ~900ms, returns 90,491,112.78 AQUA ≈ **$32.3k**.

One question before I push it: **do you want the 45,000,010 AQUA of protocol-owned locks included?** AQUA is a third-party token, not ours, so the own-token exclusion doesn't apply and my reading is that it belongs in TVL — but it is protocol-owned rather than user-deposited, so if you'd rather have user-deposited only (~45.49M AQUA ≈ $16.2k) I can exclude that batch by its unlock date, or split it out under `treasury`/`ownTokens`. Happy either way — tell me which and I'll update the PR.

FelixBruguera (changes requested):
Hi @combased-dev thanks for the reply. We'd rather not include the protocol-owned liquidity in TVL, if you can move it to treasury that would work

PR #20726, opened 2026-08-27
Add Umbra Privacy (Solana) TVL adapter

umbra-team-dev:
Add Umbra Privacy (Solana) TVL adapter

New listing. Adapter at `projects/umbra-privacy/index.js`.

**Please note the slug.** `umbra` is already taken on DefiLlama by an unrelated
DEX on Eclipse (`projects/umbra/index.js`, category Dexs). This protocol is
Umbra Privacy on Solana and needs the slug `umbra-privacy`.

Umbra runs sixteen shielded pools. Each pool is a program-derived account that
owns exactly one SPL token account, and that token account custodies every
deposit made into the pool — a shield credits it, an unshield debits it.
Fourteen of those token accounts are summed as raw balances via
`sumTokens2({ tokenAccounts })`. The other two pools hold kmSOL / kmUSDC —
non-traded receipt tokens of a third-party (Overpass) wrapper over Kamino Lend
main-market deposits — and are counted at their on-chain redemption value in
SOL / USDC: the wrapper authority's kToken holdings × the Kamino collateral
exchange rate × the pool's share of the receipt supply.

Each vault balance reconciles exactly against the sum of its own deposit and
withdrawal history, walked transaction by transaction from `preTokenBalances` /
`postTokenBalances`. Happy to share that reconciliation if useful.

Notes for review:

- `timetravel: false` — the Solana helper reads current chain state.
- Explicit `tokenAccounts` rather than `owners`, because the `owners` path drops
  positions under 1e6 raw units and two pools (WBTC, PAXG) sit below that.
- `misrepresentedTokens: true` — kmSOL / kmUSDC have no market and no price
  feed anywhere, so their two pools are converted to underlying SOL / USDC at
  the on-chain redemption rate (verified against an actual on-chain withdrawal:
  725 kmSOL burned → 735.03 SOL paid out). The other fourteen pools are counted
  as the real token balances they hold.
- No `doublecounted` flag. The converted kmSOL/kmUSDC value (~$148k of ~$394k)
  represents deposits that are also counted at Kamino Lend; the other fourteen
  pools hold plain assets. Flagging the whole adapter looked wrong, but happy
  to change it if you'd rather.

---

##### Name (to be shown on DefiLlama):
Umbra Privacy

##### Twitter Link:
https://x.com/UmbraPrivacy

##### List of audit links if any:
No public audit links yet.

##### Website Link:
https://umbraprivacy.com

##### Logo (High resolution, will be shown with rounded borders):
[...]

PR #21143, opened 2026-09-20
umbra-privacy: count the UMBRA pool as TVL, not staking

umbra-team-dev:
Follow-up to #20726. Thanks for picking that up and for adding pricing for the
Overpass wrapper tokens — dropping the unwrap logic was the right call and the
kmSOL / kmUSDC balances now price natively.

One correction to the merged version: **the UMBRA pool is not staking**, so this
moves it back into `tvl`.

### Why it isn't staking

The UMBRA pool is one of the shielded pools, mechanically identical to the SOL,
USDC and USDT pools. A user shields UMBRA into it and later withdraws to a
different address, breaking the on-chain link between deposit and withdrawal.
That is the protocol's only function and it is the same code path for every
pool — the pool PDA owns one SPL token account that custodies the deposits.

Concretely, depositing UMBRA into that pool:

- earns no rewards or yield of any kind
- has no lockup, bonding or unbonding period
- confers no governance or voting rights
- gives the depositor no claim on protocol revenue

It is custody of a user deposit for privacy, not staked protocol-token. Bucketing
it as `staking` reads on the page as though UMBRA holders are staking into the
protocol, which isn't a thing Umbra offers.

I realise the usual reason to split a protocol's own token out of headline TVL is
to stop protocols inflating TVL with their own token, and that is a sensible
default — it just doesn't describe this mechanism. Happy to discuss if you would
rather keep it separated; if so, `ownTokens` would describe it more accurately
than `staking`, since nothing is being staked.

### Effect

| | before | after |
|---|---|---|
| `tvl` | ~$258.6k | ~$276.8k |
| `staking` | ~$18.2k | removed |

Total is unchanged — the UMBRA pool balance (~91,508 UMBRA, ~$18.2k) just moves
from the `staking` bucket into `tvl`.

Tested with `node test.js projects/umbra-privacy/index.js`: $276.77k under
`solana`, no `staking` key, lint clean, no lockfile changes.

The dynamic pool discovery from your rewrite is kept as-is.

### Unrelated observation, no change made here

The `owners` path in `projects/helper/solana.js` skips any position under `1e6`
raw units (line 313). Four of the pools currently sit below that and are silently
excluded: PAXG (1,061 raw), WBTC (1,921), BOT (8,402) and SPCX (88,626) — about
$20 in total, so immaterial to TVL, but they never appear. Flagging it rather
than changing it, since it's helper behaviour and the amounts don't justify a
special case. Happy to send a separate PR if you'd like them picked up.

## Summary by CodeRabbit

- **Updates**
  - UMBRA pool token balances are now included in total value locked (TVL) calculations alongside other shielded pools.
  - UMBRA is no longer categorized or reported as a separate staking adapter.
  - Methodology documentation now reflects the updated TVL treatment.


FelixBruguera:
Hi, thanks for the PR. Sorry but deposits of the protocol's own token are only counted under staking, even if the mechanism is not exactly staking, we can't add the protocol's own token to TVL. Here's a longer explanation of our methodology: https://docs.llama.fi/list-your-project/what-to-include-as-tvl

PR #21146, opened 2026-09-20
fix(yieldnest): price coBTC vault as BTC

### LINKED ISSUE #14936: TVL calculation bug on YieldNest

tlund123:
Fixes #14936.

YieldNest's `ynCoBTCk` ERC-4626 vault uses coBTC as its underlying asset. coBTC is pegged 1:1 to BTC, but it is not currently handled as a priced core asset, which can cause the vault balance to be omitted from TVL when it contains deposits.

This change handles the coBTC vault separately:
- reads `totalAssets()` directly on-chain
- reports the underlying balance against BSC BTCB for BTC pricing
- keeps the remaining BSC vaults on the existing `erc4626Sum` path

The coBTC vault is currently nearly empty, so this is primarily a durable fix for when deposits return rather than a material change to today's TVL.

No new dependencies or lockfile changes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved YieldNest TVL reporting on BSC.
  - The coBTC vault’s underlying BTCB assets are now included in total value calculations, keeping the vault visible when assets are replenished.


FelixBruguera:
Hi, thanks for the PR. In cases like these we prefer to add pricing support for the underlying asset but as you said, the current TVL of this vault is nearly empty, so it's not worth pricing right now

PR #10823, opened 2024-07-01
Add Scroll Chain Integration to Impermax Protocol Adapter

Elishaokon13:
### Description

This PR adds the necessary configuration for the Scroll chain to the Impermax protocol adapter. By including Scroll-specific factory addresses and updating the blacklist, this integration allows DeFiLlama to support Impermax on the Scroll chain.

### Changes Made

- **Configuration Update**: Added Scroll-specific factory addresses to the existing `config` object.

### Testing

- Verified that the Scroll chain configuration is correctly integrated and does not interfere with existing chains.
- Tested the adapter using the `test.js` script to ensure that the integration works without errors.

### Related Issues

No related issues at this time.

PR #21161, opened 2026-09-21
Add Ellipse TVL adapter

fa1rtrade:
Adds a TVL adapter for **Ellipse** (ellipse.fun), an RWA bridge + DEX + launchpad on Circle's Arc chain.

**Methodology (no double-counting):**
- **Bridge custody** — the real assets backing the 1:1 claim tokens, counted in their CustodyVaults on the origin chains: CRCL and GLD on Robinhood Chain, USDT and BTCB on BSC.
- **DEX (Arc)** — only the non-bridged liquidity (USDC + ELLIPSE) in the pools. The bridged tokens held in those pools are claims already counted in custody, so they are excluded to avoid double-counting.

v4 note: Uniswap v4 keeps all reserves in one PoolManager, so USDC/ELLIPSE are read from it (captures all price ranges). Ellipse operates the v4 pools on Arc; can move to per-pool attribution if other protocols deploy v4 there.

All four origin assets are already priced by DefiLlama (CRCL, GLD on robinhood; USDT, BTCB on bsc), and USDC + ELLIPSE are priced on arc.

Chains: arc, bsc, robinhood. Current TVL ~$3.8M.

## Summary by CodeRabbit

- **New Features**
  - Added TVL tracking for Ellipse across Robinhood Chain, BNB Smart Chain, and Arc.
  - Reports Ellipse custody reserves and non-bridged USDC and ELLIPSE liquidity in Arc v3 pools.
  - Applies updated methodology to prevent double-counting bridged assets.

- **Bug Fixes**
  - Excludes Uniswap v4 PoolManager balances that cannot be attributed specifically to Ellipse pools.
  - Ellipse v4 liquidity remains excluded until pool-level attribution is available.


fa1rtrade:
Note for the maintainer: this adapter is for the **existing protocol `ellipse.fun`**, which is currently a placeholder (`module: dummy.js`, no url, category Launchpad, chain Arc, no TVL). Please point that protocol's `module` to this adapter (`ellipse`) rather than creating a new entry, so it doesn't duplicate.

Two metadata requests while you're at it:
- **Rename** the protocol from `ellipse.fun` to **Ellipse** (slug `ellipse`) — that's the preferred name.
- Set **url** `https://ellipse.fun`. Category could be RWA (it's a bridge + DEX + launchpad); happy to go with your call.

The adapter reports arc (DEX: USDC + ELLIPSE only), plus bridge custody on bsc (USDT, BTCB) and robinhood (CRCL, GLD). Thanks!

fa1rtrade:
CodeRabbit was right, and the size of the error is worth stating plainly.

I summed USDC and ELLIPSE from the v4 PoolManager on the assumption that Ellipse was the only v4 deployer on Arc. That is no longer true. Measured today: the PoolManager held 3,330,521 USDC, while the USDC in Ellipse's own v4 pools was on the order of ten thousand. In the 50,000 blocks before that reading alone, 587 USDC-paired v4 pools were initialised on Arc by third parties — the ARCH/USDC pool CodeRabbit found is one of many.

So the adapter was reporting ~3.7M of TVL for Ellipse, most of it other protocols' money. I've dropped both PoolManager entries rather than ship that. The bot's new run confirms it: arc 26.59k, robinhood 54.26k, bsc 4.77k, total 85.62k.

This now under-reports Ellipse: its real v4 liquidity — the RWA pools and the launch pools — is left out. I'd rather under-report and add it back with per-PoolKey attribution (the launch pools are identifiable by their hook, which by construction only Ellipse's launchpad can use) than over-report now.

The metadata request in my earlier comment still stands: please point the existing ellipse.fun placeholder protocol at this adapter rather than creating a new entry, rename it to Ellipse, and set the url.

FelixBruguera (changes requested):
Hi, thanks for the PR. I refactored it to split bridge and dex into their own adapters. However, all the dex pools that you're counting are Uniswap V3 pools, does ellipse have their own pools? because these ones are Uniswap's TVL

fa1rtrade:
Thanks for the split - it looks right, and thanks for keeping the v4 note in the methodology.

Two things on the listing side, since CodeRabbit flagged the first:
- Please make sure both adapters land under one Ellipse parent. If only ellipse-bridge gets registered, the Arc TVL silently drops off.
- There's already a placeholder protocol ellipse.fun (module: dummy.js, no url, no TVL). Please reuse and rename that entry rather than adding a third one. URL: https://ellipse.fun.

@FelixBruguera 
Good question, and the honest answer has two parts.

Yes, they're Uniswap pools. Ellipse doesn't run its own AMM. It's a custodial bridge plus a launchpad that deploys each new market as a Uniswap pool on Arc (v3 originally, v4 since our V5 launchpad).

But the liquidity in the dex adapter isn't user liquidity sitting in Uniswap — it's protocol-owned and permanently locked. In the ELLIPSE/bCRCL pool, 100% of the active liquidity is a single LP NFT (#10823) owned by the launch lock contract 0xCeDa2CF7f8905b737c8cEd2474Ed88B1B8E7c98d, minted at launch, with no function that can return it. 61 of our 75 markets have a lock like that. The six asset/USDC v3 pools in the list are leftovers and hold ~22 USDC in total, so >99% of that 17.97k is the locked launch position.

So it comes down to your policy, and I'll take your call either way:

If liquidity in Uniswap pools is Uniswap's TVL, drop ellipse-dex and list only ellipse-bridge. I'd rather be listed at 54k that nobody can argue with than 72k that someone can.
If protocol-locked launchpad liquidity counts (as it does for other launchpads), then the current adapter is actually understating it: it should count the locked positions across all launched markets, not just these six v3 pools. Happy to write that version instead — it'd be a larger number, which is exactly why I'd rather you tell me the rule than pick it myself.
Same reasoning as the v4 fix earlier: I'd rather under-report.

FelixBruguera (approved):

fa1rtrade:
Merged, thanks — and understood on the dex, your call makes sense.

One thing left, from the listing side: the protocol entry still points at the placeholder. Right now ellipse.fun has module: dummy.js, no url, no chains and no TVL, so the merged adapter isn't wired to anything.

Could you point that entry at ellipse-bridge, rename it to Ellipse, and set url https://ellipse.fun? Category: RWA / Launchpad, whichever fits your taxonomy.

fa1rtrade:
Two metadata issues on the listed protocol, unrelated to the TVL adapter itself:

The description on api.llama.fi/summary/fees/ellipse (and on the protocol page) belongs to a different project: "Launchy lets users launch tokens on Starknet with locked Ekubo liquidity and creator fee sharing." Ellipse is on Arc, not Starknet, and has no relation to Ekubo. Suggested replacement: "Ellipse is an RWA bridge, DEX and token launchpad on Arc. It bridges CRCL, GLD, USDT and BTCB into 1:1 claim tokens backed by custody vaults on their origin chains, and runs Uniswap v4 pools with market-hours-aware dynamic fees."
url is empty; it should be https://ellipse.fun.

FelixBruguera:
> Two metadata issues on the listed protocol, unrelated to the TVL adapter itself:
> 
> The description on api.llama.fi/summary/fees/ellipse (and on the protocol page) belongs to a different project: "Launchy lets users launch tokens on Starknet with locked Ekubo liquidity and creator fee sharing." Ellipse is on Arc, not Starknet, and has no relation to Ekubo. Suggested replacement: "Ellipse is an RWA bridge, DEX and token launchpad on Arc. It bridges CRCL, GLD, USDT and BTCB into 1:1 claim tokens backed by custody vaults on their origin chains, and runs Uniswap v4 pools with market-hours-aware dynamic fees." url is empty; it should be https://ellipse.fun.

Hi, thanks for flagging this. I've updated the description. The links for new listings are not added immediately, it will be enabled after some time

PR #21173, opened 2026-09-21
Add Ravenhood on-chain treasury positions

troudev1133-ship-it:
Adds the missing on-chain Ravenhood treasury sources on Robinhood Chain.\n\nChanges:\n- unwraps DAO-staked UPDex Slipstream NFT positions from the UPDex voter/gauges\n- counts DAO-held Alandale Algebra/CL position NFTs\n- counts the Alandale V2 RVH/WETH LP held directly and staked in its gauge\n- counts DAO-held veLUTE NFTs as locked LUTE exposure\n- keeps RVH exposure under ownTokens instead of regular treasury TVL\n\nTested with:\n\\\\nnode test.js projects/treasury/ravenhood.js --chain robinhood\n\\\\nResult: robinhood TVL 78.53k, ownTokens 64.70k at test-time prices.

## Summary by CodeRabbit

- **New Features**
  - Treasury metrics now account for LUTE, UPDex, Alandale, and veLUTE holdings.
  - Total value and own-token calculations include staked positions, locked LUTE, concentrated-liquidity positions, and V2 liquidity holdings.
  - Alandale positions are valued using current market ticks or proportional pool reserves, as applicable.
  - Supported inclusion and exclusion rules are applied consistently when calculating totals.


projects/treasury/ravenhood.js
```diff
@@ -40,9 +98,104 @@ async function updexTvl(api) {
...
+    api,
+    position,
+    tick: Number(states[i].tick),
+    blacklistedTokens,
+    whitelistedTokens,
+  }))
+  return api.getBalances()
+}
+
+async function alandaleV2({ api, blacklistedTokens = [], whitelistedTokens = [] }) {
```
ccipherx, line 116: it is the total LP staked in the gauge, not the DAO's stake. It's correct only while the DAO is the sole staker. Could you read the DAO's stake directly from the gauge?

projects/treasury/ravenhood.js
```diff
@@ -40,9 +98,104 @@ async function updexTvl(api) {
...
+      calls.push({ target: gauges[i], params: [DAO_WALLET, index] })
+  })
+  if (!calls.length) return []
+  return api.multiCall({
+    abi: 'function stakedByIndex(address depositor,uint256 index) view returns (uint256)',
+    calls,
+  })
+}
+
+async function alandaleCl({ api, blacklistedTokens = [], whitelistedTokens = [] }) {
```
ccipherx, line 104: use `unwrapUniswapV3NFT` and drop addAlgebraPosition, tickToPrice and the ABIs

projects/treasury/ravenhood.js
```diff
@@ -40,9 +98,104 @@ async function updexTvl(api) {
...
+  const pools = await api.multiCall({
+    target: UPDEX_VOTER,
+    abi: 'function pools(uint256) view returns (address)',
+    calls: Array.from({ length: Number(count) }, (_, i) => i),
+  })
+  const gauges = (await api.multiCall({
+    target: UPDEX_VOTER,
+    abi: 'function gauges(address) view returns (address)',
+    calls: pools,
+    permitFailure: true,
```
ccipherx, line 118: not needed, can be removed

ccipherx (changes requested):

troudev1133-ship-it:
Thanks for the review. All three addressed in c344d91.

**1. `alandaleV2` gauge balance.** You were right, `LP.balanceOf(gauge)` is the gauge's whole stake. Now reads `LP.balanceOf(DAO)` plus `gauge.balanceOf(DAO)` and adds them.

Worth noting the reported TVL barely moves, because the DAO happens to be the only staker in that gauge today: `LP.balanceOf(gauge)` and `gauge.balanceOf(DAO)` both return 286.3747 LP right now. So this is a correctness fix for the moment a second staker shows up, exactly as you said, not a number change.

**2. `alandaleCl`.** Replaced with `unwrapUniswapV3NFT({ ..., isAlgebra: true })`. Removed `addAlgebraPosition`, `tickToPrice`, `ALANDALE_POSITIONS_ABI` and `ALANDALE_GLOBAL_STATE_ABI`.

**3. `permitFailure` on the `gauges()` lookup.** Removed. I read your comment as pointing at that line specifically, since `gauges(address)` is a mapping getter that cannot revert. Confirmed on chain: across all 93 pools from the voter it returns 93 real gauges, 0 zero-addresses and 0 reverts.

I kept `permitFailure` on the `stakedLength` call just below it, because that one does fail: `stakedLength`/`stakedByIndex` only exist on the CL gauges, and 6 of the 93 gauges revert on it. Happy to restructure that differently if you would prefer.

If you actually meant the whole gauge walk could go, that one I would push back on. Staked UPDex positions leave the DAO wallet and are held by the gauge, so owner-based discovery does not see them. There are 3 such positions right now (181294, 181035, 181670) and dropping the lookup would mean either losing them or hardcoding IDs that go stale.

`node test.js projects/treasury/ravenhood.js --chain robinhood` passes, total ~79.4k, in line with before.

ccipherx (approved):

PR #21194, opened 2026-09-22
feat(mercurifi): TVL adapter for the mercurifi launchpad on Arc

mercurifinance:
# mercurifi — TVL adapter

One new file: **`projects/mercurifi/index.js`**, chain `arc` only. The fees adapter for the same protocol was
merged earlier today as [dimension-adapters#9612](https://github.com/DefiLlama/dimension-adapters/pull/9612).

**Name:** mercurifi · **Website:** https://launch.mercuri.finance · **Twitter:** https://x.com/mercurifinance ·
**Chain:** Arc (5042) · **Category:** Launchpad · **Token:** none · **Github:** https://github.com/mercuri-finance/mercuri-launch-contracts

## What the protocol is

mercurifi Launch, a bonding-curve launchpad on Arc and the first app from mercurifi. A launch is one transaction: the factory deploys the token and its own
`BondingCurve`, and the token trades against that curve priced in USDC, which is also Arc's gas token. The buy
that sells the curve out opens a Uniswap v4 pool at the curve's last price in the same transaction and sends the
full-range position to a `LiquidityLocker` that has no withdrawal path.

## What is counted

1. **Native USDC held by curves that have not graduated.** A curve's reserve is its own balance in the chain's
   gas token (18 decimals), so it is summed at `ADDRESSES.null`. Graduated curves are skipped rather than summed
   as zero: a graduated curve keeps nothing — it hands its reserve to the pool and forwards the remainder.
2. **The USDC side of the locked Uniswap v4 positions**, through `resolveUniV4` with `whitelistedTokens:
   [ADDRESSES.arc.USDC]` (the 6-decimal ERC-20 interface the pools use).

The locker exposes `positionOf(address token)`, so the position ids are read straight from it with a multiCall
instead of enumerating `PositionManager` `Transfer` logs. Launched tokens and unclaimed fees are excluded.

`doublecounted: true`, because the locked positions already sit inside Uniswap V4's TVL on Arc.

## Test

```
node test.js projects/mercurifi/index.js

--- arc ---
USDC                      27.94
Total: 27.94
```

That is exactly the USDC held by the nine live curves today, checked against `eth_getBalance` on each curve
(27.942354 USDC) and against the protocol's own read API, https://launch-api.mercuri.finance/v1/stats.

**Is a small number correct?** Yes. The protocol went live on 21 September 2026 and nothing has graduated yet,
so the second leg contributes zero today and the whole TVL is the USDC sitting in live curves. The Uniswap v4
path is exercised by the same code the fees adapter's `Graduated` handling covers and will start reporting with
the first graduation.

`npx eslint projects/mercurifi/index.js` and `git diff --check` pass. No dependencies added, no shared files
touched.

## Summary by CodeRabbit

* **New Features**
  * Added MercuriFi Launch TVL tracking on the Arc network.
  * Reports USDC reserves held in active bonding curves and associated Uniswap V4 liquidity positions.
  * Excludes liquidity from bonding curves that have graduated.
  * Includes coverage beginning September 21, 2026.
  * Returns no TVL result when no launches or eligible liquidity positions are available.


projects/mercurifi/index.js
```diff
@@ -0,0 +1,60 @@
...
+  // reserve to the pool and forwards the remainder, so it is skipped rather than summed as zero.
+  const phases = await api.multiCall({ abi: 'uint8:phase', calls: launches.map(i => i.curve) })
+  const liveCurves = launches.filter((_, i) => +phases[i] !== PHASE_GRADUATED).map(i => i.curve)
+  await sumTokens2({ api, owners: liveCurves, tokens: [NATIVE_USDC] })
+
+  // 2. USDC side of the permanently locked Uniswap V4 positions. The locker records the position it holds for
+  // each token, so the ids are read from it rather than enumerated from PositionManager Transfer events.
+  const positions = await api.multiCall({
+    target: LIQUIDITY_LOCKER, abi: 'function positionOf(address) view returns (uint256)',
+    calls: launches.map(i => i.token), permitFailure: true,
```
ccipherx: remove permitfailure as positionOf is a public mapping and returns 0 for non-graduated tokens, so it won't revert. 

`permitFailure: true` only hides RPC failures, which could silently under-report TVL. Better to let the adapter throw.

ccipherx (changes requested):

projects/mercurifi/index.js
mercurifinance: Done in 00787d5. Removed permitFailure, so an RPC failure now throws instead of under-reporting. Also dropped the null guard in the filter since results can't be null anymore. Test still passes (arc 631.66 USDC today).

ccipherx (approved):

